PR #25, opened 2020-03-21
Translate README.md to English

vertonghenb:
Since I can't read mandarin and I guess not everyone can, we should translate the docs to english.


ElderJames:
Excellent, but it needs to be perfected.

We need to refer to the documentation of ant-design to create an [English](https://github.com/ant-design/ant-design/blob/master/README.md) and a [Chinese](https://github.com/ant-design/ant-design/blob/master/README-zh_CN.md) README respectively.

vertonghenb:
> Excellent, but it needs to be perfected.
> 
> 
> 
> We need to refer to the documentation of ant-design to create an [English](https://github.com/ant-design/ant-design/blob/master/README.md) and a [Chinese](https://github.com/ant-design/ant-design/blob/master/README-zh_CN.md) README respectively.

So what is the requested change to this PR ?

PR #32, opened 2020-03-22
Add EditorConfig for naming conventions

vertonghenb:
- Invalid Property Naming convention will now throw an error instead of a warning, which will act like a Linter. However the build will still succeed on the GitHub workflow and locally(with errors) due to [this issue that has to be resolved by Microsoft](dotnet/roslyn#33558)
- Disabled the XML comments, it's handy but not always needed at this point.

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+[*.cs]
+
+# CS1591: Missing XML comment for publicly visible type or member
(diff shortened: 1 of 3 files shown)


vertonghenb:
Finished, the build cannot be stopped using this linter but the error is more clear now.

PR #43, opened 2020-03-26
激进的开发者，升级到最新发布版本作为构建版本

lindexi:
构建版本仅影响构建，不影响运行版本


ElderJames:
I want to keep the build as a minimum, because we can be alerted to change documentation and other build tools when incompatibilities occur.

PR #50, opened 2020-03-26
Fix nuget icon

lindexi:
[Packaging Icon within the nupkg · NuGet/Home Wiki](https://github.com/NuGet/Home/wiki/Packaging-Icon-within-the-nupkg )

[Packaging Icon within the nupkg · NuGet/Home Wiki](https://github.com/NuGet/Home/wiki/Packaging-Icon-within-the-nupkg )

从官方文档上说，推荐将 Icon 放在包里面，同时支持的格式是 jpg 和 png 格式

It is a path to an image file within the package, often shown in UIs like nuget.org as the package icon. Image file size is limited to 1 MB. Supported file formats include JPEG and PNG. We recommend an image resolution of 128x128.


lindexi:
图片

PR #66, opened 2020-04-06
Docs/demo building

geffzhang:
集成abp 模块化框架


ElderJames:
Thanks for your contribution.

Integrating other frameworks adds unnecessary complexity to the component library and makes it difficult for contributors to get started.

So I still think I will do this integration in another project, and then I will refer to this PR.

PR #81, opened 2020-04-22
chore: update AddEventListener & stringify event args

Brian-Ding:


components/core/JsInterop/interop.ts
```diff
@@ -1,95 +1,104 @@
...
+    let dom = getDom(selector);
+    dom.blur();
 }
 
 export function log(text) {
-  console.log(text);
+    console.log(text);
 }
 
 export function BackTop(element) {
```
ElderJames, line 104: Indent two letters

Brian-Ding: fixed

PR #92, opened 2020-05-03
feat: add js methods

mutouzdl:


components/core/JsInterop/interop.ts
```diff
@@ -43,8 +46,8 @@ export function addDomEventListener(element, eventName, invoker) {
   if (element == 'window') {
     window.addEventListener(eventName, callback);
   } else {
-    let dom = getDom(element);
-    (dom as HTMLElement).addEventListener(eventName, callback);
+      let dom = getDom(element);
```
ElderJames, line 49: We need to indent only two characters at here.

mutouzdl: fixed

PR #118, opened 2020-05-16
fix:  AntDrawer title 无法显示问题。

1002527441:
fix:  AntDrawer closeable 无法关闭的问题


ElderJames:
Close this pr because it duplicates with #123 .

PR #132, opened 2020-05-19
docs:input demo done.

1002527441:


1002527441:
change div to br in input demo

PR #153, opened 2020-05-25
Make github main language as C#

baiyunchen:
Hi Team:
I found this project is marked as `CSS` project in github. So I create this pr to make github use `C#` as main language.

This pr will make `.css` and `.less` file list as C#, after apply these changes, the github display will as below:
image


ElderJames:
Thank you for your contribution @baiyunchen .
The less file is useful for showing that we synchronize the styles from ant-design, but you can try to let razor files be counted as c#.

PR #202, opened 2020-06-08
feat: form supports loading state

TimChen44:
提供From组件载入状态属性，功能类似“table”的“Loading”属性
Provide loading status attribute, function similar to "table" "Loading" attribute

### DIFF
--- a/components/form/Form.razor
+++ b/components/form/Form.razor
@@ -4,24 +4,26 @@
 @using Microsoft.AspNetCore.Components.Forms;
 @using AntDesign.Internal;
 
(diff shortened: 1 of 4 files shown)


ElderJames:
Thanks for your PR, @TimChen44 !
This feature is not available in Ant-Design. Is there any usage scenario for this feature?

TimChen44:
This function is inspired by the "Loading" property of "Table", the purpose is to simplify coding, and there is no reference to "Ant-Design"
此功能从“Table“的”Loading“属性获得启发，目的是简化编码，并没有参考”Ant-Design“

PR #259, opened 2020-06-27
feat: clearall method to clear all items in the list

Ratomir:
I have to clear all selected tag values in the input filed of a select component in the form. So, I made a couple of changes in the base code of the Select component. Here are the explanations of changes.

1. ClearAll() method
    Exposed method to a parent component.
2. ResetState() method
    I made this method to avoid code duplicate of OnClearClick and ClearAll methods. It is always a better approach to have one main method for two or more same actions.
3. Removed all code of OnClearClick() method and called ResetState() method.

Please, notice that the md Select/demo/clear-values.md file should be rewritten based on the appropriate documentation. I just copied it from the bordered.md, but you know better than me what should be in there.

I am open to any kinds of suggestions. :) 

Kind regards,
Ratomir

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -244,6 +244,15 @@ namespace AntDesign
                 _ => SelectMode == SelectMode.Default ? OneOf<string, RenderFragment>.FromT1((option?.ChildContent)) : option?.Label ?? option.Value,
             };
         }
+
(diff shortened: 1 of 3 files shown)


ElderJames:
Thank you for your contribution!
That's a nice feature, we will review it later.

PR #286, opened 2020-06-30
feat: new mentions components

1002527441:


components/mentions/style/index.less
```diff
@@ -14,7 +14,7 @@
   padding: 0;
   overflow: hidden;
   line-height: @line-height-base;
-  white-space: pre-wrap;
+  white-space: inherit;
```
ElderJames, line 17: Don't modify the less files form ant design

PR #295, opened 2020-07-02
fix: drawer OnOpen bull

TimChen44:


components/drawer/DrawerTemplate.razor.cs
```diff
@@ -25,7 +25,8 @@ protected async Task CloseAsync(TResult result)
         protected override void OnInitialized()
         {
             base.OnInitialized();
-            DrawerRef.OnOpen.Invoke();
+            if (DrawerRef.OnOpen != null)
```
ElderJames, line 28: use `?.` syntax

TimChen44: OK

PR #306, opened 2020-07-04
feat: improve range picker

mutouzdl:


Epictek:
still seems to be an issue when selecting two dates in the same month

mutouzdl:
> still seems to be an issue when selecting two dates in the same month, good work otherwise 👍

thanks, I try to fix it

PR #310, opened 2020-07-04
feat: add animating for button component when mouseup event fired

1002527441:


components/button/Button.razor
```diff
@@ -2,7 +2,10 @@
 @inherits AntDomComponentBase
 
 <CascadingValue Value="this">
-    <button class="@ClassMapper.Class" style="@Style" id="@Id" type="@HtmlType" @onclick="HandleOnClick" disabled="@Disabled">
+    <button class="@ClassMapper.Class" style="@Style" id="@Id" type="@HtmlType"
+            @onclick="HandleOnClick" disabled="@Disabled"
+            @onmouseup="OnMouseUp"
+            ant-click-animating-without-extra-node="@_animating.ToString().ToLower()">
```
ElderJames, line 8: Please use `@(_animating?"true":"false")` directly.

1002527441: updated

1002527441: updated!

PR #313, opened 2020-07-05
feat: allow clearing datepicker view

Epictek:
fix: some styling regarding range pickers

Wasn't sure how I was supposed to clear the datepicker value as the date can't be null unless we make it nullable?

Edit: Also starting to feel that DatePickerInput might need to be cleaned up a bit, it's starting to become a mess of nested if statements

### DIFF
--- a/components/date-picker/RangePicker.razor
+++ b/components/date-picker/RangePicker.razor
@@ -9,26 +9,27 @@
                     Disabled="Disabled"
                     PopupContainerSelector="@PopupContainerSelector"
                     OnVisibleChange="visible => AutoFocus = visible"
-                    OverlayEnterCls="slide-up-enter slide-up-enter-active slide-up"
-                    OverlayLeaveCls="slide-up-leave slide-up-leave-active slide-up"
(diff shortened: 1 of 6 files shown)


components/date-picker/RangePicker.razor
```diff
@@ -9,26 +9,27 @@
                     Disabled="Disabled"
                     PopupContainerSelector="@PopupContainerSelector"
                     OnVisibleChange="visible => AutoFocus = visible"
-                    OverlayEnterCls="slide-up-enter slide-up-enter-active slide-up"
-                    OverlayLeaveCls="slide-up-leave slide-up-leave-active slide-up"
+                    OverlayEnterCls="slide-up-enter slide-up-enter-active slide-up ant-picker-dropdown-range"
+                    OverlayLeaveCls="slide-up-leave slide-up-leave-active slide-up up ant-picker-dropdown-range"
```
mutouzdl, line 13: Maybe the "up" before "ant-picker-dropdown-range" should delete?

PR #327, opened 2020-07-07
Form context, IsModified property

Ratomir:
I exposed a method to trigger is form modified of the EditContext of the EditForm. 
So, still, I am not sure why is triggering at the second input trigger, in any field on the form, or if you type two characters in the same input filed.
@ElderJames    you can test this and check it, but I think that is related to the EditContext and EditForm of the AspNetCore.Components. I'll check it. 

Still, I am thinking that is a good time to expose this method and keep it as public in the Form class.

Kind regards,

### DIFF
--- a/components/form/Form.razor.cs
+++ b/components/form/Form.razor.cs
@@ -39,6 +39,8 @@ namespace AntDesign
         [Parameter]
         public EventCallback<EditContext> OnFinishFailed { get; set; }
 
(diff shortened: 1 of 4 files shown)


ElderJames:
Thank you for your contribution @Ratomir !

I really like this feature. Could you please also add it to the `IForm` interface? We want users to be able to manipulate the Form through this interface.

PR #328, opened 2020-07-07
DocScrape action

Epictek:
Action to add indices to DocSearch


ElderJames:
Awesome! How can I handle these two PR?

Epictek:
Just need to add secrets to the repo with `algolia_application_id` and `algolia_api_key`. Other PR would require you to modify the docsearch.js file to have your **public** api keys.

PR #340, opened 2020-07-10
try to fix #334

### LINKED ISSUE #334: input unusual  by chance
A community friend told me that there is a problem with the input control.
But i can't reproduce the problem right now.

Here is the gif of the problem(Provide by this friend):
input bug

mutouzdl:
Calling StateHasChanged while typing would cause this problem by chance. I don't know why yet.

Input's OnKeyUp function would call StateHasChanged. So i delete it.
After doing so, Input no longer support instant verification.


mutouzdl:
更准确地说，是在修改CurrentValue时，如果调用了StateHasChanged，有概率引起这个问题.
--------------------------------------------------------------------------------------------------
More precisely Calling StateHasChanged while modifying CurrentValue would cause this problem by chance.

PR #395, opened 2020-07-21
fix(module: upload): fix event listener

adsioj:
Disabled值改变时,重新监听或移除关联点击事件

### DIFF
--- a/components/upload/Upload.razor.cs
+++ b/components/upload/Upload.razor.cs
@@ -9,6 +9,10 @@ namespace AntDesign
 {
     public partial class Upload : AntDomComponentBase
     {
+        private bool _disabled;
+
+        private bool _disabledChanged;
+
         [Parameter]
         public Func<UploadFileItem, bool> BeforeUpload { get; set; }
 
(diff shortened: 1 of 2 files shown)


components/upload/Upload.razor.cs
```diff
@@ -86,7 +102,18 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
+            {
+                if (Disabled)
+                {
+                    await JSRuntime.InvokeVoidAsync(JSInteropConstants.removeFileClickEventListener, _btn);
+                }
+                else
+                {
+                    await JSRuntime.InvokeVoidAsync(JSInteropConstants.addFileClickEventListener, _btn);
+                }
+                _disabledChanged = false;
```
ElderJames, line 115:
```suggestion
            if (_disabledChanged)
            {
                _disabledChanged = false;
                if (Disabled)
                {
                    await JSRuntime.InvokeVoidAsync(JSInteropConstants.removeFileClickEventListener, _btn);
                }
                else
                {
                    await JSRuntime.InvokeVoidAsync(JSInteropConstants.addFileClickEventListener, _btn);
                }
```

PR #435, opened 2020-07-29
fix(module: input): incorrect OnPressEnter

ElderJames:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


components/input/Input.cs
```diff
@@ -75,6 +75,8 @@ public class Input<TValue> : AntInputComponentBase<TValue>
 
         public Dictionary<string, object> Attributes { get; set; }
 
+        private string _inputValue;
```
mutouzdl, line 78: Should we use TValue instead of string ?

ElderJames: You are right, I have fixed it.

PR #457, opened 2020-08-04
build(gulp): add .less source files as part of the output

diegofrata:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [X] Other (about what?)

### 💡 Background and solution

Currently the library does not pack the .less files, so there's no way for the users to reference variables or anything else that they may want to use as part of their development. The current workaround is to manually copy these into the project. With this change, the .less files are shipped alongside the minified CSS and JS, which can be used with plugins such as LigerShark.WebOptimizer.

For example, with this change I can potentially do this, or some other approach to using it:

```
@import '_content/AntDesign/less/style/color/colors.less';

.logo {
  color: @blue-1;
}
```

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Make .less files part of the library output under the less folder.    |
| 🇨🇳 Chinese |           |


ElderJames:
Thank you for your contribution @diegofrata .
As far as I know, the web assets files only exist in /_content after debugging or publishing. Can we import these files at compile time?

diegofrata:
@ElderJames you're right, it's only there after publishing, which is exactly my use case. I want to pre-process the files during runtime as opposed to compile-time! Checkout https://github.com/ligershark/WebOptimizer it does SASS and LESS pre-processing at runtime (as part of the first request) and makes the build pipeline much simpler to run as there's no extra steps for NPM and gulp.

ElderJames:
It seem pretty good. Can we leverage this to enable theme style switching at run time? Just like in the official documentation and in the Ant Design Pro preview site. At compile time, we can import from the NPM package.

image

image

diegofrata:
I am not familiar with how the theme switch is implemented in the main project, I would have to have a look. But the intention of this change is to make things such as color variables and mixins available at runtime so developers using runtime preprocessing tools such as LigerShark are not excluded. It only adds a few kilobytes to the final published app and loading these files is entirely optional.

ElderJames:
Ok, I'll merge this PR first and see what new features you can bring to us later, or another project that uses this feature.

Issue #461, opened 2020-08-05
[Feature Request] Carousel: add GoToSlide, Next, and Previous methods.


moonreign:
GoToSlide(slideIndex) : go to slide index
Next(), Previous(): move to next slide or previous respectively.

moonreign:
The following code should do it.

\components\carousel\**Carousel.razor.cs**

```
        public void Next() => GoTo(_slicks.IndexOf(_activeSlick) + 1);

        public void Previous() => GoTo(_slicks.IndexOf(_activeSlick) - 1);

        public void GoTo(int index)
        {
            if (index >= _slicks.Count)
            {
                index = 0;
            }
            else if (index < 0)
            {
                index = _slicks.Count - 1;
            }

            Activate(index);
        }
```

ElderJames:
Thank you for contacting us @moonreign .
This is a godd idea, and we welcome you to create a PR.

PR #462, opened 2020-08-05
feat(module:carousel): add new methods (#461)

moonreign:

### 🤔 This is a ...

- [X ] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


components/carousel/Carousel.razor.cs
```diff
@@ -49,6 +49,24 @@ public partial class Carousel : AntDomComponentBase
...
+        public void GoTo(int index)
+        {
+            if (index >= _slicks.Count)
+            {
+                index = 0;
+            }
+            else if (index < 0)
+            {
+                index = _slicks.Count - 1;
+            }
```
ElderJames, line 65: Why is this logic? Should it be the other way around?

moonreign: It's wrap around logic akin to how the Bootstrap carousel works: https://getbootstrap.com/docs/4.0/components/carousel/
When index is past the end of the list, it will move to first slide, and when user is trying to move backwards past the first slide it will move to the N-1 slide (last). 

The idea is to have constant flow when use uses Next() Preview() if that's not what the user want they can use Goto(i) with extra logic to for example not go past the bounds of the list, etc.

ElderJames:
Please add a demo.

moonreign:
> Please add a demo.

Demo added.

PR #472, opened 2020-08-07
chore: format code with ReSharper after push

Epictek:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Github action to format  code after every push, example: https://github.com/Epictek/ant-design-blazor/commit/1d62190316cc2535c8a019ec4e86f5537dac0435

We might want to tweak the default settings before merging?

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


components/form/FormItem.razor
```diff
@@ -6,8 +6,9 @@
     @if (!string.IsNullOrEmpty(Label))
     {
         <AntDesign.Col @attributes="GetLabelColAttributes()" class=@($"{_prefixCls}-label")>
-            <label class=@(Required ? $"{_prefixCls}-required" : _labelCls)> @(Label)</label>
-        </AntDesign.Col>
+        <label class=@(Required ? $"{_prefixCls}-required" : _labelCls)> @(Label)</label>
+        </AntDesign.
```
Epictek, line 10: Would seem like the formatter is breaking some of the Razor pages, will investigate this

diegofrata: ReSharper is absolute shit when it comes to formatting Razor files. Is there a way you can limit it to .cs files?

Epictek: That's a shame, I chose ReSharper for it's ability to format Razor files :( Guess I will investigate using dotnet format instead

Epictek:
Will close this for now and replace it with a [dotnet format](https://github.com/dotnet/format) prehook that only runs on .cs files

PR #476, opened 2020-08-08
feat(module: confirm): support blocking to get the result of confirm

zxyao145:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  support blocking to get the result of confirm  |
| 🇨🇳 Chinese |  支持阻塞（非委托/回调）获取Confirm弹窗的结果   |


components/modal/ModalRef.cs
```diff
@@ -92,5 +92,8 @@ public async Task UpdateConfig(ConfirmOptions config)
             await _service?.Update(this);
         }
 
+        internal TaskCompletionSource<bool> TaskCompletionSource { get; set; }
```
TimChen44, line 95: 此处bool修改成enum是否更好，便于将来的扩展。
Is it better to modify bool to enum here to facilitate future expansion. 
Such as:
```
    public enum ConfirmResult
    {
        None = 0,
        OK = 1,
        Cancel = 2,
        Abort = 3,
        Retry = 4,
        Ignore = 5,
        Yes = 6,
        No = 7
    }
```

TimChen44:
感谢，你的方法很好，其余类似功能可以参考你的方法进行改进。
Thank you, your method is very good, other similar functions can refer to your method for improvement.

components/modal/ModalRef.cs
zxyao145: For the confirm component, there are only only two types of buttons: OK or Cancel

ElderJames: Shall we put it in the generic `ModalRef<T>` ?

TimChen44: 如果是对于表单编辑关闭确认，就会存在“保存”，“不保存”，“取消”三个状态。
If it is to confirm the closing of the form editing, there will be three states of "Save", "Do not save" and "Cancel"。

PR #570, opened 2020-09-05
fix(module: table): fix collection modified exception

diegofrata:

### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [X] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

This change does a few things:

1) Fix the issue ``System.InvalidOperationException: Collection was modified; enumeration operation may not execute.`` reported by an user in the Slack channel. 

2) Removes a potenticial performance issue by avoiding multiple evaluations of a same IEnumerable when possible;

3) Introduces a way of recording the markup generated by a component at the end of a test so it can be compared against future changes.

The last item is a proposition on how we can compare results of tests without a ridiculous amount of effort -- all you need to do is run the test once to record the output and the next executions will be measured against it. Would be interested in hearing thouhgts about it.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/tests/table/TableTests.cs
+++ b/tests/table/TableTests.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.Collections.Generic;
+using Bunit;
+using Microsoft.AspNetCore.Components;
+using Xunit;
+
+namespace AntDesign.Tests.Table
+{
+    public class TableTests : AntDesignTestBase
+    {
+        class Person
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+            public string Surname { get; set; }
+        }
+
+        private IRenderedComponent<Table<Person>> CreatePersonsTable(
+            IReadOnlyList<Person> persons,
+            Action<ComponentParameterBuilder<Table<Person>>> callback = null,
+            bool enableSelection = false)
+        {
+            return Context.RenderComponent<Table<Person>>(x =>
+                {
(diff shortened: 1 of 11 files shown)


tests/table/TableTests.cs
```diff
@@ -0,0 +1,159 @@
...
+using Bunit;
+using Xunit;
+
+namespace AntDesign.Tests.Table
+{
+    public class TableTests : AntDesignTestBase
+    {
+        class Person
+        {
+<<<<<<< Updated upstream
```
ElderJames, line 12: Is there some conflict that hasn't been resolved?

ElderJames (changes requested):
Thanks for your contribution. Please resolve the git conflicts.

tests/table/TableTests.cs
diegofrata: I think I did something weird last night when trying to push. 😂 It's fixed now.

diegofrata:
@ElderJames the PR should be good now.

ElderJames (approved):
Thank you.

PR #694, opened 2020-10-20
tree enhancement

kefengwei:
### tree enhancement

1. change `TreeNode`'s `ChildNodes` accessbility to public
2. add `FindFirstOrDefaultNode` funcation
3. add `ExpandToNode` function

### DIFF
--- a/components/tree/TreeNode.razor.cs
+++ b/components/tree/TreeNode.razor.cs
@@ -392,10 +392,10 @@ namespace AntDesign
         /// <summary>
         /// 查找节点
         /// </summary>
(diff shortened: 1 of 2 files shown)


components/tree/TreeNode.razor.cs
```diff
@@ -389,6 +389,36 @@ private IEnumerable ChildDataItems
 
 
         #endregion
+        /// <summary>
+        /// 查找节点
```
ElderJames, line 393:
```suggestion
        /// Find a node
```

ElderJames (approved):
LGTM. Thanks.

Issue #761, opened 2020-11-04
Unable to clear Select component's value


Gareth064:
I want to press a button to reset the value in a Select component so that it returns to just showing the Placeholder text and not have an actual value, but it isn't working how I'd expect.

**Select Component**
```csharp
<Select Style="width: 200px" Placeholder="Select an Option" @bind-Value="app">
        <SelectOption Value="YES">Yes</SelectOption>
        <SelectOption Value="NO">No</SelectOption>
</Select>
```
**Button Component**
```csharp
<Button Type="primary" OnClick="Reset">Reset</Button>
```
**Code**
```csharp
@code{
    string app;

    private void Reset()
    {
        app = null;
    }
}
```
When the select component loads, it loads with just the placeholder.
I select an option.
When I press Reset and set the two-way binded 'app' variable to null, the selected option remains as it was.

How do I clear the value so it returns to just the placeholder again.

ElDiddi:
Hi @Gareth064,

this is a known issue in the 0.3.0. A basic ClearAll function is implemented in the nightly build but it's not perfect right now.

Just take a look at this [demo](https://ant-design-blazor.github.io/en-US/components/select#components-select-demo-clear-values).

Gareth064:
Hi ElDiddi

Brilliant, it's good to know it's on its way.

Thanks
Gareth
________________________________

Issue #727, opened 2020-10-25
bug(module: select): Custom tags cannot be created if a part of the tag lable is already in use


ElDiddi:
### Describe the bug
If a `select` is `Mode="tags"` custom tags cannot be created if a part of the tag lable is already in use.

### Steps to reproduce (please include code)
Demo - Automatic tokenization
1. Create a custom token with the value `test1`
2. Try to create a second token with the value `test` (this token cannot be created)

### Further technical details
- AntDesign Nuget 0.3.0

Issue #726, opened 2020-10-25
bug(module: select): Custom tags are not unchecked if they are cleared in the input field with the clear button


ElDiddi:
### Describe the bug
If a `select` is `Mode="tags"` and the `AllowClear` option is used, the clear button clears the input field, but custom tags are not unchecked.

@ElderJames 
I have a design question. Should the custom tags still appear in the selection list or should they be deleted like in the React implementation?

### Further technical details
- AntDesign Nuget 0.3.0

ElderJames:
We can delete them by default and allow user to save them in the options is they want.

Issue #724, opened 2020-10-25
[Select] Please add TValue for value type conversion


Yuanxw612:
The select value type is String, the actual business has other types, and you want to add TValue to specify the value type.

```
<Select TValue="int">
</Select>
```

@ElDiddi

ElDiddi:
Hello Yuanxw,

I think that `TValue` is difficult to implement in the current `select` module. Users can create their own tags, if `TValue` is an `integer`, which value should be set automatically? If you only allow `GUID` and `string` you can have the `GUID` generated automatically and if you allow `string` you can use the label value. Whereby `string` can generate an error if the `value` of the label corresponds to an existing value.

The easiest way would be to have two different modules, one that corresponds to a normal ComboBox with the usual WPF parameters like DataContext, SelectedItem, SelectedValue, ItemTemplate etc. and one that allows tags that only support `GUID` as value. Another approach would be to support only `GUID` in the current `select` module.

The current approach is for me at least a little bit difficult to maintain, because you have to pay attention to all combinations.

@ElderJames 
Do you have an idea? Maybe I think in the wrong direction.

Kind regards

Yuanxw612:
If the TValue doesn't fit,I think you can just use enumerations, @ElDiddi  @ElderJames  what do you think?
```
<Select ValueType="@SelectValueType.Guid">
</Select>
```

ElDiddi:
Hi @Yuanxw612,

sorry for the delay. I am working on this module intensively in my spare time. Maybe I will be able to send the PR by the weekend.

Yuanxw612:
@ElDiddi you did great!

ElDiddi:
Hey @Yuanxw612,

i send the PR for the select modul with alot of new features. Maybe you can test it and report some issues if you find any.
Take a look to the rewritten demos for this module, because the databinding has changed.

https://github.com/ant-design-blazor/ant-design-blazor/pull/786
https://github.com/ElDiddi/ant-design-blazor/tree/select-rewritten1

Yuanxw612:
@ElDiddi The complete rewrite of the SELECT component is exciting. i was touched by your contribution . good

PR #786, opened 2020-11-14
rewritten (module: select): Almost completely new

ElDiddi:
Hello everyone,

finally I am done, for now. 

!!! This PR is a Breaking Change !!!
Users who work with the select-module have to change their code after the merge.

Whats new?
- Datasource IEnumerable<TItem> - @bind-Value and @bind-Values
- Typeparams TItemValue, TItem
- Two-Way binding for the modes default, multiple and tag.
- OnDataSourceChange event when the DataSource changes
- Decoupling the SelectOptions from the data model, maybe this would help for virtualization later. (virtualization is not implemented atm)
- Keyboard support (Arrow keys up/down, Home, End & Enter for selection)
- Sorting for groups and labels (None, Asc, Desc)
- Reflection for Value, Label, Disabled, GroupName
- EventCallback OnSelectedItemChanged<TItem> and OnSelectedItemsChanged<IEnumerable<TItem>>
- Optional search for all three modes
- CreateCustomTag Action, here the developer can add his own logic
- LabelTemplate & ItemTemplate simplified
- IgnoreItemChanges to improve performance through Reflection

The demos and the API description are completely revised.

There were two problems in relation with a modal. The first error (popup container width) I could fix. The second error does not only affect the select-module but all modules that have integrated the overlay components.
If you switch between the web pages and then want to display an overlay within a modal, it will not be rendered. As soon as you reload the web page the overlay works again. There were already several issue entries.

@ElderJames 
It would be nice if you could take a look at the PR, maybe you could create a separate branch first. So the users can test the revision. The Chinese version of the documentation requires an adjustment.

There is certainly still plenty of room for improvement. 

- [x] New feature
- [x] Bug fix
- [x] Demo update
- [x] Component style update
- [x] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     Breaking change      |
| 🇨🇳 Chinese |    Breaking change      |

- [x] Doc is updated
- [x] Demo is updated
- [x] Changelog is provided


lhtuling:
image

ElderJames:
Wow thank you @ElDiddi ! You are amazing!

I will take a look at the PR. Can you help me see which [issues](https://github.com/ant-design-blazor/ant-design-blazor/issues?q=is%3Aissue+is%3Aopen+label%3A%22module%3A+select%22) this PR has fixed?

ElDiddi:
PR affected the following issues: 
DefaultValue of select works incorrect #235
[select]Related select elements displaying issue #473
[select] Default value issue #508
[SELECT] form binding issue - can't reset #515
关于 select 的问题 #695 <-- should be fixed, worked for me
[Select] Please add TValue for value type conversion #724
bug(module: select): Custom tags cannot be created if a part of the tag lable is already in use #727
bug(module: select): Custom tags are not unchecked if they are cleared in the input field with the clear button #726
Unable to clear Select component's value #761


There are some requests for OnChange, OnSelect, OnRemove events, I will take a closer look at this in the next few days.

Yuanxw612:
@ElDiddi You are amazing!  love you

components/select/SelectOption.razor.cs
```diff
@@ -1,169 +1,173 @@
...
             await base.OnInitializedAsync();
         }
 
-        protected override void OnInitialized()
+        protected void SetClassMap()
         {
+            ClassMapper.Clear()
+                .Add("ant-select-item")
+                .Add(ClassPrefix)
+                .If($"{ClassPrefix}-disabled", () => IsDisabled)
```
ElderJames, line 124: We can use the `GetIf` to do the string concatenation each time. And no need to call `SetClassMap()` so many time.
```suggestion
                .GetIf(()=>$"{ClassPrefix}-disabled", () => IsDisabled)
```

ElderJames:
No problems in general.
But for the docs and demos, I hope to cover as many functions as possible against the ant design react.

ElDiddi:
> No problems in general.
> But for the docs and demos, I hope to cover as many functions as possible against the ant design react.

After the merge I planned to rework the demos a bit. For example the whole buttons could be replaced by a drop down.
At least the RemoteData demo never worked for me, I could put it back in. The BigData demo with 36 items could be included as soon as virtualization is implemented.

components/select/SelectOption.razor.cs
ElDiddi: Hi James,

i tried this: 
```
            ClassMapper.Clear()
                .Add("ant-select-item")
                .Add(ClassPrefix)
                .GetIf(() => $"{ClassPrefix}-disabled", () => IsDisabled)
                .GetIf(() => $"{ClassPrefix}-selected", () => IsSelected)
                .GetIf(() => $"{ClassPrefix}-active", () => IsActive)
                .GetIf(() => $"{ClassPrefix}-grouped", () => SelectParent.IsGroupingEnabled);
```

I need a StateHasChanged after the properties IsSelected, IsDisabled, IsHidden, IsActive. Otherwise there are display issues.

Like this:

```
        public bool IsSelected
        {
            get => _isSelected;
            set
            {
                if (_isSelected != value)
                {
                    _isSelected = value;
                    StateHasChanged();
                }
            }
        }
```

What is the benefit to use GetIf? I can't see the difference. :s

components/select/SelectOption.razor.cs
```diff
@@ -121,10 +121,11 @@ protected void SetClassMap()
             ClassMapper.Clear()
                 .Add("ant-select-item")
                 .Add(ClassPrefix)
-                .If($"{ClassPrefix}-disabled", () => IsDisabled)
-                .If($"{ClassPrefix}-selected", () => IsSelected)
-                .If($"{ClassPrefix}-active", () => IsActive)
-                .If($"{ClassPrefix}-grouped", () => SelectParent.IsGroupingEnabled);
+                .GetIf(() => $"{ClassPrefix}-disabled", () => IsDisabled)
```
ElderJames, line 124: Sorry, you don't need `GetIf` in this case. I was wrong.

PR #787, opened 2020-11-15
feat(modul: menu) - MenuItem Title

ElDiddi:
Hi there,

a little PR for the menu modul. 

What's new?
+ MenuItem - `[Property] Title` - Allows to set the Title directly as parameter and not only as ChildComponent.

Changes:
+ Menu - `[Property] Selectable` - Angluar implementation adapted.

EN-US Documentation updated.

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
-

### 💡 Background and solution
-

### 📝 Changelog
-

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

- [x] Doc is updated

### DIFF
--- a/components/menu/Menu.razor.cs
+++ b/components/menu/Menu.razor.cs
@@ -50,6 +50,9 @@ namespace AntDesign
         public bool Selectable { get; set; } = true;
 
         [Parameter]
+        public bool Multiple { get; set; }
+
+        [Parameter]
         public bool InlineCollapsed
         {
             get => _inlineCollapsed;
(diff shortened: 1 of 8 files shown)


ElderJames:
Hi @ElDiddi , thank you for your active participation!

But I think `IgnoreSelectionAfterClick` isn't used very often, can we do that when it is only in the dropdown?

ElDiddi:
Hello James,

that is a good question. 

I think that the option is also useful in a normal menu. For example if you open an external web page in a new tab from a menu item. In that case the menu item would be selected even though you are still on the old page.
However, it would make more sense if that property is set per menu item and not globally in the menu. I could quickly adjust that.

ElderJames:
Yes, you are right. We can also take a look into angular's routerlink implement.

ElDiddi:
> Yes, you are right. We can also take a look into angular's routerlink implement.

Hehe i can't read the React/Angular source code. Thats why i try to use Blazor :D

ElderJames (approved):

Issue #518, opened 2020-08-19
OnSelect and OnDeselect event couldn't fire on Select element


justmine66:
```
<Modal Title="@("使用已有类目")"
                                       Visible="@_oldCateroriesVisible"
                                       OnOk="@HandleOkAsync"
                                       OnCancel="@HandleCancel"
                                       CancelText="@("取消")"
                                       OkText="@("确定")">
                                    <Select Style="width:100%" Mode="multiple" AllowClear ShowSearch ShowArrow
                                            Placeholder="请选择已有类目"
                                            OptionLabelProp="label"
                                            OnSelect="OnOldCategorySelect"
                                            OnDeselect="OnOldCategoryDeselect">
                                        @foreach (var item in OldCategories)
                                        {
                                            <SelectOption Label="@item.title" Value="@item.key">@item.title</SelectOption>
                                        }
                                    </Select>
                                </Modal>
```
```private void OnOldCategorySelect(OneOf<string, LabeledValue> item, SelectOption option)
        {

        }

        private void OnOldCategoryDeselect(OneOf<string, LabeledValue> option)
        {

        }
```

anddrzejb:
@justmine66 Is there a chance to check this again? It seems to me that it is working with the new design. If not, can you provide a repro?

anddrzejb:
@ElderJames I think this can be closed...The mentioned methods `OnSelect` and `OnDeselect` do not exist anylonger. I believe they were removed in favour of single method `OnSelectedItemChanged` of the `Select` component.

Issue #695, opened 2020-10-20
关于 select 的问题


lyjcooi:
我在使用 select 组件的时候，我把官方文档的里面的代码复制上去，放在页面中，项目启动起来的时候就会提示“JavaScript interop calls cannot be issued at this time. This is because the component is being statically rendererd. When prerendering is enabled, JavaScript interop calls can only be performed during the OnAfterRenderAsync lifecycle method.” 异常，我想知道这是什么原因

mutouzdl:
@ElderJames render-mode改为ServerPrerendered后，还是没法重现

`    <app>
        <component type="typeof(App)" render-mode="ServerPrerendered" />
    </app>
`

Issue #261, opened 2020-06-28
Keyboard navigation


Zoxive:
I noticed while going over the online examples that keyboard navigation appears to be missing in most of the components.

For example https://ant-design-blazor.github.io/en-US/components/select if you click on the first select example and try using the arrow keys up/down nothing happens. Also pressing the tab key should move to the next input, but nothing happens. (or shift tab to move backwards)

The react equivalent components (https://ant.design/components/select/) has the correct keyboard navigation behavior.\

I didnt see it mentioned anywhere on any of the roadmaps, just wanted to see if there was a timeline for adding it.

ElderJames:
Thank you for contacting us @Zoxive .
This actually a issue.I think this is caused by the loss of tabindex. We will fix it as we adapt the form controls, in next milestone.

Issue #235, opened 2020-06-19
DefaultValue of select works incorrect


zzzzzz7946:
```
<Select  DefaultValue=@(defaultValue) >
    <SelectOption Value="1">Jack</SelectOption>
    <SelectOption Value="2">Lucy</SelectOption>
    <SelectOption Value="3" Disabled>Disabled</SelectOption>
    <SelectOption Value="4">YaoMing</SelectOption>
</Select>

public string defaultValue{ get; set; }
 protected override async Task OnInitializedAsync()  {
   await Task.Delay(3000);
        defaultValue = "4";
}
```
这种情况下 select的默认选中值就设置不了了

mutouzdl:
DefaultValue只能生效一次，后续再更改是不会生效的，虽然这里用了delay，但实际上我发现，Select的初始化在OnInitializedAsync执行完之前已经执行了。


----------------translate by bing----------------
Default Value can only take effect once, and subsequent changes will not take effect, although delay is used here, but in fact I found that the initialization of Select was performed before OnInitialasync was executed.

PR #800, opened 2020-11-17
rewritten (module: select): Almost completely new - update 2

ElDiddi:
_GitKraken didn't want to open my repository anymore, after rebuilding the repository Github doesn't recognize the old PR or didn't want to append the PR to the existing one. Therefore a new PR._

!!! This PR is a Breaking Change !!!
Users who work with the select-module have to change their code after the merge.

Whats new?
- Datasource IEnumerable - @bind-Value and @bind-Values
- Typeparams TItemValue, TItem
- Two-Way binding for the modes default, multiple and tag.
- OnDataSourceChange event when the DataSource changes
- Decoupling the SelectOptions from the data model, maybe this would help for virtualization later. (virtualization is not implemented atm)
- Keyboard support (Arrow keys up/down, Home, End & Enter for selection)
- Sorting for groups and labels (None, Asc, Desc)
- Reflection for Value, Label, Disabled, GroupName
- EventCallback OnSelectedItemChanged and OnSelectedItemsChanged<IEnumerable>
- Optional search for all three modes
- CreateCustomTag Action, here the developer can add his own logic
- LabelTemplate & ItemTemplate simplified
- IgnoreItemChanges to improve performance through Reflection
- The demos and the API description are completely revised.

-- Update 2 --
- Simplification of the upgrade path ***
- Re-Added the property LabelInValue but it generates a functional Json string that you can parse with `Newtonsoft.Json` or `System.Text.Json`. It is only available when the SelectOption created without a DataStore. If you use the DataStore binding, you get the `TItem`.
- modified the CalendarHeader
- Prettified some demos
- Added the namespace AntDesign.Select.Internal for all classes which the user does not need.

*** Simplification
If you do not want to create the data binding using a DataSource, you can now use the ChildComponent `<SelectOptions></SelectOptions>` and create the `SelectOption` directly. This should avoid unnecessary source code which could be created by a DataSource binding. 
```
<Select @bind-Value="@_selectedValue"
        TItemValue="int"
        TItem="string">
	<SelectOptions>
                <SelectOption TItemValue="int" TItem="string" Value=1 Label="Jack" />
                <SelectOption TItemValue="int" TItem="string" Value=2 Label="Lucy" />
                <SelectOption TItemValue="int" TItem="string" Value=3 Label="Yaoming" Disabled />
                <SelectOption TItemValue="int" TItem="string" Value=4 Label="Frieda" />
	</SelectOptions>		
</Select>
```

** Note ***
Don't use a inline initialization of the Datastore. Like this:
```
<select DataSource="@(new List<MyClass> { new MyClass { Id=1, Name="Lucy" } })"
```

There is certainly still plenty of room for improvement.
The documentation for the demos needs to be translated into Chinese.

-- Update 4 --
- Added the old demos again to make the demos comparable to React/Angular.
- Changed some PropertyNames 
- Fixed a bug which triggered the OnSelectedItemsChanged event twice.

### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

PR affected the following issues:
DefaultValue of select works incorrect #235
Keyboard navigation #261
[select]Related select elements displaying issue #473
[select] Default value issue #508
[SELECT] form binding issue - can't reset #515
关于 select 的问题 #695 <-- should be fixed, worked for me
[Select] Please add TValue for value type conversion #724
bug(module: select): Custom tags cannot be created if a part of the tag lable is already in use #727
bug(module: select): Custom tags are not unchecked if they are cleared in the input field with the clear button #726
Unable to clear Select component's value #761

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElDiddi:
GitHub tease me, i haven't check in the `components/AntDesign.csproj`

ElderJames:
You should rebase the master branch.

Yuanxw612:
@ElDiddi 

After reading your writing method, I am a little confused about why TItemValue and TItem should be used at the same time, and SelectOption should also be used like this.I don't think it makes sense.You can do this entirely with a parameter, such as ValueType

Regarding the use of value types, I suggest
```
<Select @bind-Value="@_selectedValue" ValueType="int">
	<SelectOptions>
                <SelectOption Value=1 Label="Jack" />
                <SelectOption Value=2 Label="Lucy" />
                <SelectOption Value=3 Label="Yaoming" Disabled />
                <SelectOption Value=4 Label="Frieda" />
	</SelectOptions>		
</Select>
```

ElDiddi:
Hey @Yuanxw612,

you have to specify the TItemValue and TItem explict for each SelectOption otherwise a compilation error will occur. I have not yet figured out how to pass TItemValue and TItem from the parent element to avoid the explicit naming. If I have tracked the iusses on github/aspnetcore correctly the typeparam cannot be derived at the moment. It was requested as a feature. 
I can not use TValue because one of the AntDesign Base classes converts the type of TValue to string. Which results in the fact that TValue would always be string and you can save yourself the hassle. (thats a internal AntDesign issue)
If you look at commercial 3rd party component manufacturers, there is no one who offers the flexibility to choose between the direct initialization of SelectOption and a binding via DataSource. On github/aspnetcore this feature for the Select component was requested about a year ago and has not yet been implemented.

TimChen44:
@ElDiddi 
检查演示时发现一个问题。
A problem was found when checking the demo.

"自动分词"的Demo在复制带有”，“符号的文本后，组件并没有正确分词。
After copying the text with the symbol of "," in the Demo of "Automatic tokenization", the component did not segment correctly.

ElderJames:
Hey @ElDiddi , has #695 been fixed?

ElderJames:
@ElDiddi I think you may fixed #518 in this PR. Could you confirm?

ElderJames:
In demo multiple selection, the display order of the selected items have to be the order of selection, not the order of options.

currently:
image


expected:
image

ElderJames (approved):
Merge this PR first and fix the rest through others.

Issue #793, opened 2020-11-16
[FeatureRequest] Set first day of the week for DatePicker based on Localization


NPadrutt:
### Is your feature request related to a problem? Please describe.
Currently the first day of the week is always shown as Sunday in the Datepicker. But a lot of Localizations it's Monday (for example German Localizations).

### Describe the solution you'd like
An solution would be to add a value to the localization, based on which afterwards in the `DatePickerDatetimePanel.razor ` the first day is set. Since there is already the list with the day headers (Su, Mo, Tu, etc) on the Localization and that would have to adjust as well this might be the easiest way. 

Based on what I saw the diff currently is always set to 7 for Sunday. This would have to be 6 for Monday. When we can pass down the Calendar as a Cascading Parameter we could get that via the Localization. If that is a possible solution I could create a PR for that.

PR #803, opened 2020-11-18
feat (module:datepicker) Add FirstDayOfWeek option and german localization

NPadrutt:
(Unfortunately I didn't find out how to create the PR to a feature branch. I think I don't have the permission to do that)

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#793 

### 💡 Background and solution

1.  Currently in the DatePicker Sunday is always the first day of the week. This should be easy configurable, since in a lot of reginos this differs.
2. Default:
image
FirstDayOfWeek set to Monday:
image
3. First there is a new Parameter on the DatePicker. Instead of fix calculating the difference for sunday in the DatePickerDatePanel, I added a helper which returns the correct constant based on the passed day.
To get the correct header name (like "Su", "Mo", etc.) I made another helper method which returns an array with the correct order and localization based on the passed culture. Since the localization here is provided by .net directly the manual localization is no longer necessary and I removed those from the `IDateLocal` interface.
Also I added an `IDateLocal`  implementation for Swiss German and return this for swiss german, austrian german and germany german since they are identical.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add a new option "FirstDayOfWeek" to the DatePicker. Localization for de-CH, de-DE and de-AT |
| 🇨🇳 Chinese |           |


ElderJames:
Hi @NPadrutt , thank you for your great contribution.

I think you may need to modify it further. I want FirstDayOfWeek to be an attribute of i18n.

Currently, date picker does not use the new i18N implementation that use [locale configs](https://github.com/ant-design-blazor/ant-design-blazor/tree/master/components/locales) and the [`LocaleProvider`](https://github.com/ant-design-blazor/ant-design-blazor/tree/master/components/locale-provider).

NPadrutt:
hey @ElderJames 

uff, I see. I'm happy to adjust the implementation to use the new locale system. But I think refactor the whole l18n implementation for the datepicker is a bit out of the scope of this PR.

ElderJames:
This is a good news!

Sure, you can feel free to force push to this PR or Create another one if you wish.

NPadrutt:
What I mean is, that I'm not sure that I have the time to dedicate to do the whole refactoring, since it looks like something that could take quite a bit, especially for a first timer / someone never worked with the current locale system.

NPadrutt:
I tried something anyway. What I couldn't find out is how to rerender the picker panel when the language changes. Currently the header changes but the actual first day of the week still is the one first loaded.

Furthermore this is now a small breaking change, since the CultureInfo Parameter on the DatePicker is removed, since that is now loaded via the LocaleProvider.

Last but not least, there are around 10 localization and formats missing in the json files I had to add. Those would have to be added to all other json files as well, together with the new "firstDayOfWeek" field. For the latter I have a default to Sunday.
What I noticed during this change, is that pretty much all json's do have invalid format values. For example the en-GB:

      "dateFormat": "D/M/YYYY",
      "dayFormat": "D",
      "dateTimeFormat": "D/M/YYYY HH:mm:ss",

I think this would have to be:

      "dateFormat": "d/M/yyyy",
      "dayFormat": "d",
      "dateTimeFormat": "d/M/yyyy HH:mm:ss",

At least the "YYYY" in uppercase is not recognized by .net and just printed.

ElderJames:
Wow, amazing!

Since the time formatting string is what I copied from the react side, we needed to adjust it to the FORMAT supported by .NET. 
And the `firstDayOfWeek` is need to be added in locales. You can adjust it for.NET first.

NPadrutt:
> And the `firstDayOfWeek` is need to be added in locales.

Do you mean with that the new field I added? 
image

Or do you mean something I would have to adjust? 

> Also, I see that you're using LocaleProvider.CurrentLocale.DatePicker directly. But I think that is better to add a Locale property that default to LocaleProvider.CurrentLocale.DatePicker. Like what we did in table:
> 
>         [Parameter]
>         public TableLocale Locale { get; set; } = LocaleProvider.CurrentLocale.Table;
> Then users can pass in a custom Locale object.

Ah, I see. I'll adjust that.
Any idea on the rerender topic?

ElderJames:
We can set the type of `firstDayOfWeek` as number, it's more appropriate than using English.
The `shortWeekDays` also need to add into locale object.

NPadrutt:
I would advice against a number. Otherwise you either have to parse it everytime you use it to the DayOfWeek enum which can become a performance issue or work with magic numbers in the code which does make it rather error prone and not easy to work with. Since c# is english based I think english is a rather good compromise.

The shortWeekDays a generated based on the current locale via the .net framework, so I don't see why we would want to manually localize those.

ElderJames:
Alright, just do as you say 😄.

NPadrutt:
I adjusted the access to the DatePickerLocale as you suggested.
Not sure if that solved the rerender issue on changing the language or if it just doesn't occur in WASM, but so far I can't reproduce it anymore on my device. 

Let me knowm if the rest is alright the way it is. Then I would go on to adjust the demos and docs.

NPadrutt:
Ah, what comes to mind.. I currently access the current CultureInfo via the LocaleProvider. Should there be a parameter analog to Locale, so people can override that?

ElderJames:
Isn't there a `locale` field? We can use it to convert to `CultureInfo`.

NPadrutt:
The field uses the DatePickerLocale analog to Table. The CultureInfo is on the Locale level itself.
I commit a suggestion in a second, the you can see if you like it that way.

NPadrutt:
@ElderJames alright, I readded the `CultureInfo` field. I think in any way that makes sense, since A) with that it's no longer a breaking change and B) you can easily overwrite it. Otherwise we would have to use the `Locale` instead of `DatePickerLocale` although we don't use like 90% of `Locale` which would make it unintuitive to use.

I also adjusted the demo and api doc. So from my side we would be good to go, unless you have concerns to address.

ElderJames:
Thank you very much, @NPadrutt.

I wilreview this PR soon. It make a great sense for this project.

ElderJames:
Here is some locales else used in ant design. Could you please adjust the field name?

https://github.com/react-component/picker/tree/master/src/locale

NPadrutt:
@ElderJames I'm sorry I don't understand. What field name would you like me to adjust? Also, did you mean to send me a link to the react locale files instead of the ones in this project?

ElderJames:
I mean that make our field names the same as theirs

NPadrutt:
the new fields I added are not present in the files you sent me as far as I can see. But they emphasize the same naming convention.

ElderJames:
When running locally, I found some problems:
1. The placeholder of rang pickers doesn't work in Chinese.
2. In Chinese data picker, the first day of week should be Monday. However, when I set "Monday", the day of the fist date is wrong.

components/calendar/internal/CalendarHeader.razor
```diff
@@ -16,7 +16,7 @@
             endYear = Calendar.ValidRange[1].Year;
         }
 
-        string yearFormat = Calendar.CultureInfo.GetDateLocale().YearFormat;
+        string yearFormat = LocaleProvider.CurrentLocale.DatePicker.Lang.YearFormat;
```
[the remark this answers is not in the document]
NPadrutt: What do you mean with the local parameter as well as date picker?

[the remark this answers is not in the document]
NPadrutt: oh actually, this change is a mistake in general. that is in the calendar module which was not adjusted on purpose.

NPadrutt:
To your first point see my [comment above](https://github.com/ant-design-blazor/ant-design-blazor/pull/803#issuecomment-730382171):

> Last but not least, there are around 10 localization and formats missing in the json files I had to add. Those would have to be added to all other json files as well, together with the new "firstDayOfWeek" field. 

I can set a default in the `DatePickerLocal` class to english for all the values, so that we never just have an empty field if that helps.

To the second see here in the same comment, I think that this is related to what I wrote here:

> What I couldn't find out is how to rerender the picker panel when the language changes. Currently the header changes but the actual first day of the week still is the one first loaded. If you have a solution or idea for this please let me know.

and as a follow up [here](https://github.com/ant-design-blazor/ant-design-blazor/pull/803#issuecomment-730410257):

> Any idea on the rerender topic?

and [here](https://github.com/ant-design-blazor/ant-design-blazor/pull/803#issuecomment-730444759):

> Not sure if that solved the rerender issue on changing the language or if it just doesn't occur in WASM, but so far I can't reproduce it anymore on my device.

From what I see if you are running the site in the Blazor WASM version this problem doesn't occur since the panel is rerendered. If you have any input on that I'm still very interested to hear it, since I'm out of ideas here.

components/calendar/internal/CalendarHeader.razor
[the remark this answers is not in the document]
NPadrutt: I added the same parameters on the calendar as on the date picker and replaced the static calls with those.

NPadrutt:
Maybe to add on the second point: the probably is the dynamically changing of the language. If you have them set during the first render it works as expected. For example you can see that in the demo in the localization part. I adjusted that one to use en-US with first day Sunday and chinese with first day monday.

ElderJames:
Thank you @NPadrutt , 

When I set `firstDayOfWeek` is "Monday", the day of week is right but the 1st in this month is wrong.

Currently:
image

Expected:

image

NPadrutt:
@ElderJames Oh, I'm sorry, I misunderstood you before and thanks for pointing that out. It might be that I mixed up the dates, during my testing but I will have a look at it once I'm back home. 
If you see something else, let me know, then I can have a look at all points at once.
In terms of the new fields I had ti add to the locale file, is a default in the DatePickerLocale class an acceptable solutions or do you have something else in mind?

ElderJames:
> In terms of the new fields I had ti add to the locale file, is a default in the DatePickerLocale class an acceptable solutions or do you have something else in mind?

I think it is Ok :-)

NPadrutt:
Using an enum actually was my original plan, but the JsonSerializer throw an exception when it tried to deserialize the value to an enum. But I actually just found a solution for that. So I will adjust that aswell.

NPadrutt:
@ElderJames I just checked the issue you described. It seems that is something off when you have the first day of the week set by the localization file. When I set it via a manually created Locale object it works as expected as far as I can see (see Globalization demo). 

image
image

It seems there is also a bit of a difference between WASM and Server Side. I'll investigate.

NPadrutt:
@ElderJames  could you have another look? I think I was able to fix the problem.

ElderJames:
All good, only one needs improvement is that the `ShortWeekDays` may have to config in locale jsons. 
Because not every language needs two letters. For example, Chinese and Japanese, only one is needed.

NPadrutt:
I see. The fun of localization ^^
The modification you requested are made. I also merged the current changes from the master to avoid potential merge conflicts.

components/date-picker/types/DatePickerPlaceholder.cs
```diff
@@ -1,38 +1,34 @@
...
-                DatePickerType.Time => (info.GetDateLocale().StartOfDate, info.GetDateLocale().EndOfDate),
-                DatePickerType.Quarter => (info.GetDateLocale().StartOfQuarter, info.GetDateLocale().EndOfQuarter),
-                _ => (info.GetDateLocale().StartOfDate, info.GetDateLocale().EndOfDate),
+                DatePickerType.Date => (locale.Lang.StartDate, locale.Lang.EndDate),
+                DatePickerType.Week => (locale.Lang.StartWeek, locale.Lang.EndWeek),
+                DatePickerType.Month => (locale.Lang.StartMonth, locale.Lang.EndMonth),
+                DatePickerType.Year => (locale.Lang.StartYear, locale.Lang.EndYear),
+                DatePickerType.Time => (locale.Lang.StartDate, locale.Lang.EndDate),
+                DatePickerType.Quarter => (locale.Lang.StartQuarter, locale.Lang.EndQuarter),
+                _ => (locale.Lang.StartDate, locale.Lang.EndDate),
```
ElderJames, line 31: Why not use the existing `RangePlaceholder`,`RangeYearPlaceholder`,`RangeMonthPlaceholder` and `RangeWeekPlaceholder`?

ElderJames: We may need to add `RangeQuarterPlaceholder`

NPadrutt: Simply because I overlooked them.

NPadrutt: Simply because I overlooked them.
Also, maybe it's just me, but using an array of strings for localization feels a bit weird in C# since you would have to access it afterwards in code with the index which in my opinion is rather intuitive. So far at least from the code references in VS they are used nowhere so we still can go either way.

NPadrutt:
As a constructive feedback: I really like the detailed review and feedback, please don't get me wrong. But it would make the review process much more efficient and less cumbersome if change requests would come more or less together. Most or all change requests came one by one each on a different day and after I submitted a change for the previous. With the respective code being present since at least the moment when I refactored the localization of the date picker.
Again, don't want to blame someone, it's just thought as a constructive feedback to improve the project.

ElderJames:
I'm sorry I made you feel uncomfortable @NPadrutt .
Your work has made a great difference. Thank you very much.

So far, this PR has been fine.

ElderJames (approved):

NPadrutt:
@ElderJames no no, you didn't make me uncomfortable. I just wanted to give a feedback to where I could see the review process being more efficient. :) 
Your work with the detailed reviews and feedbacks and your work with this library in general is highly appreciated! :)

ElderJames:
Well, looking forward your contributions in our project :)

PR #804, opened 2020-11-18
Chart doc

NPadrutt:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

When I used the chart library I stumbled about a couple of roadblocks. I tried to update the documentation so that those are are described.
Also, I translated some of the code examples to english. Since there seems to be no localization for the examples in the chart demo I think this makes it more accessible and overall easier to use.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames (approved):
LGTM.

Issue #772, opened 2020-11-09
IconService is using address from Base tag to retrieve icons.json file


Ttxman:
My application is running behind reverse proxy and the address is being redirected co custom endpoint `https://externaladdress/serviceEndpoint/.....`
I am running ant-design-pro-blazor as server side blazor app.

Our firewall configuration is very strict and do not allow the webserver to reach outside of localhost and few whitelisted addresses.

My problem is that SVG icons are not showing in this scenario.

When digging through logs I found that webserver is trying to reach `https://external-address/service-endpoint(proxy)/_content/AntDesign/icons/icons.json`
I traced the address origin to base tag specified in `_Host.cshtml`, where we have absolute address (`https://externaladdress/serviceEndpoint/`) to make everything work through proxy.

And after more digging through your source code i found the culprit: https://github.com/ant-design-blazor/ant-design-blazor/blob/87df452c256c093f74504830f3455b091a35a90b/components/icon/IconService.cs#L74

This service get injected NavigationManager which somehow loads the base address specified in base tag and then intend  to make http request from webserver to itself to reach the icon list. But the request fails misserably because public app address is not reachable from the webserver.

Issue #668, opened 2020-10-05
Improve perfomance of loading icons


Epictek:
The current implementation of the Icons requires uses HttpClient to retrieve the icons when requested, this makes the icons slow to load compared to other assets on the page.

I would propose bundling the icon svgs as individual components using a gulp file similar to the ant-design icons repos.

This would increase the bundle size but with [compression](https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-3.1#compression) shouldn't be a big deal. initial testing I got it down to 116kb which is pretty good considering there a JS date libaries bigger than that... and with some linker magic it might be possible to make it smaller?

![](https://user-images.githubusercontent.com/1466270/95122283-2b464f80-0748-11eb-80b1-628748577ad1.png)

ElderJames:
It seems to make sense, and we can try it later. Thanks for the researching, @Epictek .

gabrielfreire:
Icon loading is terribly slow indeed

zxyao145:
Can some commonly used icons be written directly into the Blazor as string, instead of being loaded through Ajax?

PR #810, opened 2020-11-19
PoC(module: iconservice) - Direct loading without HttpWebClient

ElDiddi:
It is only a little PoC of how to customise the IconService so that the icons can be loaded directly and without HttpWebClient.

Therefore I created a new static class, in which the icons are listed. 

As @Epictek has already written, the compressed assembly will be 160-200kb larger. 

This covers following issues:
#772
#668


Epictek:
how is the Icon store generated, Can it be automated?

ElDiddi:
Hey @Epictek,

i have written a small external C# application which scans the icon folders and generates the class file code.

ElderJames:
We should generate icons form [ant-design](https://github.com/ant-design/ant-design-icons/tree/master/packages/icons-svg/svg). How about using Github Action like ours styles-sync-bot?

ElderJames:
Hey @Epictek , what do you think of this PR?

Epictek:
Think we need the script used to generate the class before merging.

I also made something similar a while ago but with Razor files and was a very much WIP

https://www.github.com/Epictek/ant-design-icons/tree/blazor/packages%2Ficons-blazor%2Fscripts%2Fgenerate.ts

ElderJames:
If we decide to embed it in the assembly, then I think we can directly generate the icon components as the same to ant-design react.
 
e.g. `<UserOutlined />`

Epictek:
Sounds good to me :)

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

ElderJames (approved):

Issue #865, opened 2020-12-07
Performance issue for slider


anddrzejb:
### Describe the bug
There is a part of code that checks if a new value provided for the slider is different than previous. This check is meaningless as the update is happening basically always (probability of running an update is probably around 99.99% - the percentage value isn't really calculated, just guessed 😉 ), for 2 reasons:
- the values that are delivered are of type ```double``` and it is highly unlikely a user will click in exactly the same pixel; and since a user is going to click most likely into a different pixel, then the check most likely will give a false positive
- once the value change procedure starts running, the values are anyway rounded to a step representation; so event without moving a mouse and continuously clicking in exactly the same pixel, the update will most likely anyway happen (it is very unlikely for a user to click on a pixel that will evaluate to exact step value).

### Steps to reproduce (please include code)
Add debug statements to ```Slider.razor.cs``` to the ```LeftValue``` property setter:
```csharp
            set
            {
                if (_leftValue != value)
                {
                    Debug.WriteLine($"Old left value: {_leftValue}, new delivered value: {value}");
                    _leftValue = Math.Max(value, Min);
                    _leftValue = Math.Min(_leftValue, RightValue);
                    _leftValue = GetNearestStep(_leftValue);
                    Debug.WriteLine($"New left value: {_leftValue}, new delivered value: {value}");
                    SetStyle();
                    //CurrentValue = TupleToGeneric((_leftValue, RightValue));
                    CurrentValue = DataConvertionExtensions.Convert<(double, double), TValue>((_leftValue, RightValue));
                }
            }
```
Create simple component with slider
```razor
<Slider TValue="(double, double)" , Min="@_testMin" Max="@_testMax" Value="(_start1, _end1)" />
@code {
    private double _testMin = 0;
    private double _testMax = 60;
    private double _start = 10;
    private double _end = 30;
}
```
Start clicking and observe the debug output window:
![slider_samePoint_recalculation](https://user-images.githubusercontent.com/6518006/101337468-6407bf80-3884-11eb-94f6-752c83325e81.gif)

You will notice that even though the same point is clicked in every instance (9.7979797979798), slider evaluates this is still a new point.
### Further technical details
- AntDesign commit 58d29af (Dec 5, 2020)
- Running net5.0
- Visual Studio 16.8.2

Issue #864, opened 2020-12-07
Range slider not updating properly when both values change at the same time


anddrzejb:
### Describe the bug
If a ranged slider has ```Values``` bound to 2 variables (of ```double``` type of course), and both variables are changed significantly, then only one edge is updated.

### Steps to reproduce (please include code)
```razor
<Slider TValue="(double, double)" , Min="@_testMin" Max="@_testMax" Value="(_start, _end)" />
<Button OnClick="TestClick">Change both variables (@_start1 - @_end1) </Button>
@code {
    private double _testMin = 0;
    private double _testMax = 60;
    private double _start = 10;
    private double _end = 30;

    private void TestClick()
    {
        _start1 = 40;
        _end1 = 58;
    }
}
```
The result is (notice 2 separate clicks on the button to achieve the expected result):
![slider_2variable_change_err](https://user-images.githubusercontent.com/6518006/101334620-c8288480-3880-11eb-9ea2-424f99bf1585.gif)

Expected result: both edges should be updated (notice single click)
![slider_2variable_change_expected](https://user-images.githubusercontent.com/6518006/101334753-f4dc9c00-3880-11eb-89ec-de490c808571.gif)

### Further technical details
- AntDesign commit 58d29af (Dec 5, 2020)
- Running net5.0
- Visual Studio 16.8.2

Issue #859, opened 2020-12-05
Double-edged slider will lock on left edge only after first usage.


anddrzejb:
### Describe the bug
Cannot move right edge once left edge was moved. 

### Steps to reproduce (please include code)
```csharp
<Slider TValue="(double, double)" , Min="@_testMin" Max="@_testMax" DefaultValue="(_start, _end)" />
@code {
    private double _testMin = 0;
    private double _testMax = 60;
    private double _start = 10;
    private double _end = 30;
}
```
And gif:
![slider_right_broken](https://user-images.githubusercontent.com/6518006/101265311-0c673800-374e-11eb-8919-9d06286d67d9.gif)
I was also able to  recreate it using the [documentation for slider](https://antblazor.com/en-US/components/slider).

### Further technical details
- AntDesign commit 58d29af (Dec 5, 2020)
- Running net5.0
- Visual Studio 16.8.2

### My analysis
First, I thought that the problem is in ```Slider.razor.cs``` in the usage of ```bool _right = true``` field. It seemed to me that this field is used for evaluation of which edge is being changed. But later on I figured that this is just to start-off a calculation that is assuming the change applies to the same edge as the last change, and actualy during the calculation in ```CalculateValueAsync``` it is really evaluated there which edge needs to be updated (although favours current scenario ```_right```). It seems to me that the problem there is that it deals with numbers and comparisons that unfortunatelly are too absolute for user experience. So in my example:
- Min: 0, Max: 100
- Start : 10, End: 30  

1. If I change end by as small margin (let's say to 29) and ```_right = true``` then what is evaluated is this ``` 29 < 10 //new value < current start```. If ```true``` then assumes that start edge is being changed, otherwise, end edge is being changed.
2. If now I try to do a similar change to start (let's say to 9) with ```_right = true``` then again the same evaluation is done ``` 9 < 10 //new value < current start```. This time it is ```true``` so ```_right``` is set to ```false``` and left/start edge is update with the new value. From now on the left edge scenario will be favoured.
3. If now I try to change again the end (let's say to 28) and ```_right = false``` then new evaluation is run (according to the left edge scenario) ``` 28 > 29 //new value > current end```. This evaluates to ```false```, so slider assumes that actually left edge had changed (which is wrong, because end/right edge had been changed). 

[...]

Issue #855, opened 2020-12-04
Initial values for range in slider component are rendered as a single sided.


anddrzejb:
### Describe the bug
When I tried to use slider and supply ```Value="(_start, _end)"``` the slider initially was rendered with only single (_start) thumb.

### Steps to reproduce (please include code)
A basic setting like this will show what I mean:
```csharp
<Slider TValue="(double, double)" , Min="@_testMin" Max="@_testMax" Value="(_start, _end)" />
@code {
    private double _testMin = 0;
    private double _testMax = 60;
    private double _start = 10;
    private double _end = 30;
}
```
The result:
![image](https://user-images.githubusercontent.com/6518006/101219743-242ab780-368d-11eb-9192-1f0cc5006eab.png)

Expected result:
![image](https://user-images.githubusercontent.com/6518006/101219469-c9915b80-368c-11eb-9d72-1bcc46b7b62d.png)

### Further technical details
- AntDesign commit 2a3d171 (Dec 3, 2020)
- Running net5.0
- Visual Studio 16.8.2

### My analysis:
```[Parameter]``` attribute on ```Range``` was commented out and it is evaluated based on ```TValue``` type. The problem is that it is evaluated in ```OnInitialized()``` but it is already used right after the value is set (in the base class) and assumes default ```false```. A quick fix I did - I removed the evaluation of ```TValue``` type from ```OnInitialized()``` method and made a following fix to the property ```Range```:
```csharp
        /// <summary>
        /// dual thumb mode
        /// </summary>
        //[Parameter]
        private bool? _range;
        public bool Range
        {
            get
            {
                if (_range == null)
                {
                    Type type = typeof(TValue);
                    Type doubleType = typeof(double);
                    Type tupleDoubleType = typeof((double, double));
                    if (type == doubleType)
                    {
                        _range = false;
                    }
                    else if (type == tupleDoubleType)
                    {
                        _range = true;
                    }
                    else
                    {
                        throw new ArgumentOutOfRangeException($"Type argument of Slider should be one of {doubleType}, {tupleDoubleType}");
                    }
                }
                return _range.Value;
            }
        }
```
This fix solved seems to solve this problem.

PR #866, opened 2020-12-07
fix: Slider edge lock, range initial values, both range value change

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#859 
#864 
#855 
#865 
#868 
#867 

### 💡 Background and solution
#855 solution as described in the issue
#864 If detected both values are changed, then the simplest solution is to reset ```_leftValue``` and ```_rightValue``` to their default min/max and let the procedure do "from scratch" evaluation
#855 In connection to described solution in the issue, there was another issue - the edges were still moving on click when a click was initiated and current active window (at least in Windows OS) was not the browser. The changes include evaluating of ```JsonElement``` data in ```OnMouseDown``` and ```OnMouseUp``` event handlers.
#868 Reverse property change refreshes styles
#867 Value property is overridden to handle the validation. New method ```SortValue``` is created to sort values inside tuple
#865 Added ```Clamp``` function that is used to evaluate new edge values if they are within allowed parameters (between slider's min/max and for range before right edge or  after left edge). ```Clamp``` also gets nearest step. This is done regardless as new value has to be rounded to a step value before comparison with previous value can be done. The if previous value equals current value, no actions is performed.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/slider/Slider.razor.cs
+++ b/components/slider/Slider.razor.cs
@@ -634,6 +634,11 @@ namespace AntDesign
 
             if (Range)
             {
+                if (IsLeftAndRightChanged(value))
+                {
+                    _leftValue = double.MinValue;
+                    _rightValue = double.MaxValue;
+                }
                 LeftValue = DataConvertionExtensions.Convert<TValue, (double, double)>(value).Item1;
                 RightValue = DataConvertionExtensions.Convert<TValue, (double, double)>(value).Item2;
             }
@@ -641,6 +646,12 @@ namespace AntDesign
             {
                 RightValue = DataConvertionExtensions.Convert<TValue, double>(value);
             }
+        }
+
+        private bool IsLeftAndRightChanged(TValue value)
+        {
+            (double, double) v = DataConvertionExtensions.Convert<TValue, (double, double)>(value);
+            return (v.Item1 != LeftValue) && (v.Item2 != RightValue);
         }
     }
 }
--- a/components/slider/Slider.razor.cs
+++ b/components/slider/Slider.razor.cs
@@ -198,8 +198,21 @@ namespace AntDesign
         /// <summary>
         /// reverse the component
         /// </summary>
+        private bool _reverse;
+
         [Parameter]
-        public bool Reverse { get; set; }
+        public bool Reverse
+        {
+            get { return _reverse; }
+            set 
+            {
+                if (_reverse != value)
+                {
+                    _reverse = value;
+                    SetStyle();
+                }
+            }
+        }
 
         /// <summary>
         /// The granularity the slider can step through values. Must greater than 0, and be divided by (<see cref="Max"/> - <see cref="Min"/>) . When <see cref="Marks"/> no null, <see cref="Step"/> can be null.
@@ -674,7 +687,7 @@ namespace AntDesign
             return (typedValue.Item1 != LeftValue) && (typedValue.Item2 != RightValue);
         }
 
-        private TValue _value;
+        private TValue _value;        
 
         /// <summary>
         /// Gets or sets the value of the input. This should be used with two-way binding.
--- a/components/slider/Slider.razor.cs
+++ b/components/slider/Slider.razor.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text.Json;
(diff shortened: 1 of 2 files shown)


ElderJames:
Thank you for your contirbutions @anddrzejb !

It seems the reverse demo is not work, and the pointer is not in the center of the handle,.

image

anddrzejb:
> Thank you for your contributions @anddrzejb !
> 
> It seems the reverse demo is not work, and the pointer is not in the center of the handle.
> 
> image

Yeah, I noticed about 30 min ago, that my fix broke single edged slider. Also, the fix for not moving the edge when clicked was only applied to double-edged slider. I will have a fix for that soon. I will also address the performance issues and another maybe bug or maybe a design flaw, but you have to tell me if this is an intended behaviour or should this be fixed. 

I am referring to the situation, when a new set of values are provided, that are in reversed the order. So first I put ```(10, 30)``` and then ```(30, 10)```. For example this scenario:
```csharp
<Slider TValue="(double, double)" , Min="@_testMin" Max="@_testMax" DefaultValue="(_start, _end)" />
<Slider TValue="(double, double)" , Min="@_testMin" Max="@_testMax" DefaultValue="(_end, _start)" />
<Button OnClick="Reverse">Reverse </Button>
@code {
    private double _testMin = 0;
    private double _testMax = 60;
    private double _start = 10;
    private double _end = 30;

    private void Reverse()
    {
        var t = _start;
        _start = _end;
        _end = t;
    }
}
```
Notice that in the second slider first value provided is ```_end``` and then ```_start```. That causes slider to be rendered as a single point:
slider_reversed_order_err

I already have a fix for that. It goes in pair with #865 but if this is an intended design, then I can disable it. Although I imagine if the values can be served by an API, this scenario is not impossible.

components/slider/Slider.razor.cs
```diff
@@ -416,9 +426,39 @@ private async void OnMouseDown(MouseEventArgs args)
...
+                bool ctrlKey = jsonElement.GetProperty("ctrlKey").GetBoolean();
+                bool metaKey = jsonElement.GetProperty("metaKey").GetBoolean();
+                bool shiftKey = jsonElement.GetProperty("shiftKey").GetBoolean();
+
+                _edgeClicked = _edgeClickArgs.ClientX == clientX
+                            && _edgeClickArgs.ClientY == clientY
+                            && _edgeClickArgs.CtrlKey == ctrlKey
+                            && _edgeClickArgs.MetaKey == metaKey
+                            && _edgeClickArgs.AltKey == altKey
+                            && _edgeClickArgs.ShiftKey == shiftKey;
```
ElderJames, line 453: The `_edgeClickArgs` would be null.

```suggestion
                _edgeClicked = _edgeClickArgs != null
                            && _edgeClickArgs.ClientX == clientX
                            && _edgeClickArgs.ClientY == clientY
                            && _edgeClickArgs.CtrlKey == ctrlKey
                            && _edgeClickArgs.MetaKey == metaKey
                            && _edgeClickArgs.AltKey == altKey
                            && _edgeClickArgs.ShiftKey == shiftKey;
```

anddrzejb: Yes, exactly. Among other things, I have this fixed in coming commit.

anddrzejb: Just tried to commit but I am getting this error
```
husky > pre-commit (node v14.15.1)
/d/dotnet/ant-design-blazor_forked/node_modules/.bin/lint-staged: line 5: cygpath: command not found
internal/modules/cjs/loader.js:883
  throw err;
  ^

Error: Cannot find module 'D:\lint-staged\bin\lint-staged.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Function.executeUserEntryPoint  as runMain 
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
husky > pre-commit hook failed (add --no-verify to bypass) 
```
Of course I can run suggested --no-verify but should I do it? What is this? I never used husky. I tries rebuilding, restarting, running ```npm install```...

ElderJames: It is used to check the if the commit message is following the commit message standard.
You can skip it because I would change the commit message before merge PRs.

ElderJames (approved):
LGTM.

ElderJames:
Thank you @anddrzejb .

Issue #875, opened 2020-12-09
Alert does not render description from ChildContent


szmalec:
### Describe the bug
An Alert component with a description in ChildContent does not display this description.

### Steps to reproduce (please include code)
```
<Alert ShowIcon Type="@AlertType.Info" Message="Alert msg">
    Child content text used as alert description
</Alert>
```
but this works OK:
```
<Alert ShowIcon Type="@AlertType.Info" Message="Alert msg" Class="ant-alert-with-description">
    Child content text used as alert description
</Alert>
```

### Further technical details
I am using: 0.5.0-nightly-2012081156

ElderJames:
Thank you @szmalec

PR #881, opened 2020-12-11
Make description element visible if passed in ChildContent

szmalec:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#875 

### 💡 Background and solution

Fix is simple - modified conditions which are used to apply `ant-alert-description` to span with desciption (span was rendered but with default `display: none` style).
One thing to mention - the user can simultaneously pass the value in the `Description` parameter and define the content in `ChildContent`. In this case, two span tags with a description are inserted and render side by side instead of one below the other. I did not try to fix this because it is a rather rare case and I also feel average about CSS;]

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/alert/Alert.razor.cs
+++ b/components/alert/Alert.razor.cs
@@ -113,7 +113,7 @@ namespace AntDesign
                 .If($"{prefixName}-no-icon", () => !ShowIcon)
                 .If($"{prefixName}-closable", () => Closable)
                 .If($"{prefixName}-banner", () => Banner)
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
LGTM

PR #899, opened 2020-12-17
Implement optional key debouncing

### LINKED ISSUE #579: Performance issue - Lag when typing text quickly in text input
It seems a bug in the handling of the key up event of the text input control, specially when it is used in forms.

If I type different characters quickly in the text input,, the word appears after a few seconds and there is a warning in the console:

L: GC_MINOR: (Nursery full) time 2.05ms, stw 2.06ms promoted 169K major size: 6848K in use: 6102K los size: 1024K in use: 354K

There is no lag if I hold a key down, so I think the issue it is with the key up event.

You can reproduce the issues easily in the Form Size demo, just try to type a word quickly in the first input.

image

Hona:
Closes #579 

Code based on source from @Magehernan

If the parameter is not set, then there is no behaviour change, once the debounce milliseconds is set, then this will activate the changes.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/input/Input.cs
+++ b/components/input/Input.cs
@@ -222,7 +222,7 @@ namespace AntDesign
                 _compositionInputting = false;
             }
 
-            await ChangeValue();
+            await ChangeValue(true);
 
             if (OnBlur.HasDelegate)
             {
--- a/components/input/Input.cs
+++ b/components/input/Input.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Text.Json;
+using System.Threading;
 using System.Threading.Tasks;
 using AntDesign.JsInterop;
 using Microsoft.AspNetCore.Components;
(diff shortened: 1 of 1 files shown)


ElderJames:
Thank you for contribution @Hona 

I think we could also add another flag to change value on blur.

Hona:
@ElderJames Do you mean forcing the update without the debounce? If so then it should be done in the last commit

ElderJames:
Yes, that can preserves the behavior of the build-in InputText component. It listen the blur event instead input or keyup.

ElderJames:
Hi @Hona ,I have test the last commit, and it also would stuck at change value sometime. So I think we can only change value with blur event when ` DebounceMilliseconds = 0`. It's the same behavior of the build-in InputText.

ElderJames (approved):
Thank you!

PR #903, opened 2020-12-19
fix(module:slider): drag, select, focus and rapid reposition fix

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
This addresses a number of small issues with slider:
1. When clicked and dragged to the left, often before first movement to the left a movement to the right was executed: 
slider_goes_back_on_click

2. When dragging, especially after page load, nothing was happening: 
slider_drag_lag

3. For ranged slider, when dragging one edge beyond the other range, the focus was not following (I am referring to transparent gradient border that is showing when edge gets focus) :
slider_focus_fix

4. For ranged slider, when quickly dragging one edge beyond the other edge, the "not changed" edge also moves:
slider_non_active_edge_not_locked

5. When dragging an edge, text selection should be inactive:
slider_drag_and_select

### 💡 Background and solution
1. Required changes to calculation method of the edge position.
2. Previously added method ```IsMoveInEdgeBoundary``` was removed from ```OnMouseMove``` event handler.
3. When side switch is discovered, a js focus method is called on appropriate edge.
4. When ranged slider is used, original values of each edge are stored when mouse down is registered. During side switch, the stored values are persisted, to ensure they do not change.
5. Added to class ```@{slider-prefix-cls}``` css ```user-select:none```

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/slider/Slider.razor
+++ b/components/slider/Slider.razor
@@ -2,7 +2,7 @@
 @typeparam TValue
 @inherits AntInputComponentBase<TValue>
 
-<div class="@ClassMapper.Class" style="@Style" id="@Id" @ref="_slider"
+<div class="@ClassMapper.Class" style="user-select:none;@Style" id="@Id" @ref="_slider"
      @onmousedown="(e)=>OnMouseDown(e)">
     <div class="ant-slider-rail">
     </div>
--- a/components/slider/Slider.razor.cs
+++ b/components/slider/Slider.razor.cs
@@ -28,6 +28,8 @@ namespace AntDesign
         private bool _mouseMove;
         private bool _right = true;
         private bool _initialized = false;
+        private double _initialLeftValue;
+        private double _initialRightValue;
 
         private string RightHandleStyleFormat
         {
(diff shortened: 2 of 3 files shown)


ElderJames:
Thank you @anddrzejb for the great contribution.
Everything is good but the style changing is not a good idea, because the styles are synchronized from the ant-design.
So it's best to go there repo and fix the style.

anddrzejb:
So I see 1 option then: hard code it in the slider div in the ```style``` attribute. Do you agree @ElderJames ?

ElderJames:
> So I see 1 option then: hard code it in the slider div in the `style` attribute. Do you agree @ElderJames ?

@anddrzejb It's OK.

ElderJames (approved):

Issue #144, opened 2020-05-23
Missing tooltip in slider component


1002527441:
the following function onAntSlider.
1. TooltipVisible, 
2. TooltipPlacement 
3. TipFormatter 

above function is not implmented.

Brian-Ding:
tooltip is pending on the Tooltip component

ElderJames:
@Brian-Ding tooltip is done. #131

Issue #383, opened 2020-07-19
Unable to style DatePicker due to OverlayTrigger div


Epictek:
Using the following blazor code:

```HTML
<RangePicker TValue="DateTime?" ClassName="col-span-7"/>
```

Results in the following HTML:

```HTML
<div class="" style="display: inline-flex; " id="ant-blazor-4e752cb7-5b21-44ae-87aa-4fb22f9d3aa3" _bl_2dbe0536-9d98-42bf-a788-09db05be7761="">
<div class="ant-picker ant-picker-default col-span-7 ant-picker-range" id="ant-blazor-bd2ecd23-a441-4cc7-89bb-efbd2e5d84b5" _bl_5155abb8-1947-4c6c-ad49-07ff1a7933a7="">
---
</div
</div>
```

it seems the `display: inline-flex` div is being created by OverlayTrigger. Is it possible to remove the `div` from OverlayTrigger or less ideally pass the class parameter through to it?

anddrzejb:
It seems this is what I am trying to discuss in #898. Glad to see I am not the only one that sees this as a potential issue.

PR #937, opened 2020-12-29
feat(module:overlay): OverlayTrigger not bound to a div

### LINKED ISSUE #898: (Discussion needed) Should the wrapping div be removed from components using Overlay (like Tooltip)?
I was working on improving slider experience. I tried adding a tooltip to the edge(s) and I found out, that when wrapping an edge in a tooltip, the edge was being wrapped by a ```div ``` element. That was causing visual displacement of the edge (was falling our of the slider line) like in this image: 
image

The reason was because the wrapping ```div``` has a hard coded style ```display: inline-block``` in tooltip. Removing it fixes the issue, but also breaks other demos (for example tooltip demo). Or another solution is to add to the tooltip's style attributes something like this: ```Style="display: inline;"```, which in turn renders to this weird sequence:
image

but it fixes the displacement issue.
That sent me on a research trip - how does react ant-design addresses this problem? I found out, that it does not wrap into additional ```div```.

So I started wandering how can this be achieved in blazor. That was a challenge, because the parent (tooltip) does not keep any reference of the child (for example slider egde) nor it has any simple access to it. After a lot of research and experiments, I found out the project already has ForwardRef inside (I guess from [this blog entry](https://www.samprof.com/2019/06/03/blazor-forwardref)) - a kind of solution on how to get a reference to a child from a parent. With that, I was able to make appropriate changes to the tooltip and I manage to get rid of the ```div``` and keep the functionality of the tooltip. There are still some other challenges there (like attaching context event), but I do not want to purse a solution that will never be used.
My problem is that this kind of change requires also changes to ```OverlayTrigger``` component which in turn effects a number of other components. Also, that kind of change would be a breaking change - anyone using the components that rely on ```OverlayTrigger``` would have to alter their code to include ```ForwardRef``` in their calls. For example, my new trigger code looks like that (notice extra ```@ref``` in the ```span``` element):
```razor
<Tooltip Title="@("prompt text")" Visible="true">
    <span @ref="@context.Current">Tooltip will show on mouse enter.</span>
</Tooltip>
```

My question is this - should we be interested in getting rid of the extra ```div```? If yes, does anyone have a better suggestion on how to get reference to the child in the parent? Can such reference be extracted from ```RenderFragment```?

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Eventually if implemented, fixes #898 and #383

### 💡 Background and solution
For the ```div```-free solution, the parent element (for example ```Tooltip```) needs to have a reference to its child. Here it is solved using ```ForwardRef``` and template approach. This solution if applied to currently existing ```OverlayTrigger```, would unfortunately represent a breaking change. So for purpose of discussion and additional work, I created new versions of ```OverlayTrigger```, ```Overlay``` and ```Tooltip``` with suffix ```Unbound```:
- ```OverlayTriggerUnbound```
- ```OverlayUnbound```
- ```TooltipUnbound```  

```Unbound``` keyword should indicate that this is an overlay that is not bound to a div (or actually to anything). In the tooltip ```Basic.razor``` demo file I also created a tooltip that uses the new approach.  

In below gif notice that 1st tooltip has a wrapper (uses the regular```Tooltip```) and 2nd tooltip does not have any wrapper but directly shows the tooltip's child element ```<span>``` (uses newly created ```TooltipUnbound```).
overlayunbound_sample

Also - this example is not yet fully done for trigger. Because the events can no longer be attached to ```div```, they are attached to child element using the reference (```ForwardRef```). This PR in current state does not handle ```@oncontextmenu``` event. But it should handle all the other events that are defined on the regular ```Tooltip```. I guess the ```@oncontextmenu``` event can be implemented in the same manner as the other events.
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### CODE REVIEW

**Comment by ElderJames** [comment]
Thank you very much @anddrzejb , it help a lot.

But if you implement a new component with a lot of repetitive code, it would increase maintenance costs and confuse users.
Could you implement this feature by inheriting from the original component?

We might be more comfortable with using a new template parameter:

```
<Tooltip Title="@("prompt text")">
    <Unbound>
        <span  @ref="@context.Current">Tooltip (div-unbound) will show on mouse enter.</span>
    <Unbound>
</Tooltip>
```

However, I have to say that this is a great feature and I really hope to merge it into the project.

**Comment by anddrzejb** [comment]
I added this as a draft PR exactly for that purpose. I was hoping to get input on the idea and use others experience. Also, I did not want to spent a ton of time just to find out the solution is not acceptable. That is why I just did a code repeat - to get the discussion started as well as have a working example. 

To address your point - at first was trying to inherit from original component, but for the solution to work, ```ChildComponent``` has to be overridden with ```ChildComponent<ForwardRef>```. And this cannot be done (different types). And blazor does not allow to have 2 properties with name ```ChildComponent```. That is why I was forced to use current solution. Of course, I could extract as much as possible of the common code to avoid code duplication (to base classes I guess). Here it is just a proof-of-concept. I will give another shot to your proposal but I am also open to suggestion on how to solve the issues I foresee...

**Comment by anddrzejb** [comment]
@ElderJames following your suggestion, I did some research on multiple child elements and managed to revert to single component use. This seems to work. However there is still one (minor?) problem - I did not manage to replicate ``` @oncontextmenu:preventDefault``` for unbounded ```Tooltip```. It seems to me that a change to ```interop.ts``` and ```DomEventService``` would be required (to add additional optional argument ```bool preventDefault = false```).

**Comment by ElderJames** [comment]
Yes, a little bit of necessary JS is acceptable.

**Comment by anddrzejb** [comment]
@ElderJames now full functionality of ```Tooltip``` has been replicated in the ```<Unbound>``` version. Please have a look at the JS changes as this is not my strong suite (and connected changes in ```DomEventService```). 
I will stop working on this until you tell me to either fix something or that everything is fine. 
If you accept this change, I will proceed to make similar changes to all other components that use ```OverlayTrigger```, so all of them can benefit from the unbounded version. Once that work is done, I will change this PR to ready for review.

File: components/tooltip/Tooltip.razor.cs [hunk 1]
```diff
@@ -55,6 +61,27 @@ internal override async Task Hide(bool force = false)
...
+        protected override Task OnAfterRenderAsync(bool firstRender)
+        {
+            if (firstRender && Unbound != null)
+            {
+                Ref = RefBack.Current;
+                DomEventService.AddEventListener(Ref, "mouseover", (async e => await OnTriggerMouseEnter()), true);
+                DomEventService.AddEventListener(Ref, "mouseout", (async e => await OnTriggerMouseLeave()), true);
+                DomEventService.AddEventListener(Ref, "focusin", (async e => await OnTriggerFocusIn()), true);
+                DomEventService.AddEventListener(Ref, "focusout", (async e => await OnTriggerFocusOut()), true);
+                DomEventService.AddEventListener(Ref, "contextmenu", OnContextMenu, true, true);
```
ElderJames, line 73: You may have to remove event listeners in dispose method.

site/AntBlazor.Docs/Demos/Components/Tooltip/demo/Basic.razor
```diff
@@ -1,3 +1,9 @@
 ﻿<Tooltip Title="@("prompt text")">
-    <span>Tooltip will show on mouse enter.</span>
+    <span>Tooltip (bound to div) will show on mouse enter.</span>
+</Tooltip>
+<br />
+<Tooltip Title="@("prompt text")">
+    <Unbound>
+        <span @ref="@context.Current">Tooltip (unbound) will show on mouse enter.</span>
+    </Unbound>
 </Tooltip>
```
ElderJames, line 9: I feel that `Unbound` is not easy for users to understand. Is there a more direct and precise word for it?

anddrzejb: Generally I do no like it either, but we have to somehow communicate this is different approach. The ```Unbound``` is widely used in VB and this is where I took my inspiration from, but it was just a quick hack. Since the idea is to keep both solutions, somehow we need to communicate to the consumers which solution is being currently used. We are already here at a disadvantage, because we won't communicate (I guess at least) that the consumer is currently using div version of a component (so without reading docs, this won't be clear - I image there will be a lot of questions about it in the future). Best way would be maybe for the consumers to express what would be most intuitive? Otherwise it's just us giving proposals that may be clear to us but not necessarily to others. 
However, I can also give maybe a few proposals:
- ```<divless>```
- ```<wrapperless>```
- ```<wrapless>```
- ```<boxless>```
- ```<unboxed>```  

These are couple of suggestions. I was trying to suggest something that may directly indicate no div. Mind that English is not my primary language, so there's that...

components/tooltip/Tooltip.razor.cs
anddrzejb: So something like this I figure (I moved action definitions to separate methods)
```csharp
protected override Task OnAfterRenderAsync(bool firstRender)
{
    if (firstRender && Unbound != null)
    {
        Ref = RefBack.Current;
        DomEventService.AddEventListener(Ref, "mouseover", OnTooltipMouseEnter, true);
        DomEventService.AddEventListener(Ref, "mouseout", OnTooltipMouseLeave, true);
        DomEventService.AddEventListener(Ref, "focusin", OnTooltipFocusIn, true);
        DomEventService.AddEventListener(Ref, "focusout", OnTooltipFocusOut, true);
        DomEventService.AddEventListener(Ref, "contextmenu", OnContextMenu, true, true);

    }
    return base.OnAfterRenderAsync(firstRender);
}

private void OnTooltipMouseEnter(JsonElement jsonElement) => OnTriggerMouseEnter();
private void OnTooltipMouseLeave(JsonElement jsonElement) => OnTriggerMouseLeave();
private void OnTooltipFocusIn(JsonElement jsonElement) => OnTriggerFocusIn();
private void OnTooltipFocusOut(JsonElement jsonElement) => OnTriggerFocusOut();

protected override void Dispose(bool disposing)
{
    DomEventService.RemoveEventListerner<JsonElement>(Ref, "mouseover", OnTooltipMouseEnter);
    DomEventService.RemoveEventListerner<JsonElement>(Ref, "mouseout", OnTooltipMouseLeave);
    DomEventService.RemoveEventListerner<JsonElement>(Ref, "focusin", OnTooltipFocusIn);
    DomEventService.RemoveEventListerner<JsonElement>(Ref, "focusout", OnTooltipFocusOut);
    DomEventService.RemoveEventListerner<JsonElement>(Ref, "contextmenu", OnContextMenu);
    base.Dispose(disposing);
}
```

ElderJames: Exactly :)

anddrzejb: Done. Commit 2c00885

site/AntBlazor.Docs/Demos/Components/Tooltip/demo/Basic.razor
ElderJames: Users would wonder why there is a bound 😂

anddrzejb: Well...we could always change ```ChildContent``` to ```Bound``` or something similar just to indicate that there are 2 options, but that idea will create a breaking change and probably will make users scratch their heads on why there is even such option like div-bounded.

So. I see 3 options:
1. Keep 2 solutions - 1 with ```ChildContent``` (as is) and 1 with whatever we are going to name it (the currently used ```Unbound```)
2. Introduce a breaking change - and go with (for the sake of argument) ```Bound``` and ```Unbound``` versions
3. Introduce a breaking change - get rid of div-bounded solution and make current ```Unbound``` a default (thus removing the necessity to use ```<Unbound>``` tag at all) but force the users to use ```@context``` in their child element (that ```@context``` is the breaking change). 

The question you probably should ask yourself is - eventually, which solution should be preferred? As a full disclosure - I am obviously biased and I think it is clear I will try to promote the solution (or at least the idea) I presented here. However I feel that eventually the ```Unbound``` version should become the default, which in turn will introduce a breaking change. Since ant-design-blazor is still at its 0.x version, it is expected for it to keep evolving in violate ways which would create breaking changes. If that is going to be the case, then maybe it would be good to keep the ```Unbound``` version as is without spending too much time on figuring out a name that anyway eventually will be retired. We could keep the ```Unbound``` as is for now, see how it is going to work, fix any issues that are going to arise. And once you decide to make it default, then introduce a breaking change.

ElderJames: Thank you, I'm prefer to the first optio :)

anddrzejb: So the ```<Unbound>``` tag stays? Are there any other suggestions? Shall I start working on other components that rely on ```OverlayTrigger```?

ElderJames: Keep `<Unbound>`, and introduce the difference between with the `ChildContent`. 
Yeah, you can please continue to work on the other components.

anddrzejb: Please check commit 8eb8a04. I have no clue on how to approach documentation changes especially when it deviates in functionality compared to react ant-design. 
This is only in English...

ElderJames: It's fine, and I will translate it to Chinese doc.

anddrzejb:
@ElderJames I am now working on `Popover` and I need advise. In the docs, `Popover` uses `Button` component as its child element. This is a component and not an element and it does not exposes its `ref`, so I cannot do
```razor
<Popover ContentTemplate="@_content" Title="Title">
    <Unbound>
        <Button type="primary" @ref="@context.Current">Hover me (unbound)</Button>
    </Unbound>
</Popover>
```
because I get a compilation error `Cannot implicitly convert type 'AntDesign.Button' to 'Microsoft.AspNetCore.Components.ElementReference'`. I cannot find any parameter in any of the base classes that would actually expose element reference to a dom element. Am I missing something?

ElderJames:
In the end of the [post of ForwardRef](https://www.samprof.com/2019/06/03/blazor-forwardref), gave a solution for components. It may be:

```
<Popover ContentTemplate="@_content" Title="Title">
    <Unbound>
        <Button type="primary" RefBack="@context" >Hover me (unbound)</Button>
    </Unbound>
</Popover>
```

anddrzejb:
That unfortunately does not work. I actually tried that when I was going through what is available in the base classes (sorry for not communicating that clearly). It may have something to do with the fact that `RefBack` is not really attached to the top level dom element of the `ChildContent`. I will be trying to figure it out...

ElderJames:
Maybe we need to see how does MatBlazor do that.

anddrzejb:
I will try to work it out. For now I've done `Dropdown` and `Popconfirm`, but I imagine `Popconfirm` may have the same issue if used with a component, so this issue has to be solved first.
Also I was thinking - some components use `Dropdown` for their rendering (for example `DatePicker`). Should they be changed, so the (internally) will start using the `<Unbound>` child content?

ElderJames:
Yes, we need to work on using `<UnBound>` for internal components after this PR is merged.

anddrzejb:
@ElderJames So I found out why `<Button>` was not working with `RefBack`. Have a look at the commit a77b788 - basically it seems that the `<button>` html tag had no `@ref=Ref` set. That means that basically all components that can be used with `Overlay` will have to have this fixed (whenever missing) for their primary html tag. Is there any list of components that can be used with `Overlay` or are we looking at all components. Anyway it seems logical that `Ref` from `AntDomComponentBase` was supposed to be set to the significant/key element and probably should be set whenever it is missing (for example `<InputGroup>` should probably have it at the `<span>` element while `<textarea>` should have it for both `<span>` and `<textarea>` version, etc). Do you agree?

ElderJames:
@anddrzejb I'm sorry for that.
We may forget it because we never used it.The feature you're working on might be required so all components need to be checked.

anddrzejb:
@ElderJames I was working on `SubMenu` but it seems to me it is pointless to switch it to `<Unbound>` version just because it uses `OverlayTrigger`. This particular component does not use `div` element but `li` element. Regardless of the approach, that element is still necessary, so I think we can leave it as it is. What do you say?

ElderJames:
@anddrzejb  It's fine.

anddrzejb:
@ElderJames Have a look at commit fcf2ab4. I adapted `Select` to used `<Unbound>`. However I had to make couple of changes there, since the `Select` neither the dependent `SelectContent` do not inherit from any of the AntComponentBase classes:
1. I added to `SelectContent`:
    -  `ForwardRef` and 'ElementReference' to  (normally I was getting them from `AntComponentBase` and `AntDomComponentBase`). That actually got me wondering - should those also inherit from one of the 2?
    - `EventCallback<MouseEventArgs> OnArrowClick` that is going to handle clicking on the `Select` icon. I noticed this was not properly handled before (for example react ant-design does not expand the select options if icon is set to `Loading`, but ant-design-blazor does). And also switching to `<Unbound>` stopped expanding the select options at all when arrow icon was clicked.
2. As a consequence of arrow-click issue, I had to expose `OnDivClick` event in `OverlayTrigger` as public method.
3. In `Select.razor` I use the exposed `OnDivClick` method to pass it to the event callback.

anddrzejb:
Good idea & done. This is what you had in mind? (commit cfc2c63)

ElderJames:
It's nice. But I personally feel the method name `Show` is a little confusing.

anddrzejb:
No problem. I was kind-of following the logic of `OverlayTrigger.Show()` but I guess since it calls anyway an `OnDivClick`, it is more appropriate.

ElderJames:
@adsioj Would you fix #144 ? And there is a popover in avatar component, and a tooltip in table column component.

anddrzejb:
@ElderJames This whole ordeal with `OverlayTrigger` not bound to a div is because I was working on `Slider` (including the works needed to solve #144). I did not want to create a separate `Tooltip` and wanted to reuse existing. Actually I think I have already done most work in `Slider` & `Tooltip` area, but it is awaiting for this PR to be merged. 

Thanks for reminding me about other components that might need changes in connection to this PR. I will do them probably today.

On another note - I have some real issues with using `<Unbound>` in `AutoComplete`. I will probably post here a non-working commit for you to have a look - I must be blind to something, because the errors I get make no sens to me. I will let you know in details which commit it will be and what my issues are...

ElderJames:
Sure, you can create another issue to point out the details.

anddrzejb:
@ElderJames I did `Avatar` but after looking at table's `Column` component and its usage of the `Tooltip` - it seems to me the change there is not really needed. `Tooltip` adds its `div` element (with class parameter `"ant-table-column-sorters-with-tooltip"`) that apparently is also a part of react ant-design. We would have to anyway add that `div` element manually...

anddrzejb:
@ElderJames So I have just commited f28186e partial changes for `Autocomplete`. It works without `<div>` for childless `Autocomplete`. However, I cannot figure out how to use the `<Unbound>` with autocomplete's `ChildContent`. I tried making it generic (`RenderFragment<ForwardRef> ChildContent`) so I could pass `<Unbound>` context to it, but I get compiler error ` The type or namespace name 'ForwardRef' does not exist in the namespace '__Blazor.AntDesign'`. What I wanted to achieve was to transfer the ref from `ChildContent` of `Autocomplete` to OverlayTrigger's `Unbound`. 
I also tried with new `[Parameter] RenderFragment<ForwardRef> UnboundContent` (thus leaving the `ChildContent as it was), but if I tried to use it:
```razor
<AutoComplete Options="options" OnInput="OnInput" AllowFilter="false">
    <UnboundContent>
        <AutoCompleteSearch Placeholder="input here" @bind-Value="@value" RefBack="context"/>
    </UnboundContent>
```
I again got the same compiler error. 

For some reason this does not seem to work. Either I do not know something or am blind to something. Any suggestions?

ElderJames:
How about using CascadingParameter?

anddrzejb:
I think I need more direction...I was trying to think how can I use CascadingValue but cannot figure it out what you had in mind. You were suggesting to pass down OverlayTrigger's context as a CascadingValue? Can you give me a code sample from the top of your head?

components/slider/Slider.razor.cs
```diff
@@ -320,17 +340,56 @@ private double RightValue
         [Parameter]
         public Action<TValue> OnChange { get; set; }
 
+        [Parameter]
+        public bool HasTooltip { get; set; } = true;
```
anddrzejb, line 344: I decided to move from the logic of having `TipFormatter` as null to indicate that the tooltip should not be shown at all. My problem here was that I could not pass null in the markup
```razor
<Slider ... TipFormatter="null"/>
```
(and other modifications) were showing compilation errors.

components/slider/Slider.razor.cs
```diff
@@ -320,17 +340,56 @@ private double RightValue
...
+
+        /// <summary>
+        /// Slider will pass its value to tipFormatter, and display its value in Tooltip
+        /// </summary>        
+        private bool _isTipFormatterDefault = true;
+        private Func<double, string> _tipFormatter = (d) => d.ToString(LocaleProvider.CurrentLocale.CurrentCulture);
```
anddrzejb, line 350: Previous `Func<string>` was replaced so `Default` values could also be shown using `TipFormatter`. Now each edge will provide its current value to and expect a string back.

components/slider/Slider.razor.cs
```diff
@@ -508,21 +591,33 @@ private async void OnMouseMove(JsonElement jsonElement)
...
+                {
+                    _tooltipRightVisible = TooltipVisible;
+                    _toolTipRight.Visible = TooltipVisible;
+                }
+
+                if (_tooltipLeftVisible != TooltipVisible)
+                {
+                    _tooltipLeftVisible = TooltipVisible;
+                    _toolTipLeft.Visible = TooltipVisible;
+                }
```
anddrzejb, line 616: This part looks probably weird and confusing, but this was the only way I could find to deal with keeping the `Tooltip` visible while dragging slider's edge. Additional concern here was that clicking on an edge would keep the `Tooltip` visible. 
So `_tooltipRightVisible` and `_tooltipLeftVisible` are passed to `Overlay` as parameters (to `Visible` property). Any change to one of those parameters either forces the `Overlay` to show or to hide itself (in `OnParametersSetAsync()`). Here though, I was trying to deactivate force showing (during dragging) but I did not want the `Tooltip` to vanish. That is why there is also a call to `Overlay.Visible` that is omitting the `OnParametersSetAsync` and forced hiding.
I also tried using only call to `Overlay.Visible` but that get's reset on each refresh, so `Tooltip` was vanishing during dragging.

ElderJames: It's fine.

ElderJames: > The only "beef" is that I get a compliant from VS that Component parameter 'Visible' should not be set outside of its component.

You can replace it with a public method, `SetVisible(visible: bool)` etc.

anddrzejb: Will do it.

anddrzejb:
@ElderJames Commit 5acca04 is a substantial change. I went through all components and applied the following changes (whenever applicable):
- similar changes as done to the `<Button>` component (missing `@ref`, refer to [this comment](https://github.com/ant-design-blazor/ant-design-blazor/pull/937#issuecomment-757528984) and your following comment about necessity to make the changes)
- applied to `CascadingValue` flag `IsFixed="@true"` whenever `Value="this"` (according to #992)

Less "bundled" changes:
1. I made changes to components in `typography` folder. I applied Steve Sanderson's suggestion to hard-code sequence numbers (according to #992).  
2. There was also a change in tab-tests - the changed test was asserting for "Content 2" but markup was giving back "/n/n  Content 2/n". So it I added `.Trim()`.  
3. Component `LabelTemplateItem` (part of `Select`) and `SpaceItem` (part of `Space`) do not inherit from `AntDomComponentBase` so I added missing `ElementReference Ref` and `ForwardRef RefBack` manually. I experimented with making those classes inherit from `AntDomComponentBase` and all seemed ok, but eventually I decided to roll back to `ComponentBase` and add missing properties. Let me know if you expect them to inherit from `AntDomComponentBase`.
4. I discovered that `Rate` and `Upload` components use `Tooltip`. I changed the `Rate` component to use `<Unbound>` but I noticed there is a quite substantial change coming with PR #1001 for `Upload` component, so I will do the change once this PR is merged.

anddrzejb:
@ElderJames Can you address the [comment](https://github.com/ant-design-blazor/ant-design-blazor/pull/937#issuecomment-758489083) regarding `Column`'s tooltip? Would you like to proceed anyway with making it `<Unbound>`?

ElderJames:
Yes, Column's tooltip should switch to unbound usage and make it align center.

anddrzejb:
@ElderJames I think that my last commit 59aebd2 concludes the works within the scope of this PR (I've got to admit it was more work than I planned, but quite fruitful). With this commit, I finally solved the issue I had with `AutoComplete`. If you have anything else that you think should be done within this PR, please let me know. Otherwise I will soon change this draft into ready for review.

ElderJames:
I think there are enough modifications to this pr that we can deal with it first.

components/upload/Upload.razor
```diff
@@ -57,89 +57,90 @@
...
-                        <div class="ant-upload-list-item ant-upload-list-item-uploading ant-upload-list-item-list-type-picture-card">
-                            <div class="ant-upload-list-item-info">
-                                <div class="ant-upload-list-item-thumbnail">@Locale.Uploading</div>
-                                <span class="ant-upload-list-item-name ant-upload-list-item-name-icon-count-1" title="@file.FileName">
-                                    @file.FileName
-                                </span>
-                            </div>
-                            <div class="ant-upload-list-item-progress">
-                                <Progress Percent="file.Percent" StrokeWidth="2" Type="ProgressType.Line" ShowInfo="false"></Progress>
+                    <Unbound>
```
ElderJames, line 60: I find the `Class` in Unbound usage is not work, and here is missing the `containerClass` so the style is broken.

ElderJames, line 60: I find the `Class` in Unbound usage is not work, and here is missing the `containerClass` so the style is broken.

anddrzejb: That is absolutely correct. This is the situation when `Tooltip`'s div was actually used for something. Same situation as with `Table`'s `Column`. I fixed it in the commit 9256ef1.

ElderJames (approved):

ElderJames:
Finally we merged this PR, thanks a lot for this time!

anddrzejb:
Thanks as well! I did not expect my attempt on making slider with tooltip will be so time consuming, but it was rewarding nevertheless.

ElderJames:
Yes, you've solved a big problem!

PR #987, opened 2021-01-15
Added Static class IconType for intellisense

porkopek:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [X] Other A convenience static class for icons, with the same sugar as an Enum

### 🔗 Related issue link

The `Icon Type` expects a string, not an enum, but there is no way to use this component without insert a magic string as the Type. With this static class, you don't need to remember the name of the icons, nor you can write the wrong name, besides it provides intellisense. 

### 💡 Background and solution

I made a static class to provide intellisense and type safety for icons
image

### 📝 Changelog

No breaking changes, just more convenience

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |       X    |
| 🇨🇳 Chinese |           |


ElderJames:
Thank you for contribution @porkopek .

It helps a lot. Did you make it by a generator script? Could you offer it that we can use for Icon synchronization?

porkopek:
Sure. I did it with a little script just to generate the code, not the file. But if you point me where in the code you put the auto-generated files, I could write a class that generate the file automatically, so no need to synchronize it

ElderJames:
You can put it in the ./scripts directory. I want to generate the icon components from svg files. Like the icon components in react version.

porkopek:
I've created a C# static class that generates the IconType, or can be used to generate all kind of static class that work as enums, as IconType does.
It's a C# class, because it uses `IconStore.GetAllIconNames()` to fill the icons type, so it didn't feel right to put it under `/scripts` and I put it under `/icon/internal`, but I don't know how are you going to call it to create the file.
The class has a method `GenerateIconTypeFile(string path)` that creates this file.
You can use `GetClassString(string className, IEnumerable<string> properties)` to create the string of the class (that can be then generated on disk) passing a collection of properties. So, for example color properties, you could generate the file calling
```
            var colors = new List<string> { "primary", "secondary", "tertiary" };
            var classString =StaticClassGenerator.GetClassString("ColorType", colors);
            StaticClassGenerator.GenerateFile(classString, "some-path", "ColorType.cs");
```

Let me know if this is useful for the project or needs any changes

ElderJames:
Thank you @porkopek .

I'm prefer to generate the icon class form https://github.com/ant-design/ant-design-icons/tree/master/packages/icons-svg/svg

porkopek:
This is just to have intellisense in `<Icon Type="**you have intellisense here**" />`, because even if you create individual icon components for each type, you are going to keep maintaining this <Icon Type /> component

Issue #873, opened 2020-12-09
dropdown's overlay rendered with wrong top and left value but the dropdown button works


zmayor:
### Describe the bug
the dropdown's overlay rendered at bottom-left of the page

### Steps to reproduce (please include code)

```
@page "/test"

<div>
    @*do not works correctly*@
    <Dropdown Trigger="new []{TriggerType.Click }">
        <Overlay>
            <Menu>
                <AntDesign.MenuItem>
                    1st menu item
                </AntDesign.MenuItem>
                <AntDesign.MenuItem>
                    2nd menu item
                </AntDesign.MenuItem>
            </Menu>
        </Overlay>
        <ChildContent>
            Dropdown
        </ChildContent>
    </Dropdown>
    @*works correctly*@
    <DropdownButton Trigger="new []{TriggerType.Click}" OnClick='e => { Console.WriteLine("Click on left button."); }'>
        <Overlay>
            <Menu>
                <AntDesign.MenuItem>
                    <a target="_blank" rel="noopener noreferrer" href="http://www.alipay.com/">
                        1st menu item
                    </a>
                </AntDesign.MenuItem>
                <AntDesign.MenuItem>
                    <a target="_blank" rel="noopener noreferrer" href="http://www.taobao.com/">
                        2nd menu item
                    </a>
                </AntDesign.MenuItem>
                <AntDesign.MenuItem>
                    <a target="_blank" rel="noopener noreferrer" href="http://www.tmall.com/">
                        3rd menu item
                    </a>
                </AntDesign.MenuItem>
            </Menu>
        </Overlay>
        <ChildContent>
            Dropdown
        </ChildContent>
    </DropdownButton>
</div>
```

### Further technical details
- AntDesign Nuget Package version
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version

zmayor:
![dropdown_bug](https://user-images.githubusercontent.com/18019463/101569820-2768cf00-3a10-11eb-9973-79ceeea6dbe4.png)

ElderJames:
Thank you for contacting us @zmayor .

I think this problem has been fixed in the PR #925 . Please try again.

zmayor:
Sorry. It wasn't fixed.

anddrzejb:
Tested this against master from 2021-01-15 (commit 81d6603) - I could not recreate it. 
[...]

PR #989, opened 2021-01-16
fix(module:overlay): getFirstChildDomInfo will return passed dom instead of body when no child element exists

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#873 

### 💡 Background and solution
The bug was appearing in `Dropdown` when `ChildContent` was not an element but a simple text:
```razor
<ChildContent>
    Dropdown 
</ChildContent>
```
Here, "Dropdown" is not an element. If it was wrapped in (for example) `<span>`, this bug would not surface.
The problem here is that the underlying `Overlay` is trying to get Childcontent's first child. By default with current approach child content is wrapped in a `div` - so `Overlay` will try to get `div`'s first child. If there is no element inside (for "Dropdown" there is no element), the `Overlay` will be delivered `<body>` element.

The fix is to check if first child is null and when it is, then just return the passed element (for that issue it will be `div`). 

**Possible issues**
This is a quick fix. The `getFirstChildDomInfo` is only used by `Overlay` and `OverlayTrigger`. It does not really return what it promises based on its name. Currently it either returns (correctly) first child  of the passed dom element or top parent (`<body>`). With this PR, it will either return first child of the passed dom element or the dom element itself. It probably should return null, but that would require much more code changes. The possible issues comes from naming - what will a consumer expect and how will current/proposed resolution match/miss consumer's expectations.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/JsInterop/interop.ts
+++ b/components/core/JsInterop/interop.ts
@@ -256,7 +256,9 @@ function slideTo(targetPageY) {
 
 export function getFirstChildDomInfo(element) {
   var dom = getDom(element);
-  return getDomInfo(dom.firstElementChild);
+  if (dom.firstElementChild)
+    return getDomInfo(dom.firstElementChild);
+  return getDomInfo(dom);
 }
 
 export function addClsToFirstChild(element, className) {


ElderJames (approved):
It helps a lot

PR #999, opened 2021-01-18
feat(module: table): add DataIndex parameter

Zonciu:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | add DataIndex parameter |
| 🇨🇳 Chinese | 增加DataIndex参数 |

### DIFF
--- a/components/table/Column.razor
+++ b/components/table/Column.razor
@@ -83,6 +83,10 @@ else if (!IsHeader && RowSpan != 0 && ColSpan != 0)
             {
                 @(Formatter<TData>.Format(Field,Format))
             }
+            else if (GetValue != null)
+            {
+                @GetValue(RowData);
+            }
             else
             {
                 @Field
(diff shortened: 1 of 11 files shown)


components/table/Column.razor
```diff
@@ -83,6 +83,10 @@ else if (!IsHeader && RowSpan != 0 && ColSpan != 0)
             {
                 @(Formatter<TData>.Format(Field,Format))
             }
+            else if (GetValue != null)
+            {
+                @GetValue(RowData);
```
ElderJames, line 88: make `GetValue` compatible with formatter please.

ElderJames (approved):

mehmetuken:
it's support nested list with index ?

sample
```cs
public class DemoDto{
   public string Name {get;set;}
   public List<DemoValueDto> Values {get;set;}
}

public class DemoValueDto{
   public string Test {get;set;}
}
```

DataIndex for DemoDto Values.0.Test like?

ElderJames:
Hi @mehmetuken , it doesn't support that at this PR. Please create another issue that we may do it later.

PR #1004, opened 2021-01-20
Make sider smarter

Hona:
When the sider has a breakpoint for 0 width, ie:
image

When it is opened like:
image

And the user clicks off the menu it will now hide, providing a much better user experience.

### DIFF
--- a/components/layout/Sider.razor
+++ b/components/layout/Sider.razor
@@ -1,7 +1,7 @@
 @namespace AntDesign
 @inherits AntDomComponentBase
 
-<aside class="@ClassMapper.Class" style="@SiderStyles @Style">
+<aside class="@ClassMapper.Class" style="@SiderStyles @Style" @onfocuslost="OnFocusLost" >
     @if (_isCollapsed && CollapsedWidth == 0)
     {
         <span class="ant-layout-sider-zero-width-trigger ant-layout-sider-zero-width-trigger-left"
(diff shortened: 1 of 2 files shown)


ElderJames:
Thank you for contribution @Hona 
But it seems doesn't work, and even throw a exception.

```
fail: Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost[111]
      Unhandled exception in circuit 'IJ1Udu-gLceG5B9lxQt5A0JsOHsqSMhm6ba8pgpEvaw'.
      System.AggregateException: One or more errors occurred. (InvalidCharacterError: Failed to execute 'setAttribute' on 'Element': '@onfocuslost' is not a valid attribute name.)
       ---> System.InvalidOperationException: InvalidCharacterError: Failed to execute 'setAttribute' on 'Element': '@onfocuslost' is not a valid attribute name.
         at Microsoft.AspNetCore.Components.RenderTree.Renderer.InvokeRenderCompletedCallsAfterUpdateDisplayTask(Task updateDisplayTask, Int32[] updatedComponents)
         --- End of inner exception stack trace ---
```

ElderJames:
I have fixed the missing trigger at #1007. Did it solve your problem?

Hona:
Ok I tested the latest pre-release, and it looks great - now you can toggle the sider open + *closed*. Thanks for the work! @ElderJames 
image
image

PR #1008, opened 2021-01-20
fix(module:overlay): Interop delElementFrom() exception on page refresh

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

#### Commit 6ff010c
Removes js event handler firing on window resize from `OverlayTrigger` (this event handler is used to move the overlay as window resizes) on Dispose().
#### Commit be31e33
When `Overlay` is disposed, it tries to remove itself from the DOM. This works well when a page is changed using built in navigation or when a particular component is removed by the framework. 
This does not work well when the page is for example refreshed or new address is typed manually, because the dispose method is trying to remove DOM element that no longer exists.
To recreate:
* start the project `npm start`
* navigate to `https"//localhost:5001`
* navigate to `components/tooltip`
* hover over any control that should have a tooltip (so the tooltip appears)
* press F5
* observe the console; in about a minute (at least in my dev machine it takes exactly a minute) the exception will pop:
image

The problem was fixed by registering a js event `beforeunload` that when fired changes a private flag `bool _isReloading`. If this flag is set to `true`, the call to js `delElementFrom` is not firing during disposal.

### DIFF
--- a/components/core/Component/overlay/Overlay.razor.cs
+++ b/components/core/Component/overlay/Overlay.razor.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Text.Json;
 using System.Threading.Tasks;
 using AntDesign.JsInterop;
 using Microsoft.AspNetCore.Components;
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
LGTM

Issue #919, opened 2020-12-23
Two menu hightlight same time


godshades:
### Describe the bug
Hi, I just try example from this "Open current submenu only": https://antblazor.com/en-US/components/menu
and I edit a little
 ```
<MenuItem Key="1" RouterLink="" RouterMatch="NavLinkMatch.All">Home</MenuItem>
<MenuItem Key="2" RouterLink="counter" RouterMatch="NavLinkMatch.All">Counter</MenuItem>
<MenuItem Key="3" RouterLink="fetchdata" RouterMatch="NavLinkMatch.All">Fetch data</MenuItem>
<MenuItem Key="4" RouterLink="user" RouterMatch="NavLinkMatch.All">Users</MenuItem>
```
but when I choose Home => Counter => OK (highlight each like normal)
but I choose  Home again. it's highlight both menu, I need choose again to hightlight only 1
![image](https://user-images.githubusercontent.com/12244851/102982527-3c2f8180-453d-11eb-94ca-74cbeb098ab4.png)

### Further technical details
- AntDesign Nuget Package version: 0.4.1
- Include the output of `dotnet --info`: .Net Core 5.0
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version: VS Code window

Issue #833, opened 2020-11-27
The menu shows incorrect menu items with routing


CAPCHIK:
When I use Menu with some MenuItem for routing two menu items regularly shows as selected. But the route changed successfully, and basic NavLink works correctly.
![ezgif com-gif-maker (5)](https://user-images.githubusercontent.com/12235437/100439657-a812ed00-3058-11eb-848a-c44abf071594.gif)

### Steps to reproduce (please include code)
The project with error on package 0.4.1 - https://github.com/CAPCHIK/AntBugExample/tree/master

With a nightly package, the same bug exists - https://github.com/CAPCHIK/AntBugExample/tree/nightly

The menu with error:
```razor
<Menu DefaultSelectedKeys=@(new[] { "1" })
      Mode=MenuMode.Inline
      Theme=MenuTheme.Dark>
    <MenuItem Key="1" RouterLink="/" RouterMatch="NavLinkMatch.All">
        <Icon Type="home" Theme="outline"/>
        <span>Home</span>
    </MenuItem>
    <MenuItem Key="2" RouterLink="/counter" RouterMatch="NavLinkMatch.All">
        <Icon Type="unordered-list" Theme="outline" />
        <span>Counter</span>
    </MenuItem>
    <MenuItem Key="3" RouterLink="/fetchdata" RouterMatch="NavLinkMatch.All">
        <Icon Type="plus" Theme="outline"/>
        <span>Fetch data</span>
    </MenuItem>
</Menu>

All steps to reproduce can be found in 5 simple commits to the repository.
```

### Further technical details
- AntDesign Nuget Package version: `0.4.1` | `0.5.0-nightly-2011270850`
- `dotnet --info`
```bash
.NET SDK (reflecting any global.json):
 Version:   5.0.100
 Commit:    5044b93829

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19041
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.100\

Host (useful for support):
  Version: 5.0.0
  Commit:  cf258a14b7

.NET SDKs installed:
  3.1.401 [C:\Program Files\dotnet\sdk]
  3.1.403 [C:\Program Files\dotnet\sdk]
  5.0.100 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
[...]

Issue #737, opened 2020-10-28
Infinite loop of Menu


houliren:
当多个MenuItem 的RouterLink 设置同一个url时，整个menu会出现无限循环。代码 如下：
 <MenuItem RouterLink="@child.WebUrlOrArgs" RouterMatch="NavLinkMatch.All" Key="@child.Id.ToString()">@child.Name</MenuItem>
当两个menuitem的child.WebUrlOrArgs是同一个的时候，menu会进入到无限循环的渲染中。虽然这种业务需求可能不多，但是一旦配置的时候错误了，整个页面会进入的死循环的假死状态，无限未响应。

When the `RouterLink` property of multiple `MenuItem`s are set up as the same url, the whole Menu control would have infinite loop. The code is as below:
```
<MenuItem RouterLink="@child.WebUrlOrArgs" RouterMatch="NavLinkMatch.All" Key="@child.Id.ToString()">@child.Name</MenuItem>
```
If we set up the `child.WebUrlOrArgs` of two `MenuItem`s as the same value, the `Menu` control also gets an infinite loop so the page will be freezing as well.

jrhodnik:
I just encountered this bug as well. `MenuItem`s with `RouterLink`s that start the same cause an infinite loop freeze, so long as the page URLs are routable. 

Example: Click one of the "Zero" entries and it'll freeze. The pages must be routable to prevent page refresh though.
```razor
<Menu>
    <MenuItem Key="2" RouterLink="zero">Zero</MenuItem>
    <MenuItem Key="3" RouterLink="zero/something">Zero Again</MenuItem>
    <MenuItem Key="4" RouterLink="one/something">One</MenuItem>
    <MenuItem Key="5" RouterLink="two/something">Two</MenuItem>
</Menu>
```

PR #1027, opened 2021-01-21
fix(module:menu): fix loop on same route & performance & duplicate highlight

anddrzejb:
🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#737 
#833 
#919 
#1014 

### 💡 Background and solution
All changes are directly connected with issues. 
Plus performance optimization, as stated below.

During tests I found out that single click was causing cascading calls that were causing multiple unnecessary refreshes. After some serious debugging and trial and error I think I got to the point it is only doing necessary calls. Just to give a ballpark figure - with 5 menus, single click was causing 5 refreshes of each menu on average (my total was 29 refreshes overall). With this PR only changing menus are refreshing (so 2 changes in total - what is weird, 3 refreshes if order of menu clicking is reversed).

### DIFF
--- a/components/menu/Menu.razor
+++ b/components/menu/Menu.razor
@@ -1,7 +1,7 @@
 ﻿@namespace AntDesign
 @inherits AntDomComponentBase
 
-<CascadingValue Value="this" IsFixed="@true">
+<CascadingValue Value="this" IsFixed>
     <ul class="@ClassMapper.Class" style="@Style" id="@Id" direction="ltr" role="menu" @ref="Ref">
         @ChildContent
     </ul>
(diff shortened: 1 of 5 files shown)


ElderJames (approved):
LGTM

ElderJames:
We also need to add the expand/collapse animation of the submenu.

anddrzejb:
Yeah, I noticed that. I will try to do some work in that area.

Issue #851, opened 2020-12-03
Form EditContext is not updated when Model changes


NPadrutt:
### Describe the bug
Currently the edit context of the Form is created `OnInitialized`. When the model later changes the EditContext of the form does refer to the wrong model and validation creates results like this:

![image](https://user-images.githubusercontent.com/1764367/101100664-30e4d800-35c7-11eb-90cf-25c7f5fd51a9.png)

My understanding is that the EditContext would have to be built in the `OnParameterSet `Method. Or would that have undesired side effects? 

### Further technical details
- AntDesign Nuget Package version: 0.4.1
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version

wangjunjx8868:
you should 
if(model !=null)
{
  <Modal>
  ....
</Modal>
}

NPadrutt:
That was the work around I used in the end. But it seemed a bit redundant since I already have a IsOpen Property. But is that the intended way of using it then?

Issue #947, opened 2021-01-08
Validator message cannot be cleared in a form and the validators do not work when form model is assigned to a new object.


motl2000:
Background:
I have a modal dialog with form. The form will perform creation and update.
There are several fields with validator. The validator will be triggered when click submit button by calling form.validate ().

Issue 1: The error message cannot be cleared by calling form.Reset ().
 1. Open the dialog form. Click submit button. The validation message will display. =>It's OK.
 2. Close the dialog by clicking cancel button (form.visible=false). The dialog form is closed. =>It's OK.
 3. Open the dialog form again. The validation message still there even form.Reset () is called.=>It's NOT OK.

Issue 2: The validators do not work when form model is assigned to a new object.
Form model is initialized with a new object when OnInitialized ().
After that, if the form model is assigned to another new object, the validation does not work. It still link to the old form model.

tonyyip1969:
You may try to add a @ref to the form and the following code.

`<Form @ref="FormRef" Model="@SampleData">`

```
protected override void OnParametersSet()
{
         if (FormRef != null)
         {
             FormRef.Validate();
         }
}
```

motl2000:
> You may try to add a @ref to the form and the following code.
> 
> `<Form @ref="FormRef" Model="@SampleData">`
> 
> ```
> protected override void OnParametersSet()
> {
>          if (FormRef != null)
>          {
>              FormRef.Validate();
>          }
> }
> ```

Thanks for your reply. But it's not working. The issue is not "the validation is not triggered", but the validation message cannot be removed after the form is reset.

tonyyip1969:
I believe that the issue has been fixed in the latest release v0.5.3, try it out. 😊

motl2000:
> I believe that the issue has been fixed in the latest release v0.5.3, try it out. 😊

Hi, Tony.

I am sorry to say the issue still there.
I installed the latest version 0.6.0-nightly-2101140256 because there is an issue on the rangepicker in 0.5.3. 
I am not sure it's about the version or the issue is not fixed.

I'd like to repeat the issue is: the validation message cannot be cleared after the form is reset.

Issue #982, opened 2021-01-13
Form Validation


dblleaf:
The form Validator can't validate then the Model of Form is changed in an asynchronous method.
The code is as follow:
```razor

<h3>EditForm</h3>
<EditForm Model="@Model">
  <DataAnnotationsValidator />
  <ValidationSummary />
  <p>
    <label>User Name :</label>
    <InputText @bind-Value="@Model.UserName" />
    <ValidationMessage For="@(() => Model.UserName)" />
  </p>
  <p>
    <label>Password :</label>
    <InputText @bind-Value="@Model.Password" />
    <ValidationMessage For="@(() => Model.Password)" />
  </p>
  <button type="submit">Save</button>
</EditForm>
<h3>AntDesign.Form</h3>
<Form Model="@Model">
  <Validator>
    <DataAnnotationsValidator />
  </Validator>
  <ChildContent>
    <FormItem Label="User Name">
      <Input @bind-Value="@context.UserName" />
    </FormItem>
    <FormItem Label="Password">
      <InputPassword @bind-Value="@context.Password" />
    </FormItem>
    <button type="submit">Save</button>
  </ChildContent>
</Form>
@code {
  LoginModel Model = new LoginModel();

  protected override async Task OnInitializedAsync()
  {
    await Task.Delay(1);
    this.Model = new LoginModel
    {
      UserName = "Tom",
      Password = "123456",
    };
    await base.OnInitializedAsync();
  }
}

```
The screenshot
![image](https://user-images.githubusercontent.com/15872501/104419483-7c6da700-55b3-11eb-95d4-d8f4e803a9f4.png)

So, I had to use tools like AutoMapper or manually set the properties of original model.

Issue #516, opened 2020-08-18
If form non-empty validation should be disabled when input has been disabled


bingbingtalk:
表单中其中一下输入框在禁用状态时能否在自动验证时不进行非空等验证

If form non-empty validation should be disabled when input has been disabled.

anddrzejb:
@bingbingtalk I do not think this can be done directly by the `AntDesign`. The problem is that when validation is executed (it is done in `Required` validation attribute) the method does not know anything about controls bound to the properties. And I do not think they should (imagine a scenario when a single property is bound to several different DOM elements). However this can be done with a custom validation attribute and once PR #1035 is merged. Example:
```razor
@using System.ComponentModel.DataAnnotations;
<h3>AntDesign.Form</h3>
<Form Model="@Model" @ref="form">
    <Validator>
        <DataAnnotationsValidator />
    </Validator>
    <ChildContent>
        <FormItem Label="User Name">
            <Input @bind-Value="@context.UserName" Disabled="@Model.Disabled" />
        </FormItem>
        <FormItem Label="Password">
            <InputPassword @bind-Value="@context.Password" />
        </FormItem>
        <button type="submit">Save</button>
    </ChildContent>
</Form>
<br />
<Checkbox Checked="@Model.Disabled" CheckedChange="@ToggleChecked">Disable username</Checkbox>

@code {
    AntDesign.Form<LoginModel> form;

    LoginModel Model = new LoginModel();

    protected override async Task OnInitializedAsync()
    {
        this.Model = new LoginModel
        {
            UserName = "Tom",
            Password = "123456",
        };
        await base.OnInitializedAsync();
    }

    public class LoginModel
    {
        [RequiredUnlessDisabled]
        public string UserName { get; set; }
        [Required]
        public string Password { get; set; }
        //adding this property will actually allow to check the state of input box during validation
        public bool Disabled { get; set; }
    }

    private void ToggleChecked(bool value)
    {
        Model.Disabled = value;
        if (Model.Disabled)
            Model.UserName = null;
        //validation needs to be reset once the form input box disability is changed
        //this is part of PR #1035 that will need to be merged for this solution to work
        form.ValidationReset();
    }

    public class RequiredUnlessDisabled : RequiredAttribute
    {
        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
        {
            var model = (LoginModel)validationContext.ObjectInstance;
            if (model.Disabled)
                return ValidationResult.Success;

[...]

PR #1035, opened 2021-01-23
fix(module:form): form.reset() will make new editContext

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#947 

### 💡 Background and solution
`Microsoft.Aspnetcore.Componets.Forms.Editcontext` does not really have any way of resetting the validation messages ([source](https://github.com/dotnet/aspnetcore/issues/24563)). So in `form.Reset()` method I assign a `new EditContext(model)` method and that actually resets validation messages.


ElderJames:
Thank you.
Does it fix the other form issues？

anddrzejb:
Not yet. I was trying to address milestone issue. You can wait witj merge, I will try to have a lool at the other form validation issue.

anddrzejb:
@ElderJames Commit 61918dc fixes issues #851 & #982 (I imagine it also fixes #991 as it was marked as duplicate of #982).

anddrzejb:
@ElderJames In commit fe3a0d6 I added single extra method `ValidationReset()`. This is the result of my works on #516. In all honesty, I do not think we can make anything in the project to comply with #516 request. The problem is that the request is not to validate when textbox is disabled, but `ValidationAttribute` does not have access to bounded text box (and actually I do not think it should). So when validation is running, it has only access to the model it validates. The only other solution is to add a `bool Disabled` to the model and the use a custom validation attribute. This is when `ValidationReset()` comes in as a necessary method (and also `Reset()` cannot be used). 
Anyway, if this is accepted, I have a ready sample for #516 to explain how the request can be achieved.

Also - test are failing on tabs again. I did not change anything in tabs....I will look into that...

ElderJames:
@anddrzejb Thanks for fixing so many issues. The check is pass while I re-run it.

anddrzejb:
Got lucky here - a lot of the issues were simple fix & same fix applied to most of them. 
I will also try to fix #1036 so please wait with the merge.

anddrzejb:
@ElderJames Apparently there is nothing to be fixed in #1036 so I think you can merge this PR if you are fine with it.

ElderJames (approved):
Thank you!

Issue #1047, opened 2021-01-26
Bug: Input Disabled Attribute


shirleen1234:
### Describe the bug
Still able to focus and change value on **Input** field with **Disabled** Attribute by **pressing tab key** on keyboard.

![demo gif](https://user-images.githubusercontent.com/47659614/105826492-c5821a00-5ffb-11eb-9f17-a642d3cf81d6.gif)

### Steps to reproduce (please include code)
```
<FormItem Label="Name">
	<Input @bind-Value="@context.Name" AutoFocus="true" />
</FormItem>
<FormItem Label="Age">
	<AntDesign.InputNumber @bind-Value="@context.Age" Min="1" Max="99" Step="1" Disabled="true"/>
</FormItem>
<FormItem Label="Address">
	<TextArea Placeholder="Address lines" MinRows="3" MaxRows="5" @bind-Value="@context.Address" Disabled="true" />
</FormItem>
```

### Further technical details
- AntDesign Nuget Package 0.5.3

PR #1048, opened 2021-01-26
fix: Input, InputNumber, TextArea disabled attribute fix

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#1047 

### 💡 Background and solution
Missing `disabled` attribute on `input` element. 
There is a certain assumption here (visible in the changes in `Input.cs`): if `disabled` attribute is set to `true` in any fashion, it will always prevail.

BTW: @ElderJames I noticed that for example `Cascader` component does not have `Disabled` property. Is this by design?

### DIFF
--- a/components/input-number/InputNumber.razor
+++ b/components/input-number/InputNumber.razor
@@ -13,6 +13,6 @@
     </div>
     <div class="ant-input-number-input-wrap">
         <input @ref="Ref" role="spinbutton" aria-valuemin="@Min" aria-valuemax="@Max" autocomplete="off" max="@Max" min="@Min" step="@Step"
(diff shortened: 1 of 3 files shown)


ElderJames (approved):

ElderJames:
> BTW: @ElderJames I noticed that for example Cascader component does not have Disabled property. Is this by design?

@anddrzejb I think it is we missing them.

Issue #1065, opened 2021-02-02
Select:  The OnSelectedItemsChanged event is not triggered if the bind-value is not used


ma-hua:
The OnSelectedItemsChanged event is not triggered if the bind-value is not used.

I want to know if this is a design or a flaw

### Steps to reproduce 

[https://antblazor.com/zh-CN/components/select](https://antblazor.com/zh-CN/components/select)

SimpleSelect 

### Further technical details
- AntDesign Nuget Package version 0.6.0

ma-hua:
In addition, the SelectedItemsChanged event is triggered when the page is loaded, which was not the case in previous versions

anddrzejb:
Just to be sure - you mean the event `OnSelectedItemsChanged`?

ma-hua:
> Just to be sure - you mean the event `OnSelectedItemsChanged`?
yes, you can go to the link I mentioned above and you will see that the SimpleSelect OnSelectedItemsChanged event is not triggered.

anddrzejb:
Ok, I can see that in the docs and I recreated it in my repo. I will try to fix it in my next PR.

Issue #878, opened 2020-12-09
AllowCustomTags  / OnCreateCustomTag does not work


szmalec:
### Describe the bug
When I try to use `Mode="@AntDesign.Select.Internal.SelectMode.Tags.ToString()" AllowCustomTags OnCreateCustomTag="@(tag => Console.WriteLine($"OnCreateCustomTag: {tag}"))" EnableSearch` parameters in the `Select` component, it turns out that `OnCreateCustomTag` is never called.

### Steps to reproduce (please include code)
To reproduce use code from [demo page](https://antblazor.com/en-US/components/select#components-select-demo-tags) (_Tags_ section) and modify it by adding mentioned parameters (`EnableSearch` is already there...).

### Further technical details
0.5.0-nightly-2012081156

### Additional notes
Before [`Select` component was refactored](https://github.com/ant-design-blazor/ant-design-blazor/pull/800), it was possible to display suggested tags along with the ability to select new tags...
Looking at the source code:
![image](https://user-images.githubusercontent.com/7856921/101627585-d7453900-3a1e-11eb-95c9-4c265284305f.png)
it seems to me that the problem with `AllowCustomTags` / `OnCreateCustomTag` may be related to another problem [that I reported](https://github.com/ant-design-blazor/ant-design-blazor/issues/877)...

Issue #877, opened 2020-12-09
EnableSearch does not work in Select for tags or multiple Mode


szmalec:
### Describe the bug
The `EnableSearch` option does not work in the [`Select` component](https://github.com/ant-design-blazor/ant-design-blazor/pull/800) when `Mode="tag"` or `Mode="multiple"` is set.

### Steps to reproduce (please include code)
Use [demo page](https://antblazor.com/en-US/components/select#components-select-demo-multiple) - eg. "multiple selection" sample.

### Further technical details
0.5.0-nightly-2012081156

Issue #662, opened 2020-10-01
[select] DefaultValue does not take effect when the select component enables remote search


Seanxwy:
DefaultValue does not take effect when the select component enables remote search。
By default, the data is loaded, and contains the data required by DefaultValue, but it is not selected。

select 组件在开启远程搜索的时候，DefaultValue 不生效，默认拥有DefaultValue 需要的数据项，但是不会选中。

PR #1087, opened 2021-02-04
fix(module: select): multiple fixes and optimizations

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#257 
#594 
#662 
#877 
#878 
#1065 
and a number of other issues discovered during fixing.
### 💡 Background and solution
I started fixing the `Select` component because I wanted to fix searchable multiple choice and tags + tokenization. This is fixed now, but there are a lot of changes. There is a `Hashset` structure that stores all selectable items. I noticed this structure was being traversed very often in multiple different methods. Often the algorithm time complexity  for a method was O(2n), O(3n), O(n*n). I decided to introduce new variables to store the items for different scenarios and different parts of the component. I managed to reduce the complexity to mostly O(n) or even O(1). In several places I moved to `foreach` loop instead of `linq` loops, just to cover more ground in single loop instead of calling `linq` 2 or 3 times.
There is still room for optimization. What needs to be done is a very careful tracking of `[Parameter]` properties to see which of them are causing unnecessary component re-rendering. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | removed `AllowCustomTags` and `OnCreateCustomTag` <br>    added `PrefixIcon`         |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/JsInterop/JSInteropConstants.cs
+++ b/components/core/JsInterop/JSInteropConstants.cs
@@ -87,6 +87,8 @@ namespace AntDesign
 
         public static string BindTableHeaderAndBodyScroll => $"{FUNC_PREFIX}bindTableHeaderAndBodyScroll";
         public static string UnbindTableHeaderAndBodyScroll => $"{FUNC_PREFIX}unbindTableHeaderAndBodyScroll";
+        public static string AddPreventCursorMoveOnArrowUp => $"{FUNC_PREFIX}addPreventCursorMoveOnArrowUp";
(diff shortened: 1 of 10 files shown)


ElderJames (approved):
Awesome! It works well.

Issue #1093, opened 2021-02-05
DatePicker Internationalization es-ES Year Issue


georgeemr:
When Spanish internationalization is being used, the year is not displayed.
![image](https://user-images.githubusercontent.com/3822009/106991935-f02b4b80-673c-11eb-8412-180f5ab97b0e.png)
![image](https://user-images.githubusercontent.com/3822009/106991945-f5889600-673c-11eb-9a3f-4b511352f5ac.png)

ElderJames:
Thank you for contacting us @georgeemr 

@anddrzejb Could you please fix the locale of the datepicker?

anddrzejb:
I've been sitting on a solution for this sińce yesterday (this and #1044). I just need to polish it up and a PR is coming up real soon.

ElderJames:
That's great!

anddrzejb:
Just one question before I go to research this - one of the problems I found out is that LocaleProvider is using `CultureInfo.DefaultThreadCurrentUICulture`. Is there any reason why it is not `CultureInfo.CurrentUICulture`? Because I think in serverside when there are multiple threads, this screws up things...

ElderJames:
I just follow the document. https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-3.1&WT.mc_id=DT-MVP-5003987#localization

anddrzejb:
This is not working for server version. What I mean by that - I get server's culture info instead browser's. I guess then I will use the detection you made between wasm and server and for wasm keep as it is and for server use `CurrentUiCulture`.

ElderJames:
That's fine.

anddrzejb:
Well, that did not work. `LocaleProvider` is a static class and to use `IsBrowser()` I need to inject JsRuntime. However, I noticed that using `CurrentUiCulture` works for both server and wasm. I think `DefaultThreadCurrentUICulture` is needed if you want to change the culture in wasm. But for retrieving, `CurrentUiCulture` works - and it is better because it works for both wasm and server.

ElderJames:
I think `IsBrowser()` can be a static method. But if `CurrentUiCulture` is better, we can switch to it.

anddrzejb:
It looks to be working in both cases so I see no reason to complicate matters with `IsBrowser()`.

ElderJames:
You are right.

Issue #1044, opened 2021-01-25
DatePicker and CultureInfo


true-perfect-code:
### Describe the bug
I need to display the date in German/Swiss format. I just imagined that it would work if I would take CulturInfo Property, but it doesn't work.

`<FormItem Label="Entered on">
   <DatePicker  CultureInfo="@System.Globalization.CultureInfo.CurrentCulture" @bind-Value="@C.enteredDate" />
</FormItem>`

This is possible only if you take Format Property.

`<FormItem Label="Entered on">
   <DatePicker  Format="dd.MM.yyyy" @bind-Value="@C.enteredDate" />
</FormItem>`

![image](https://user-images.githubusercontent.com/45635963/105725095-d44bcc80-5f28-11eb-8512-3cb659468439.png)

This is good in that it gives you a workaround, because if you want to provide a form for users from different countries, then the date formatting has to run automatically. Unfortunately, this can then only be done with CulturInfo (and by reading it from the client).

By the way, the same error is present here as with Modal (see pr1012 and pr1014).

### Steps to reproduce (please include code)
`<FormItem Label="Entered on">
   <DatePicker  CultureInfo="@System.Globalization.CultureInfo.CurrentCulture" @bind-Value="@C.enteredDate" />
</FormItem>`

### Further technical details
- AntDesign Nuget Package version 0.5.0-nightly-2012300110
- The IDE Community 2019 Version 16.8.3

ElderJames:
Thank you for contacting us @true-perfect-code 

In fact, we have done a certain amount of i18n support (event German, see #803 )  , maybe in datepicker is not good enough.
We will continue to improve in the near future, but also welcome you to participate in contributing!

NPadrutt:
@ElderJames are you sure this is fixed?  I tested this on 0.7.0-nightly-2102090212 (the latest as of today) and it looks the same to me:
![image](https://user-images.githubusercontent.com/1764367/107430679-c64f9b80-6b25-11eb-809f-0ef17802f8b3.png)

PR #1097, opened 2021-02-05
fix(module:datepicker): incorrect date format strings fix

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#1044 
#1093 

### 💡 Background and solution
Most of locale `.json` file had incorrect string template (`YYYY` instead of `yyyy` and `D` instead of `d`).
`LocaleProvider` was using `CultureInfo.DefaultThreadCurrentUICulture` for detection of current culture. But this is not working well with server version - there is a high possibility it will pick up server's UI instead of browser's UI. A switch was done to `CultureInfo.CurrentUICulture` for detection.
In `DatePicker` if the `CultureInfo` attribute was used without setting `Locale` attribute, still only `CultureInfo.DefaultThreadCurrentUICulture` was being picked up. The fix is to detect if parameter `CultureInfo` was set outside and force load the `Locale` with passed `CultureInfo`. This exposed a situation when culture file does not exist (in my case I was trying to use `el-CY` and it was crashing). Fallback logic was added and default locale is picked up if no locale `.json` file is found.


ElderJames (approved):

ElderJames:
Awesome! Thank you.

PR #1100, opened 2021-02-05
fix(module:datepicker): for not nullable - on clear set to defaults

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Due to the strongly typed nature of C#, `DatePicker` can get either nullable or not nullable dates. When not nullable are used, clearing the values was inputing min value for `DateTime`. There is not much we can do about (we can decide to take current date for example). This PR fixes the not nullable clearing scenario when defaults are set - in such case, date picker will assume defaults instead of `DateTime.Min`. For nullable the picker will be cleared.

### DIFF
--- a/components/date-picker/DatePicker.Razor.cs
+++ b/components/date-picker/DatePicker.Razor.cs
@@ -174,7 +174,11 @@ namespace AntDesign
         public override void ClearValue(int index = 0)
         {
             _isSetPicker = false;
-            CurrentValue = default;
+
+            if (!IsNullable && DefaultValue != null)
+                CurrentValue = DefaultValue;
+            else
+                CurrentValue = default;
             Close();
         }
 
--- a/components/date-picker/RangePicker.razor.cs
+++ b/components/date-picker/RangePicker.razor.cs
@@ -207,8 +207,17 @@ namespace AntDesign
             _isSetPicker = false;
 
             var array = CurrentValue as Array;
-            array.SetValue(default, 0);
-            array.SetValue(default, 1);
+            if (!IsNullable && DefaultValue != null)
+            {
+                var defaults = DefaultValue as Array;
+                array.SetValue(defaults.GetValue(0), 0);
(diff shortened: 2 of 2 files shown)


ElderJames:
I think it's more like a reset method..

anddrzejb:
Yes it is. Basically I retain the original functionality TValue is nullable. But if it is not, then we cannot clear it up - it will take `default`, which is minimum.
Without this change, if you specify:
```razor
<RangePicker TValue="DateTime[]" DefaultValue="@(new DateTime[] { DateTime.Now.AddYear(-1), DateTime.Now })" />
```
you will get this result
datepicker_Defaults

But I think it is better to set it back to defaults if they exist. Again, this is **only** if TValue is not nullable and defaults exist.

ElderJames:
What if the value is made minimum when cleared, and the input box is empty?

anddrzejb:
I guess this could be done, but I think this might send a wrong message to anyone consuming the date picker. So in this scenario, the date picker would show empty but the values (if `@bind-Value="@values"`), would have `DateTime.Min`. To be honest I wouldn't like a surprise like that in production.

ElderJames:
I'm not sure which is people's want.

anddrzejb:
Ok, I understand. I made this change because this is my use case and to me seemed most logical. Please remember that ppl will still have a choice - they will be able to empty completely the fields if they decide to choose `DateTime?[]` type for value. My scenario is only for `DateTime[]` and with `DefaultValue` existing.

ElderJames (approved):

anddrzejb:
Wow, thanks 😄 that was faster than I anticipated.

Issue #1081, opened 2021-02-04
Overlay boundary detection and orientation adjustment/overlay的边界检测和方向调整


mutouzdl:
Overlay boundary detection and orientation adjustment

-----------------

overlay的边界检测和方向调整

ElderJames:
FYI
- https://popper.js.org/

PR #1109, opened 2021-02-07
feat(model: overlay): add boundary detection adjustment

mutouzdl:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#1081 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     add overlay boundary detection and orientation adjustment      |
| 🇨🇳 Chinese |     增加overlay的边界检测和方向调整      |

### DIFF
--- a/components/core/Component/overlay/Overlay.razor.cs
+++ b/components/core/Component/overlay/Overlay.razor.cs
@@ -196,8 +196,8 @@ namespace AntDesign.Internal
             Element overlayElement = await JsInvokeAsync<Element>(JSInteropConstants.GetDomInfo, Ref);
             Element containerElement = await JsInvokeAsync<Element>(JSInteropConstants.GetDomInfo, Trigger.PopupContainerSelector);
 
-            int left = GetOverlayLeftWithBoundaryAdjust(trigger, overlayElement, containerElement);
-            int top = GetOverlayTopWithBoundaryAdjust(trigger, overlayElement, containerElement);
+            int left = await GetOverlayLeftWithBoundaryAdjust(trigger, overlayElement, containerElement);
+            int top = await GetOverlayTopWithBoundaryAdjust(trigger, overlayElement, containerElement);
 
             int zIndex = await JsInvokeAsync<int>(JSInteropConstants.GetMaxZIndex);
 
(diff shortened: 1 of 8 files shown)


ElderJames (approved):
LGTM

anddrzejb:
@mutouzdl Sorry to bother you about this, but I am trying to work out the overlay boundary calculation. I cannot figure out what is the meaning of `TriggerBoundaryAdjustMode.InScroll`. Do you remember what was that? I think I possibly am able to simplify the calculations, but I cannot figure out what that enum value really stands for (or how different calculations should be). I cannot find it used anywhere...

mutouzdl:
@anddrzejb It's a little hard to understand, and I'll probably give you two examples to illustrate it.

Example1: In InScroll mode, in this case, the menu will appear on top.
This is because the browser is not high enough to fully display the menu below (even if the scroll bar is scrolled)
image


Example2: In InScroll mode, the following situation will display the menu below.
Although the browser height still doesn't seem to display the full menu, if you scroll the scroll bar, it will, so in this case, the menu will still expand below
image

If anything is unclear, I will try to explain further.

anddrzejb:
@mutouzdl Thanks for the explanation! I think I understand now. So, we have these scenarios for calculations:
- `TriggerBoundaryAdjustMode.InView` - the default, the viewport boundaries are the boundaries we use for calculation. We try to fit the overlay so it is always fully visible in the viewport
- `TriggerBoundaryAdjustMode.InScroll` - the document boundaries are the boundaries used for calculation. So even if the overlay is outside of the viewport, the overlay may still be shown as long as it does not "overflow" the document boundaries.

mutouzdl:
@anddrzejb Yes, your interpretation is correct, and your summary is very accurate!

anddrzejb:
I will add this or similar explanation to the docs once I finish.

anddrzejb:
@mutouzdl I have another question. I am trying to wrap my head around parameter `ArrowPointAtCenter`. What I can see here, this seems to me that it is doing exactly the same job as when setting one of the placements to the center placements (`Left`, `Right`, `Bottom`/`BottomCenter` or `Top`/`TopCenter`). Do I understand this correctly, or there is some more to this parameter that I am missing?

anddrzejb:
@mutouzdl  I thought so. I was a bit confused by the calculation based on the arrow shifts. But the redesign will just switch internally to appropriate cantered placement I guess.

Issue #1111, opened 2021-02-07
The component Select property DefaultActiveFirstOption is invalid


fan0217:
The component Select property **DefaultActiveFirstOption** is invalid

anddrzejb:
Apparently it is under `DefaultActiveFirstItem`... For tests purposes you can use it, but a PR is coming that will fix it to `DefaultActiveFirstOption`.

PR #1115, opened 2021-02-07
fix(module:select): property rename to follow docs

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#1111 

### 💡 Background and solution
1. Renamed `DefaultActiveFirstItem` to `DefaultActiveFirstOption` following the docs (it is in sync with react antDesign)
2. When current locale are of a country that uses comma as decimal separator, the calculated css width was delivered with a comma (for example 333,123123). Css was not understanding that and was ignoring the value.
3. Translated Chinese comment.
4. Fix similar to #662 that was applier to `DefaultActiveFirstOption`
5. Changing bound object to value outside of the `Select` component wasn't changing selection in the component.

### DIFF
--- a/components/modal/config/ModalClosingEventArgs.cs
+++ b/components/modal/config/ModalClosingEventArgs.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-using Microsoft.AspNetCore.Components.Web;
+﻿using Microsoft.AspNetCore.Components.Web;
 
 namespace AntDesign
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
LGTM

Issue #1110, opened 2021-02-07
MenuItem RouterLink is not in the right state


woodsun2018:
### Describe the bug
Click the Sub MenuItem, The Main MenuItem's activited status is not right.

### Steps to reproduce (please include code)
Create a Blazor Server Net 5.0 project, use the NavMenu.razor code show in follow, delete the NavMenu.razor.css file，and remove sidebar for div class="sidebar" in the file MainLayout.razor
```
<Menu Mode="MenuMode.Inline"
      Accordion
      Style="width:256px ;">
    <SubMenu Key="sub5" Title="MainMenu">
        <MenuItem Key="50" RouterLink="" RouterMatch="NavLinkMatch.All">Home</MenuItem>
        <MenuItem Key="51" RouterLink="counter">Couter</MenuItem>
        <MenuItem Key="52" RouterLink="fetchdata">Fetchdata</MenuItem>
    </SubMenu>
</Menu>
```

### Exceptions (if any)
Press F5 to run the software, in Edge( Chromium 88.0.705.63) browser, When I clikc the Counter sub menu, the MainMenu is not in actived status,  then I click the Fetchdata sub menu, the MainMenu is not in actived status, then I click the Counter sub menu again, the MainMenu is in actived status, click any other sub menu again, the MainMenu is in actived sometime.

### Further technical details
- AntDesign Nuget Package version 0.6.0
- Include the output of `dotnet --info`
- C:\Users\chenw>dotnet --info
.NET SDK (反映任何 global.json):
 Version:   5.0.102
 Commit:    71365b4d42

运行时环境:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.102\

Host (useful for support):
  Version: 5.0.2
  Commit:  cb5f173b96

.NET SDKs installed:
  2.2.402 [C:\Program Files\dotnet\sdk]
  5.0.102 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.24 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #1121, opened 2021-02-09
fix(module:select): new tag item label and value fix

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
When `Mode="tags"` set, newly created `SelectOptionItem` was not being properly updated. `Item` property was only created with initial letter and no updates were done further.

<s>Edit: I accidentally pushed to the same branch fixes for #1110, so I will leave them here as well... `Menu` except the fix got some extra optimization. I also noticed that accordion is not working but since @mutouzdl is doing some heavy work in that area, I will not be touching it. </s> _(moved to PR #1134)_

Edit2: 
1. Commit ac85729 is fixing FireFox issue. I discovered it in select component - for multiple/tags mode, FireFox was rising "Permission denied to access property "toJSON". A number of trial-and-error led me to a property `originalTarget` of dom element. So I am omitting that...
2. Commit 32cd7d4 is fixing the `PrefixIcon` for `Select` component. It wasn't playing nice with placeholder/input cursor. Also it was not shown when selected items were visible in the select box (for tags/multiple).

### DIFF
--- a/components/core/JsInterop/interop.ts
+++ b/components/core/JsInterop/interop.ts
@@ -156,10 +156,12 @@ export function getBoundingClientRect(element) {
 export function addDomEventListener(element, eventName, preventDefault, invoker) {
   let callback = args => {
     const obj = {};
-    for (let k in args) {
-      obj[k] = args[k];
-    }
-    let json = JSON.stringify(obj, (k, v) => {
+    for (let k in args) {                
+      if (k !== 'originalTarget') { //firefox occasionally raises Permission Denied when this property is being stringified
+        obj[k] = args[k];
+      }
+    }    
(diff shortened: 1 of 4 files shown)


ElderJames:
@anddrzejb Could you please cherry-pick the menu commit to another branch?

anddrzejb:
Ok, will do that.

ElderJames (approved):

Issue #952, opened 2021-01-08
PasswordInput


lingrepo:
PasswordInput has no focused effect
![image](https://user-images.githubusercontent.com/63399761/104015296-5e88f680-51ef-11eb-9ad9-0709f3e733b4.png)
![image](https://user-images.githubusercontent.com/63399761/104015254-4f09ad80-51ef-11eb-89a0-ba44349e9179.png)

PR #1146, opened 2021-02-19
fix(module:inputpassword): focus fix

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#952

### 💡 Background and solution
For password `<input>` tag is wrapped in a span
```html
<span class="ant-input-affix-wrapper  ant-input-password">
```
That span (base on react ant design) should get an extra class on focus `ant-input-affix-wrapper-focus`. My fix is including this class when input password receives focus.


anddrzejb:
@ElderJames Hold with merge. Same problem is with `Search` component and input with prefix/suffix.

anddrzejb:
The commit  333673c solved the problem of focus for password, search and input with AllowClear or Prefix or Suffix (all of them were not focusing). It is a unified solution, meant for all components that use `Input.cs` as their base.

ElderJames (approved):
Thanks.

Issue #598, opened 2020-09-14
[TextArea] Multiple Textareas cannot be displayed according to Minrows normally


Seanxwy:
Multiple Textareas cannot be displayed according to Minrows normally,Randomly present the TextArea as an Input style

ElderJames:
Hello @yanxiaodi , please leave your repro code.

yuanxiongwei:
```
<Row Style="margin-bottom:10px">
            <AntDesign.Col Span="24">
                <Text Strong="true">针对老师维度解说</Text>
            </AntDesign.Col>
        </Row>
        <Row>
            <AntDesign.Col Span="20">
                <FormItem Label="得分说明">
                    <TextArea @bind-Value="@context.ExplainForTeacher" MinRows="3" />
                </FormItem>
            </AntDesign.Col>
        </Row>
        <Row>
            <AntDesign.Col Span="20">
                <FormItem Label="分析说明">
                    <TextArea @bind-Value="@context.AnalysisForTeacher" MinRows="3" />
                </FormItem>
            </AntDesign.Col>
        </Row>
        <Row>
            <AntDesign.Col Span="20">
                <FormItem Label="专家建议">
                    <TextArea @bind-Value="@context.SuggestForTeacher" MinRows="3" />
                </FormItem>
            </AntDesign.Col>
        </Row>
        <Row Style="margin-bottom:10px">
            <AntDesign.Col Span="24">
                <Text Strong="true">针对学生的综合报告</Text>
            </AntDesign.Col>
        </Row>
        <Row>
            <AntDesign.Col Span="20">
                <FormItem Label="综合说明">
                    <TextArea @bind-Value="@context.ReportExplainForStudent" MinRows="3" />
                </FormItem>
            </AntDesign.Col>
        </Row>
        <Row>
            <AntDesign.Col Span="20">
                <FormItem Label="综合分析">
                    <TextArea @bind-Value="@context.ReportAnalysisForStudent" MinRows="3" />
                </FormItem>
            </AntDesign.Col>
        </Row>
        <Row>
            <AntDesign.Col Span="20">
                <FormItem Label="专家建议">
                    <TextArea @bind-Value="@context.ReportSuggestForStudent" MinRows="3" />
                </FormItem>
            </AntDesign.Col>
        </Row>
        <Row Style="margin-bottom:10px">
            <AntDesign.Col Span="24">
                <Text Strong="true">针对家长的综合报告</Text>
[...]

PR #1155, opened 2021-02-22
fix(module:TextArea): sizing bug for components outside viewport

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#598 
### 💡 Background and solution
The height of `TextArea` was calculated base on multiple artificial changes to values of hidden textarea and javascript requests to read the current sizes. The problem (I do not know why) was that when textarea was outside of the viewport, the height was calculated to 0 (zero). So instead of relying on these artificial changes, I created an extra function in `interop.ts` that allows us to request any calculated css style - `GetStyle`. For `TextArea` this function is used to obtain `line-height` that referst to row height and `padding-top` and `padding-bottom` to get the offset (this I am not sure, but it seems to me that offset in textarea is actually the sum of `padding-top` and `padding-bottom`). I think this change is also more performant, as previous was issueing `StateHasChanged()` 3 times, while this does not need to do any of the `StateHasChanged()` calls. 
I think this could be even made more performant if a dedicated function were to be created just for getting the `line-height`, `padding-top` & `padding-bottom` in a single js call (now these are 3 calls for each style). If this is acceptable, please let me know, I will implement it.
Also, I kept the old method of getting the size as a fall-back - in case my solution for some reason fails.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames:
It's fine 👍

anddrzejb:
Sorry, I do not follow:
* it is fine to create extra js method just for `Textbox`   
*OR*
* it is fine as it is and you do want any more changes?

ElderJames:
To create extra js in this PR first.

ElderJames:
Also need to fix height adjustment when pressing enter or when deleting lines.

react:
image

blazor:
image

anddrzejb:
There is a problem in here...
So from my analysis, it seems that the source of the problem is the fact that there is another hidden `<TextArea>` that is used for calculation of the text height. This hidden element is getting the same value as the original `TextArea`, but without all the formatting. My understanding is that its scrollHeight should give us the real height of the text. This is generally working, but blazor makes it kind of problematic. The hidden element's scrollHeight is checked using JSInterop. This works fine. The issue here is the order of events fired. The scrollHeight is fetched before the value from original `TextArea` component is applied to the hidden element. So the scrollHeight returns the *old* value. A sample of order of events:   

0. Let's assume `TextArea`'s starting value are 2 lines with text.
1. Press enter to create new line in `TextArea`.
2. Value is being retrieved from `TextArea` component (2 lines with text + new line)
3. scrollHeight is fetched from hidden clone of the `TextArea` (which is still seeing only 2 lines with text)
4. New value (2 lines with text + new line) is applied to hidden clone.

I tried different combinations of `StateHasChanged()` and `Task.Delay` to try to force the new value to the hidden clone before the height change, but nothing worked. All of this I tried in `ChangeSizeAsync()` method of `TextArea.razor.cs`. Any ideas here?

I also tried something else - I removed the hidden element and tried the same logic on the primary visbile `<textarea>` element. The trick here is to actually change the current height of the `<textarea>` element to `auto` (this is needed mostly to detected text shrinking, growing does not need it), then fetch its scrollHeight and then do the calculation of the real expected height and apply it.  This is working, but again, blazor is kind-of in the way. After the height style change to `auto`, it has to be pushed to the DOM. So, the browser redraws. When there is a change back to calculated height, there is a redraw again. This looks like flickering:
textArea_flickering

If this was JS, the flickering would not happen. In JS redraw would happen after all JS work is done. In this case blazor publishes the changes to DOM but the browser does not know blazor did not finish and redraws. Is there any way to stop redrawing?

At this stage, the best solution I can think of is the one I really hate - move the whole thing to JS 😢 . I am definitely open to suggestions...

anddrzejb:
@ElderJames Commit d485f4e contains the fixes. Unfortunately, I was not able to find a way to do this with blazor. For a nice experience I had to revert back to JS...I tried multiple approaches, but none of them were offering a nice experience...

ElderJames:
It still doesn't work on my machine...

anddrzejb:
I am confused...Just to confirm, I cleaned the solution, rebuilt and run this code:
```razor
<Button OnClick="@(() => add = !add)">Add/Remove</Button><br/>
@if (add)
{
    <TextArea @bind-Value="@txtSmaller" Autosize/>
    <br />
    <br />

}
No autosize<TextArea @bind-Value="@txtSmaller" />
<br />
<br />
Autosize<TextArea @bind-Value="@txtSmaller" Autosize/>
<br />
<br />
MinRows=3<TextArea @bind-Value="@txtSmaller" MinRows="3" />
<br />
<br />
MinRows=3 MaxRws=4 <TextArea @bind-Value="@txtSmaller" MinRows="3" MaxRows="4"/>
<br />
<br />
MinRows=3<TextArea @bind-Value="@txtEqual" MinRows="3" OnSelectionChange="OnSelectionChange" />
<br />
<br />
MinRows=3 <TextArea @bind-Value="@txtGreater" MinRows="3" />
<br />
<br />
MinRows=3 MaxRws=4 <TextArea @bind-Value="@txtGreater" MinRows="3" MaxRows="4"/>
<br />
<br />
MinRows=3 <TextArea @bind-Value="@txtEmpty" MinRows="3" />
<br />
<br />
MinRows=3 <TextArea @bind-Value="@txtNull" MinRows="3" />

@code {
    public bool add = true;
    public string txtSmaller = "line 1 \r\n line2";
    public string txtEqual = "line 1 \r\n line2 \r\n line3";
    public string txtGreater = "line 1 \r\n line2 \r\n line3 \r\n line4 \r\n line5" ;
    public string txtEmpty = "";
    public string txtNull;
}
```

and seems to be working as expected:
textArea_working

I tested this on:
* Firefox
* Chrome
* Vivaldi
* Edge

What am I missing?

ElderJames:
Great! It works on the preview. There is still one problem left here, the textarea will open up a little when you first type.

anddrzejb:
I do not know how I missed that. Maybe this commit 83ba002 will do the trick :smile:.

ElderJames (approved):

PR #1168, opened 2021-02-24
refactor(module: select): use Func to get or set value instead of reflection

Zonciu:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | use Func to get/set value instead of reflection  |
| 🇨🇳 Chinese | 使用Func代替反射读写数据 |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -38,10 +38,32 @@ namespace AntDesign
             }
         }
         [Parameter] public bool Disabled { get; set; }
-        [Parameter] public string DisabledName { get; set; }
+
+        [Parameter]
+        public string DisabledName
+        {
+            get => _disabledName;
+            set
+            {
+                _getDisabled = string.IsNullOrWhiteSpace(value) ? null : SelectItemPropertyHelper.CreateGetDisabledFunc<TItem>(value);
+                _disabledName = value;
+            }
+        }
+
         [Parameter] public Func<RenderFragment, RenderFragment> DropdownRender { get; set; }
         [Parameter] public bool EnableSearch { get; set; }
-        [Parameter] public string GroupName { get; set; } = string.Empty;
+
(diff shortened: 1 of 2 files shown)


anddrzejb:
@Zonciu I am not a reflection and expression tree expert, so correct me if I am wrong. My understanding after having a very brief look at your code, is that you are still using reflection, but instead of every time, you basically "compile" it on set and when it is used, it is just using compiled code instead of going through reflection again. This is where your performance boost is coming from. So we may loose a bit more time during first rendering, but gain time on every item processing when value/label is required to be extracted to set.

Zonciu:
@anddrzejb You are right, and we can make it go further if it is necessary: caching `typeof(TItem)` , `typeof(TItemValue)` , `LabelName` , `DisabledName` , `GroupName` as cache key to reuse those Expression delegates

anddrzejb:
Sounds really cool. I have seen reflection quite often used in other components (for example `DatePicker`, a big performance offender), it would be really nice if we could somehow make the helper you made universal. But this is maybe for some other time. I do not have time unfortunately to checkout your code today and I am leaving for couple of days, so either @ElderJames  can review it or I will once I came back after Sunday. Sorry... 😞 

Maybe you could propose your caching idea here as well in the meantime?

Zonciu:
I will create a Func cache commit for this.

Actually I have made a helper for property accessing, it works on `table` component, see [Path-based property access](https://ant-design-blazor.gitee.io/en-US/docs/path-based-property-access)

ElderJames (approved):

PR #1191, opened 2021-03-02
fix(module: Select) bind value change lock, tags and labels fix, overlay reposition

### LINKED ISSUE #1171: [Select] Can't change the binding value after changing the @bind-Value
组件Select：修改的bind-Value后其显示的选项并不变化，设置DefaultValue的值bind-Value也不发生变化

The select does not change the options displayed after the `@bind-Value` change. Setting the value of `DefaultValue`, `@bind-Value` also does not change.

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1171 

#### Commit 25c3dcb
During my tests, it came out that if I had a class that the label was attached to a property that it had only a getter, then `Select` was crashing. The source of problems was the PR #1168. The `_setLabel` was prepared regardless of the type of select. The property that is attached to label needs to have a setter only for `SelectMode.Tags`. 

#### Commit d4e2966
Several fixes to solve #1171. The result it this:

select_bindValueLockFix

#### 7c7d481
Update of the overlay position was not really effective, as it was doing its updates before all the relevant changes were pushed to DOM. Adding `Task.Delay(1)` solved this issue. I am not really happy with it. The other option would be maybe to use [ResizeObserver](developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) but there might be a problem with browser compatibility. 

#### Other condierations:
All operations done in mehtods `EvaluateValueChangedOutsideComponent` and `EvaluateValuesChangedOutsideComponent` are basically needed only when `Value` and `Values` are changed outside of the component. I imagine this is going to be a much less often scenario in comparison to when they are changed by the component. At this stage, when values are changed by the component, the aforementioned methods are not necessary. This is mostly a performance issue when dealing with multiple/tags mode. But it probably would be beneficial to have some kind of a way to generate a hash or some other way to be able to compare if values were changed internally or externally. If external change is detected, then run the evaluation methods.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -1249,10 +1249,15 @@ namespace AntDesign
 
             EvaluateValuesChangedOutsideComponent(values);
 
-            //if (_dropDown.IsOverlayShow())
-            //{
-            //    await UpdateOverlayPositionAsync();
-            //}
+            if (_dropDown.IsOverlayShow())
+            {
+                //A delay forces a refresh better than StateHasChanged().
+                //For example when a tag is added that is causing SelectContent to grow,
+                //this Task.Delay will actually allow to reposition the Overlay to match
+                //new size of SelectContent.
+                await Task.Delay(1);
+                await UpdateOverlayPositionAsync();
+            }
 
             OnSelectedItemsChanged?.Invoke(SelectedOptionItems.Select(s => s.Item));
             await ValuesChanged.InvokeAsync(Values);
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
LGTM.

PR #1208, opened 2021-03-05
feat(module:table): add OnExpand event

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
`OnExpan` event was missing in comparison to antd (and I needed it :smirk:).

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Added event handler `OnExpand`.       |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/table/Column.razor.cs
+++ b/components/table/Column.razor.cs
@@ -191,8 +191,11 @@ namespace AntDesign
 
         private void ToggleTreeNode()
         {
+            bool expandValueBeforeChange = RowData.Expanded;
             RowData.Expanded = !RowData.Expanded;
-            Table?.Refresh();
+            Table?.OnExpandChange(RowData.CacheKey);
+            if (RowData.Expanded != expandValueBeforeChange)
+                Table?.Refresh();
         }
 
         private void FilterSelected(TableFilter<TData> filter)
--- a/components/table/ITable.cs
+++ b/components/table/ITable.cs
@@ -28,6 +28,8 @@ namespace AntDesign
 
         internal void SelectionChanged();
 
+        internal void OnExpandChange(int cacheKey);
+
         internal void Refresh();
 
         internal void ReloadAndInvokeChange();
--- a/components/table/Table.razor.cs
+++ b/components/table/Table.razor.cs
@@ -104,6 +104,9 @@ namespace AntDesign
         public Func<RowData<TItem>, string> ExpandedRowClassName { get; set; } = _ => "";
 
         [Parameter]
+        public EventCallback<RowData<TItem>> OnExpand { get; set; }
+
+        [Parameter]
         public SortDirection[] SortDirections { get; set; } = SortDirection.Preset.Default;
 
         [Parameter]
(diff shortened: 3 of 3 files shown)


ElderJames:
Sorry, please rebase again. Since I saw a problem with the PREVIEW action, I repushed a COMMIT to change it.

components/table/Column.razor.cs
```diff
@@ -191,8 +191,11 @@ private void SetSorter(SortDirection sortDirection)
 
         private void ToggleTreeNode()
         {
+            bool expandValueBeforeChange = RowData.Expanded;
             RowData.Expanded = !RowData.Expanded;
-            Table?.Refresh();
+            Table?.OnExpand(RowData.CacheKey);
```
ElderJames, line 196: Should it be invoked when it was expanded or not?

anddrzejb: The name suggests that it should be invoked on expand only. However I do not see the reason why only then. Since the consumer will get the "state" of expanded in `RowData.Expanded` (the event is passing `RowData`), then the consumer may decide which event variant he/she is interested in. And actually, in my case, I am interested in both - when the expanded is showing and when it is hiding.
Maybe we should change the event name to "OnExpandClicked" or alike?

ElderJames: How about `OnExpandChange` ?

ElderJames:
BTW, could this PR target to `feature` branch?

components/table/Column.razor.cs
anddrzejb: Sure, no problem 👍🏻 . When I was adding it, I based it on antD, but you are right, this is better. There is always a dilemma - follow the antD or deviate. Regardless, I am glad you are ok with that solution.

anddrzejb:
Yeah, sure, but to be honest I was hoping to get it in master, since I think it is basically ready. Once you merge it with feature, will you merge it to master anytime soon?

ElderJames:
I'm not clear..Is it ok to rebase feature base on master?

anddrzejb:
I do not think so....Right now (when I write this comment) `feature` branch is 1 commit ahead and 7 behind `master`. I am not 100% sure what is going to happen to `master` if you'd rebase now, but I do not think anything good. I believe one should treat `master` as the only source of truth. If that is the case, you need to first merge from `master` to `feature`  (or rebase from `master`) and then probably merge from `feature` to `master`? Mind that I am no git guru...

anddrzejb:
I also found [this gist](https://gist.github.com/vlandham/3b2b79c40bc7353ae95a) on the subject.

ElderJames (approved):

components/table/Column.razor.cs
```diff
@@ -191,8 +191,11 @@ private void SetSorter(SortDirection sortDirection)
 
         private void ToggleTreeNode()
         {
+            bool expandValueBeforeChange = RowData.Expanded;
             RowData.Expanded = !RowData.Expanded;
-            Table?.Refresh();
+            Table?.OnExpandChange(RowData.CacheKey);
+            if (RowData.Expanded != expandValueBeforeChange)
+                Table?.Refresh();
```
ElderJames, line 198: RowData.Expanded is always not equal to expandValueBeforeChange.

anddrzejb: That is the idea. My use case is that I am checking some kind of conditions and do not allow expansion if they are not met. So if the expand was clicked but consumer's code changed it, there is no need to refresh anything.

anddrzejb: To some degree, if you wanted, you could probably simulate accordion functionality with this.

ElderJames: Ok, get it.

anddrzejb:
@ElderJames Sorry to bother you about it again, but do you think this feature is going to be merged to master anytime soon? Or at least for ver 0.8?

PR #1214, opened 2021-03-06
fix(module:select): Compiled method .ToString() was throwing AmbigiousMethod

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
In `Select` module when setting `LabelName` to a property that has overload of `.ToString()`, the `SelectItemPropertyHelper` was throwing `AmbigiousMethod` exception. For example if a `DateTime` type was passed. 
My fix is making sure a parameter-less `.ToString()` is being used.


Zonciu (approved):
LGTM

ElderJames (approved):

Issue #1284, opened 2021-03-30
Table ExpandableRow not working with ActionColumn


Magehernan:
### Describe the bug

When you have a table with ActionColumn first and you want to have the expandable button first, is impossible to configure after the commig feat(module: table): specific the position of the expand button (#1002).

### Steps to reproduce (please include code)

From the table ExpandRow example

```csharp
@using System.ComponentModel

<Table DataSource="data" RowExpandable="@((record)=>record.Data.Name!="Not Expandable")">
    <RowTemplate>
        <ActionColumn Title="Action">
            <a>Delete</a>
        </ActionColumn>
        <Column @bind-Field="@context.Name" />
        <Column @bind-Field="@context.Age" />
        <Column @bind-Field="@context.Address" />
    </RowTemplate>
    <ExpandTemplate>
        @context.Data.Description
    </ExpandTemplate>
</Table>

@code {

    Data[] data =
    {
        new() {
            Key = "1",
            Name = "John Brown",
            Age = 32,
            Address = "New York No. 1 Lake Park",
            Description = "My name is John Brown, I am 32 years old, living in New York No. 1 Lake Park."
        },
        new() {
            Key = "2",
            Name = "Jim Green",
            Age = 42,
            Address = "London No. 1 Lake Park",
            Description = "My name is Jim Green, I am 42 years old, living in London No. 1 Lake Park."
        },
        new() {
            Key = "3",
            Name = "Not Expandable",
            Age = 29,
            Address = "Jiangsu No. 1 Lake Park",
            Description = "This not expandable"
        },
        new() {
            Key = "4",
            Name = "Joe Black",
            Age = 32,
            Address = "Sidney No. 1 Lake Park",
            Description = "My name is Joe Black, I am 32 years old, living in Sidney No. 1 Lake Park."
        }
    };

    public class Data
    {
        [DisplayName("Key")]
        public string Key { get; set; }

        [DisplayName("Name")]
        public string Name { get; set; }

        [DisplayName("Age")]
        public int Age { get; set; }

        [DisplayName("Address")]
        public string Address { get; set; }

        [DisplayName("Description")]
        public string Description { get; set; }
    }
}
```

move the ActionColumn first like in this code and the button for expand won't show.
If you specify a value for ExpandIconColumnIndex that is the index for a column it works, but if not, it never will shown.

### Further technical details
- AntDesign Nuget Package version 0.7.3

Magehernan:
[...]

PR #1285, opened 2021-03-30
(fix) Table ExpandableRow not working with ActionColumn #1284

Magehernan:
### 🤔 This is a fix for issue #1284 

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/table/ActionColumn.razor
+++ b/components/table/ActionColumn.razor
@@ -55,6 +55,18 @@ else if (!IsHeader && RowSpan != 0 && ColSpan != 0)
     }
 
     <td class="@ClassMapper.Class" style="@FixedStyle @Style" rowspan="@RowSpan" colspan="@ColSpan">
+        @if (ColIndex == Table.TreeExpandIconColumnIndex && Table.TreeMode)
+        {
+            <span class="ant-table-row-indent indent-level-@RowData.Level" style="padding-left: @((CssSizeLength)(RowData.Level * Table.IndentSize));"></span>
+            @if (RowData.HasChildren)
+            {
+                <button type="button" @onclick="ToggleTreeNode" class="ant-table-row-expand-icon @(RowData?.Expanded==true?"ant-table-row-expand-icon-expanded":"ant-table-row-expand-icon-collapsed")" aria-label="@(RowData?.Expanded==true?Table.Locale.Collapse:Table.Locale.Expand)"></button>
+            }
+            else
+            {
+                <button type="button" class="ant-table-row-expand-icon ant-table-row-expand-icon-spaced" aria-label="@Table.Locale.Expand"></button>
+            }
+        }
         @ChildContent
     </td>
 }
--- a/components/table/ActionColumn.razor
+++ b/components/table/ActionColumn.razor
@@ -15,6 +15,11 @@ else if (IsMeasure)
 }
 else if (IsColGroup)
 {
+    @if (AppendExpandColumn)
+    {
+        <col class="ant-table-expand-icon-col">
+    }
+
     if (Width != null)
     {
         <col style="width: @((CssSizeLength)Width); min-width: @((CssSizeLength)Width);">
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=1285) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=1285) before we can accept your contribution.<br/><br/>:x: Magehernan [sign now](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=1285)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/ant-design-blazor/ant-design-blazor?pullRequest=1285) it.</sub>

ElderJames (commented):
Thanks for contribution! 
Please add the expand button for the tree-data mode as well.

Magehernan:
@ElderJames Hi, i add this and found another bug when using selection column but not in the fist position, so if the selection column is present it add 1 to expandcolumn so if you put in the second column the selection it will be the same as the selection and the selection column don't have code to show the expand button.... so i and looking for the position of the selection to modify the expandcolumn to show this correctly.

I will send this fix to.

Magehernan:
@ElderJames so i found that if you use a Selection Column in Tree or Expand mode... it was breaking.

I fix it adding 1 if the selection column is less or equal index that the column you want to have the expand button, but this work fine only when you have none or one Selection Column, if you have two selection Column, the value of _selection.ColIndex is the last and it will work weard.

I don't want to make a lot of changes on the code. I make it better but the problems is more deep than this.

i don't know what to do?  
do i create a issue about the more than one selection column? 
I think maybe the better solution to this is make on the Column a parameter that indicate if show or not the expand button.... or in the table put a string for the name of de column you want to put the expand button.
Maybe there are betters solution.

ElderJames:
I think we need to limit the user to setup the selection in the first column.

Magehernan:
I look to the table for react ant.design and i try to move the selection column and i think it is not posible...

So i think you are right.

Magehernan:
@ElderJames How do you link this force to one selection on first column?
I think maybe add parameters to the table to configure what type of selection the user want and remove the Column Tag <Selection /> 
Or if the user use the Selection Tag, not use the position of it and force it to be first?

Do you want i make this changes?

ElderJames:
Yes, sure, we would happy if you can help! 😀 I prefer the second option.

You can create another PR base on `feture` branch to do that.

ElderJames (approved):

Magehernan:
@ElderJames Hi again... 
I am strugling with what is best for this.
Looking at the code, i am getting more aware of the amount of complexity and problems there are.

The table only have a ISelectionColumn but if you use all you want, it will work, not perfect but will work. So this is the first problem, how to show error on the blazor page telling the dev that it is not allowed to put more than one Selection Column?

If we remove the posibility for adding more than one Selection Column then the problem that we have now will disappear.

So I think the best soluction isn't to force to the first position of the selection column. If it throw an exception when the selection column is not null and it try to assign it. Then we limit the use of it to one... so only if someone put 2, will have this exception.

https://github.com/ant-design-blazor/ant-design-blazor/blob/5263a7119ff6bd5ff4f7043a2f383f93adfa2068/components/table/Selection.razor.cs#L33-L50

If i add this code when you see the table it will throw the exepction... I don't really like this way but i am not an expert on blazor so if it can be done with an error when try to build your code then it will be a lot better.

```csharp
        {
            base.OnInitialized();

            if (Table == null)
            {
                return;
            }

            if (IsHeader)
            {
                if (Table.Selection is not null)
                {
                    throw new System.InvalidOperationException("Only one Selection column is allowed");
                }
                Table.Selection = this;
            }
            else if (IsBody)
            {
                Table?.Selection?.RowSelections.Add(this);
            }
        }
```

I think is necesary a lot of work to make it cleaner... i don't have a lot of time to do it... but i leave this observation...
The feature for Expand is always a new column that the other columns expect Selection are adding when the index are equals... 
The feature for Tree is a button inside a column, same as Expand the column check and add it or not.

I think i can be better to split the login for this two features, the expand button can be a column type.
The tree can be a parameter for the columns when you choose if it show or not.

Then you can have multiples expand and multiples tree(if someone really need something like this) and becase it need to be explisit in the code, the dev will have what them put in code.

PR #1304, opened 2021-04-02
fix(module:table): perf optimization & data source change issue

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
When data source is changing (especially when it is less then previous count), then table is trying to iterate through items that no longer exist. To recreate, run below code and click several times on the button "Generate new set of data". After few times you will see there was exception causing application crash:
```razor
@using System.ComponentModel
<Button OnClick="GenerateNew">Generate new set of data</Button>
<Table DataSource="@filteredData" RowExpandable="@((record)=>record.Data.Name!="Not Expandable")">
    <RowTemplate>
        <Column Field="@context.Name" />
        <Column Field="@context.Age" />
        <Column @bind-Field="@context.Address" />
        <ActionColumn Title="Action">
            <a>Delete</a>
        </ActionColumn>
    </RowTemplate>
    <ExpandTemplate>
        @context.Data.Description 
        <br/>
        @context.Data.Key is even: @_activationLog[context.Data.Key];
    </ExpandTemplate>
</Table>
<br/>
@string.Join(", ", filteredData.Select(x => x.Key).ToArray())
@code {
    Data[] filteredData;
    Dictionary<int, bool> _activationLog;

    private Data[] GenerateData()
    {
        return new Data[]
             {
                new() {
                    Key = 1,
                    Name = "John Brown",
                    Age = 32,
                    Address = "New York No. 1 Lake Park",
                    Description = "My name is John Brown, I am 32 years old, living in New York No. 1 Lake Park."
                },
                new() {
                    Key = 2,
                    Name = "Jim Green",
                    Age = 42,
                    Address = "London No. 1 Lake Park",
                    Description = "My name is Jim Green, I am 42 years old, living in London No. 1 Lake Park."
                },
                new() {
                    Key = 3, 
                    Name = "Not Expandable",
                    Age = 29,
                    Address = "Jiangsu No. 1 Lake Park",
                    Description = "This not expandable"
                },
                new() {
                    Key = 4,
                    Name = "Joe Black",
                    Age = 32,
                    Address = "Sidney No. 1 Lake Park",
                    Description = "My name is Joe Black, I am 32 years old, living in Sidney No. 1 Lake Park."
                }
            };
    }

    public class Data
    {
        [DisplayName("Key")]
        public int Key { get; set; }

        [DisplayName("Name")]
        public string Name { get; set; }

        [DisplayName("Age")]
        public int Age { get; set; }

        [DisplayName("Address")]
        public string Address { get; set; }

        [DisplayName("Description")]
        public string Description { get; set; }
        
        //if not existent, then will multiply the data in _dataSourceCache
        public override int GetHashCode()
        {
            return Key.GetHashCode();
        }
    }

    {
        filteredData = GenerateData().Take(3).ToArray();
        _activationLog = filteredData.ToDictionary(x => x.Key, x => (x.Key % 2) == 0);
    }
    
    void GenerateNew()
    {        
        var rand = new Random();
        int take = rand.Next(1, 4);
        filteredData = GenerateData().Take(take).ToArray();
        _activationLog = filteredData.ToDictionary(x => x.Key, x => true);        
    }
}
```
table_exceptionOnRender

### 💡 Background and solution
Commit 43e8bc4 addresses the bug. 
Commit d49eddc fixes double rendering performance issue.
To observe the performance improvement, you can put this line in `Table.razor` in line 150:
```csharp
   Console.WriteLine("Generating body...");
```
Then run the code above and observe the console. Without the commit, when clicking on the button "Generate new set of data", you will notice the message 2 times. With that commit, it is a single message.


components/table/Table.razor.cs
```diff
@@ -349,7 +349,15 @@ protected override void OnParametersSet()
...
         }
 
-        protected override bool ShouldRender() => this._shouldRender;
+        protected override bool ShouldRender()
+        {
+            if (_waitingReload)
+            {
+                _waitingReload = false;
+                Reload();
+            }
```
ElderJames, line 358: Why not write it in the AfterRender method?

Zonciu: It should not reload data in `OnAfterRender` method, the reload action should be happened before render.
@anddrzejb you can move the reload action code from `OnAfterRender` method(`Line 295 - L306`) to `ShouldRender` method. https://github.com/ant-design-blazor/ant-design-blazor/blob/3ffafdf20033070d1793435cd02b6f1ca09a3ed1/components/table/Table.razor.cs#L295-L306

anddrzejb: @Zonciu As requested, commit 7a3b1e9. It definitely makes more sense to transfer all reloads before render is done. I still fill a bit intimidated by `Table` code...

anddrzejb: @ElderJames The idea here is to optimize by minimize number of renderings - running this code `OnAfterRender` means all is already rendered, so I would say that perf optimization of number of renderings in that event is a bit late.

ElderJames: I see. You have made a great discovery.

ElderJames (approved):

components/table/Table.razor
```diff
@@ -146,6 +146,7 @@ RenderFragment<(Table<TItem> table, IEnumerable<TItem> showItems, int level)> bo
         var table = context.table;
         var showItems = context.showItems;
         var level = context.level;
+        var count = Math.Min(showItems.Count(), table._dataSourceCount);
```
Zonciu, line 149: Well, when you move reload code to ShouldRender method, here is no need to get the minimum count anymore, they are the same.

anddrzejb: Yep...You are right. We need a new PR for that...

PR #1305, opened 2021-04-02
fix(module:textarea): default to empty string

### LINKED ISSUE #1242: TextArea sets value to null when field is empty.
### Describe the bug
I have a text area with a binding to a field of type string. When the text area is empty it doesn't set the value to string.empty but to null.

This is the TextArea:

       <TextArea MinRows="3" @bind-Value="@context.Description" Type="string" />

And this is the model in abbreviated form:

```
public record CustomerProjectOverviewDto
    {
...
        public string Description { get; set; } = "";

...
    }
```

We close this issue when the correct default value is set.

### Further technical details

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
fixes #1242

### 💡 Background and solution
Whenever a value changes, the `OnInputAsync` is fired. It picks up the value from the `ChangeEventArgs` and tries to parse it with method in `AntInputComponentBase.TryParseValueFromString`. There it checks if string `IsNullOrWhitespace` and when true, resets value to the `default`. Which for `string` is null. So, I added to `TextArea` new boolean parameter `DefaultToEmptyString` and overriden `TryParseValueFromString` in `TextArea` component, where depending on the parameter value appropriate default value is assumed. 
I was also trying out the nullable types, but even then string was still defaulting to null. 

Firefox seems to be much more sensitive to js and fails quite often when working with `TextArea`. I added some additional checks to `interop.ts` to protect against these issues.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  New parameter `DefaultToEmptyString`         |
| 🇨🇳 Chinese |           |


ElderJames (approved):
LGTM

PR #1307, opened 2021-04-03
fix(module:select): load icon for multiple mode & value reload on delay

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
1. `Loading` had no effect when `Mode="multiple"`
2. When `DataSource` & `Values` are changed, this is not reflected. To reproduce, use the below code. It will simulate delayed data load and value setting after the load. Use case - user list, when selected user should meet certain conditions which will be known only after loading the list (for example from API).
```razor
<Select Mode="multiple"
        DataSource="@_persons"         
        @bind-Values="@_selectedPersons"
        LabelName="@nameof(Person.Name)" 
        ValueName="@nameof(Person.Id)" 
        TItem="Person"
        TItemValue="int"
        DisabledName="@nameof(Person.NotAvailable)"
        Placeholder="Select a person"
        AllowClear>
</Select>
<br/>
Selected persons ids: @(_selectedPersons is not null ? string.Join(",", _selectedPersons.ToArray()): "empty array")
@code
{
    class Person
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public bool NotAvailable { get; set; }
    }

	List<Person> _persons = new();
    IEnumerable<int> _selectedPersons;

	protected override async Task OnInitializedAsync()
    {
		await Task.Delay(3000);
        _persons = new List<Person> 
        {
            new Person {Id = 1, Name = "Jack"}, 
            new Person {Id = 2, Name = "Lucy"}, 
            new Person {Id = 3, Name = "Yaoming"}, 
            new Person {Id = 4, Name = "Frieda"}, 
            new Person {Id = 5, Name = "Kate"}, 
            new Person {Id = 6, Name = "Eric"}, 
            new Person {Id = 7, Name = "Olaf"}, 
        };
		_selectedPersons = new List<int> { 1 };		
    }
}
```
Notice the selecte value appears after list is loaded. But the value is not visible in the `SelectContent` until another value is selected.
select_delayedValue

### 💡 Background and solution
1. Added missing loading icon.
2. Rebuilding of select item list was done only partially when values and datasource is changing. 
### ☑️ Self Check before Merge

[...]

PR #1308, opened 2021-04-03
fix(module:select): propagation on item remove

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Issue mentioned in [this PR1307](https://github.com/ant-design-blazor/ant-design-blazor/pull/1307#issuecomment-812895579)

### 💡 Background and solution
Removeable item is basically a span with icon, that has the `@onclick` event. That event has also `@onclick:stopPropagation="true"`. What I did not know, that this does not apply to DOM events, but only to blazor events ([source](https://github.com/dotnet/aspnetcore/issues/30199)). I could probably try to make some complex js, but the simples change was to keep this as local as possible. But, if a dedicated function in `interop.ts` is required, it can probably be done.

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -1205,7 +1205,8 @@ namespace AntDesign
         /// </summary>
         internal async Task UpdateOverlayPositionAsync()
         {
-            await _dropDown.GetOverlayComponent().UpdatePosition();
+            if (_dropDown.Visible)
+                await _dropDown.GetOverlayComponent().UpdatePosition();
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
So far it seems to have solved the problem

PR #1338, opened 2021-04-12
feat(module:select): Select responsive max tag count

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Adds missing `MaxCountTag`, `MaxTagPlaceholder` and `MaxTagTextLenght`

### 💡 Background and solution
The responsive solution is based on js `window.resize` event. Originally I planned to use resize observer API, but because that change will have to span on more than a single functionality, I plan to make a separate PR that will apply a switch to that API (if successful). 

Current drawback is that `window.resize` is not really very responsive and takes some time to react.
Also, due to issues with `focus`, I changed current implementation to mirror what dotnet team had done with thier `ElementRefernece.FocusAsync()` built-in method. In fututre PR I will propose a global change on the focus method depending on the framework version used.

I am posting this to `master`, because this is bringing the `Select` component in line with antD. Please let me know if this should be in the `Feature` branch.

Also - please check API Chinese translation. I copied what I could from antD, but there are some comments from me in English.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   Added missing `MaxCountTag`, `MaxTagPlaceholder` and `MaxTagTextLenght`        |
| 🇨🇳 Chinese |           |


site/AntDesign.Docs/Demos/Components/Select/demo/maxTagResponsive.md
```diff
@@ -0,0 +1,57 @@
...
+      }}
+    >
+      <Select {...selectProps} />
+      <Select {...selectProps} disabled />
+    </Space>
+  );
+};
+
+ReactDOM.render(<Demo />, mountNode);
+```
```
ElderJames, line 57: Please delete the jsx

ElderJames:
This is very similar to the tabs navbar, but it hasn't been implemented correctly yet. 
Could you please help and see if you can repair it?

site/AntDesign.Docs/Demos/Components/Select/demo/maxTagResponsive.md
anddrzejb: You were faster 😉

anddrzejb:
I presume you are talking about the `Menu` component. I will have a look as soon as I finish with other tasks that I have planned.

ElderJames:
No, I'm talking about the tabs component.

ElderJames (approved):

anddrzejb:
Can you point me exactly to which part? The only similar functionality that I can see is in {slide section](https://antblazor.com/en-US/components/tabs#components-tabs-demo-slide) and it seems to me it behaves as it should (certainly behaves like the antD)...

ElderJames:
Not be completed. Currently our implement can't scroll with the mouse wheel event, and can't handle the overflow items.

Please notice that the drop-down part is actually the part that is not shown on the tabs, not just the part that is overflowing from the tail.
ant-tabs-slide
.

anddrzejb:
Ok, I think I get it now. Thanks for the explanation.

Issue #876, opened 2020-12-09
Problem with the Style attribute in the Input component


szmalec:
### Describe the bug
Component Input with the `Style` attribute may render differently depending on additional attributes, eg `AllowClear`.

### Steps to reproduce (please include code)
```
<Input Placeholder="Enter text" @bind-Value="_txt" Style="text-align: center" />

<Input Placeholder="Enter text" @bind-Value="_txt" Style="text-align: center" AllowClear />

@code {
    private string _txt = "Lorem ipsum...";
}
```

![image](https://user-images.githubusercontent.com/7856921/101612827-5fb9de80-3a0b-11eb-8328-dacfd0dc77e8.png)

As you can see, adding the `AllowClear` attribute resulted in wrapping the `input` tag in an additional `span` tag and the style passed in the `Style` parameter is applied to this `span` tag instead of the `input` tag. Shouldn't it work consistently in both cases?

### Further technical details
Problem in: 0.5.0-nightly-2011180810 and also in 0.5.0-nightly-2012081156

anddrzejb:
@ElderJames I was trying to tackle this issue but I think we have a general design problem here. As @szmalec pointed out, there is an extra `<span>` wrapping the `<intput>`, which is fine - react antdesign does exactly the same. The `<input>` is wrapped in extra `<span>` whenever there is any extra functionality added to the `<input>` (prefix/suffix/password/search/inputgroup/etc). The problem here is that for a consumer, the component is a single element, but reality is that it can be couple of different elements. So some of the styles that a consumer would like to apply to the element, should end up either in `<span>` or in `<input>`. 
For example if a consumer of an `<Intput AllowClear>` component would like to pass `Style="text-align: center; width: 30%"`, then the `text-align` style should be applied to `<input>`, while `width` should be applied to `<span>`. How do we detect which belongs where? There are 3 options I can currently think of:
1. Hardcode which styles should apply to which element - I do not like this solution. It is a lot of work and checking a string for all possible occurrences is not the most efficient way of doing things.
2. Split the `Style` attribute into `StyleInternal` and `StyleExternal` (the names are not really a proposal, but I needed to use something just to present the idea). So `StyleInternal` would be applied to `<input>` and `StyleExternal` to `<span>`. I do not like this idea either, as it is very unintuitive and for regular `Input` without any extra functionality (the one that does not generate `<span>` wrapper) the `StyleExternal` would be useless (and would probably be a source of a lot of issues in this repo). 
[...]

PR #1342, opened 2021-04-13
fix: multiple bugs originating from js

### LINKED ISSUE #609: Error occurs when removing the last uploaded item of the Upload component
### Describe the bug
Upload組件，移除最後一個上傳的項目會返回錯誤
依然可以將項目移除，但是會返回錯誤

Error occurs when removing the last uploaded item of the Upload component.
The item can be removed but the below exception occurs:

### Exceptions
Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
      Unhandled exception rendering component: Unknown edit type: 0
      Error: Unknown edit type: 0

### Further technical details
- AntDesign Nuget Package version 0.3.0
- The IDE VS 2019

### LINKED ISSUE #1312: 'Input' component with the 'AutoFocus' property break the layout
### Describe the bug

When the property 'AutoFocus' is enabled in the 'Input' component then the layout is breaked.

'Input' whithout 'AutoFocus':

'Input' whithout 'AutoFocus'

'Input' whith 'AutoFocus':

Ant-2

In the repository [Playground.Ant](https://github.com/henriquemotaesteves/playground/tree/csharp/ant%2F0.7.3) change the line 8 of the file [Authentication.razor](https://github.com/henriquemotaesteves/playground/blob/c24380aa98ea86866e86cc33b375838f731076f0/csharp/Playground.Ant/Playground.Ant/Pages/Authentication.razor#L8) by adding ou removing the 'AutoFocus" property.

### Further technical details

- Browser

Ant-4

0.7.3

Ant-5

Ant-3

### LINKED ISSUE #1343: Input with clear - clicking on "Clear" makes the control to lose the focus
Hello guys.
Thanks for the awesome Blazor components' library.
Would it be possible to take a look at the issue with the focus of Input after pressing the clear button?

### Describe the bug
When an <Input TValue="String" /> control has a property `AllowCleat="true"` then pressing the clear button makes the input control to lose the focus. Therefore, in order to continue the input the control must be clicked one more time.
It would be nice if clicking on the close button preserved the focus in the input component.

https://antblazor.com/en-US/components/input#components-input-demo-Clear
1) Input some text into the input control with the clear button
2) press the clear button
3) try typing something

the control has lost the focus and typing goes nowhere.
```xml
<Input
    TValue="String"
    AllowClear="true" />
```

### Further technical details
- AntDesign Nuget Package version: 0.7.4
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version: Visual studio

```bash
.NET SDK (reflecting any global.json):
 Version:   6.0.100-preview.1.21103.13
 Commit:    b8a03527b2

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19042
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.100-preview.1.21103.13\

Host (useful for support):
  Version: 6.0.0-preview.1.21102.12
  Commit:  9b2776d481

.NET SDKs installed:
  3.1.403 [C:\Program Files\dotnet\sdk]
  5.0.200-preview.21077.7 [C:\Program Files\dotnet\sdk]
  5.0.201 [C:\Program Files\dotnet\sdk]
  6.0.100-preview.1.21103.13 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.26 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.26 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.0-preview.1.21103.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.26 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.0-preview.1.21102.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.7 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.1.13 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.2 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.4 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.0-preview.1.21103.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #609 
Fixes #1312 
Fixes #1343

### 💡 Background and solution
This PR fixes several issues. Firstly it was going to only address [this conversation](https://github.com/ant-design-blazor/ant-design-blazor/issues/1312#issuecomment-816742003). Once I started changing the focus method, I kept bumping into various issues. So, in order of appearance:
1. I extracted all focus calls to extension method. This way we should be able to easily make changes in the single place. Also, even though current focus method in `interop.ts` is identical to the one in blazor, I kept the split into `netstandart2.1`/`net5` so we get on the new version of focus method if blazor team decides to make changes to it. Components that are now using new `focus` method:
  -  `Input`
  -  `InputNumber`
  -  `Radio`
  -  `RangePicker`
  -  `Mentions`
  -  `Select`
  -  `Slider`
2. I bumped into the [focus & onfocus issue](https://github.com/dotnet/aspnetcore/issues/30070) - to summerize this issue - when built-in `@onfocus` event is used with `focus` method (regardles if built-in or the one from `interop.ts`), an exception is raised. This problem exists only in Wasm. I solved it by dropping built-in `@onfocus` & `@onblur` events in favour of our own js event handling. Seems to solve the issues. These changes were applied to `Input` & `Select` components. 
3. There was a problem with `Input` component - if `AutoFocus` was set to `true`, it was not autofocusing. 
4. During built-in events fixes, I again found myself facing same issue like in #609. Finally figured it out how to properly defer a js call - thus the change in event calling. My tests did not find any issue with that deferring. 
5. `SelectContent` was registering resize event for all `SelectMode.Multiple` modes. I added extra check to not run responsive code if the select content is not responsive. Previously it was throwing a bunch of exceptions. 


anddrzejb:
Commit c3febe8 except removing unnecessary arg from focus method, also fixes #1343 - this was a trivial fix, so I included it in here.

ElderJames (approved):

ElderJames:
Select would throw exceptions while I visit the select docs.

ant-select-exception

ElderJames (changes requested):
please fix the select exceptions

anddrzejb:
I see this is coming from `MaxTagCount` responsive functionality (the method you got the exception should be used only when responsive `MaxTagCount` is enabled). I used to have a fix for that, but it seems the removal of the fix was not a good idea (there is a race condition between js and blazor events - probably on my machine correct order was happening). I will try to recreate it and fix it in a few minutes.

anddrzejb:
@ElderJames I tried on my machine multiple times. Using the docs and my own project for developing. Just to confirm, I was also trying in server-side. I cannot recreate... So I have to ask you to do some debugging for me. Once I stopped your gif I noticed that actually the exception is in a new place and this is not the race condition I was thinking it is. Can you put a `try-catch` block and check which value is actually null? It is either `_aggregateTagElement` or `_overflowElement.width`. My money is on `_aggregateTagElement`. If that's the case, then in line 106 there is an `if` statement, that I would like you to remove, but keep the `Task.Delay`. Then test it.

anddrzejb:
@ElderJames  So now I am 99% sure that browser just has a problem catching up. Just to show you my methodology - maybe you will have some remarks on how to improve either the debugging experience or a better solution:
1. I added `id="_aggregateElement"` attribute to the div that gets `@ref=_aggregateTag`
2. I changed the code (temporarily of course) in `interop.ts`
```typescript
export function getBoundingClientRect(element, log: boolean, attempt: number = 0) {
  let dom = getDom(element);
  if (log) {
    console.log("getBoundingClientRect", attempt, element, dom, document.getElementById("_aggregateElement"));
  }
  if (dom && dom.getBoundingClientRect) {
    return dom.getBoundingClientRect();
  }
  return null;
}
```
2. I modified `do-while` loop to this:
```csharp
int z = 1;
do
{
    await Task.Delay(1);
    _aggregateTagElement = await Js.InvokeAsync<DomRect>(JSInteropConstants.GetBoundingClientRect, _aggregateTag, true, z);
    if (_aggregateTagElement is null)
        Console.WriteLine($"_aggregateTagElement is null, ref id {_aggregateTag.Id}, repeating request. {z++}");
} while (_aggregateTagElement is null);
```

So when there was a failure in getting the dimensions, I was notified in the console window:
image

And this I was seeing in the browser console:
image

So, even though we see a valid id passed from blazor, the js does not see the element. To confirm that element is not there, I also requested to check by id (thus `document.getElementById("_aggregateElement")`). Since all of them are null, my conclusion is that the element does not exist yet in the browser (still rendering?). Notice that in the 2nd attempt, when the element was found, the blazor id matches with element attribute. So it is the same element, just a fraction of a second later.

ElderJames:
Although this is not a good choice, but it's not unusual to have this kind of problem with the interop.
Using loop to interoperate here maybe would cause some performance issues, which I recommend put it in the js method.

anddrzejb:
I will do that, but I worry that possibly there might be a scenario when it is acceptable to return by js a null (not to find a certain element). If we put a loop in js, then in that scenario, we will create infinite loop. 
What we can possibly do, is to add an optional parameter to js function `repeatUntilFound` or something like that. Then we put on ourselves the responsibility for possible infinite loops - and it would have to be explicit. What do you think?

anddrzejb:
To be honest, I am now confused...I was trying to implement repeat in js, but now, `ElementReference.Id` is not set in scenarios when that exception is rising. If that is the case, then even js won't help us here, because we have no way of passing identifying attribute to js...
Look at this:
image
Even though I am logging the id, it comes up as empty:
image
And obviously, null is passed to js, which is useless...
image
Again, if I use the do-while in c#, it actually works, and waits. I have no clue why previously even though we got the id, js still could not find it. I am confused to what is happening here.

anddrzejb:
@ElderJames Commit e67c7b5 keeps while, but this time I do not request for JsInterop, but I am waiting for `ElementReference.Id`. During my tests, this wait was done up to max 40 times (under heavy load), but absolute major retries finished on a single await or none at all. Mind that these measurements where done in the debugging session, which is much slower. The other change I imagine we could do here is to increase the wait from 1ms to 5ms or 10ms (or whatever you think is suitable).

ElderJames:
Good! I pick 5ms.

anddrzejb:
@ElderJames I made a wrong commit so had to do some push forcing, but it should be fine now. I changed the wait to 5ms. I probably will ask around about this situation of `ElemenetReference` still being null in `OnAfterRenderAsyc`. I wonder if [this part](https://github.com/ant-design-blazor/ant-design-blazor/blob/2768b52c5480b31b6d33f3cd0962095d20ce030f/components/core/Component/overlay/Overlay.razor.cs#L164) of `Overlay.razor.cs` has the same underlaying issue...

ElderJames:
Oh, it also should be fixed..

anddrzejb:
Yeah, I know. I saw it for the first time when I was dealing with `Slider`'s tooltip. It look really hacky to me. But so does this. Maybe there is no other choice, but before I arrive to that final conclusion, I'd like to do a bit of research. Can we schedule a possible fix for  that later, not as a part of this PR?

ElderJames:
Yes, we can fix it later. I am preparing to release 0.8, and waiting for this PR to be merged.

anddrzejb:
Oh...Once you merge, I also have another PR waiting, that will fix #876 according to your comment on that issue. I can create it once this one is merged, so it could be solved probably today? Unless you find issues 😉

ElderJames (approved):

PR #1356, opened 2021-04-15
fix(module:table): locale fix

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Crashing exceptions whenever locale set to something else than zn-CH or en-US.

### 💡 Background and solution
Applied same solutions as in PR #1054 - added defaults (taken from en-US).
Included translations in en-GB & pl-PL for the missing fields. 

### DIFF
--- a/components/locales/en-GB.json
+++ b/components/locales/en-GB.json
@@ -99,7 +99,31 @@
     "filterConfirm": "OK",
     "filterReset": "Reset",
     "selectAll": "Select current page",
-    "selectInvert": "Invert current page"
+    "selectInvert": "Invert current page",
+    "selectionAll": "Select all data",
+    "sortTitle": "Sort",
+    "expand": "Expand row",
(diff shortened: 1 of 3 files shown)


ElderJames (approved):
I will fix in next release.

anddrzejb:
Mind that if you won't add it in current release, the table will not render at all (for other regional settings).

ElderJames:
@Zonciu noticed me here was a workaround https://github.com/dotnet/runtime/issues/31433

anddrzejb:
That is very cool. And should protect tbe components from future issues like this. But are you going to add it before v0.8 release?

PR #1355, opened 2021-04-15
fix(module:overlay): do not use currentculture in string interpolation

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #639

### 💡 Background and solution
Regional settings were being taken into account when doing string interpolation in styles. 

### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #1364, opened 2021-04-15
fix(module:avatar): size calculation & CssSizeLength functionality expansion

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
In `Avatar` component `Size` parameter was not doing anything when size was explicitly passed (for example "45.55px" did nothing).

### 💡 Background and solution
Following the advice from PR #1355 I extended `CssSizeLength` to handle decimal & double data types. Also, it can handle no unit (need for example in `scale` or `z-index`). I created tests to check how the handling of different numeric data types works. 

I fixed the issue with `Avatar` and `Size` parameter.

Side note: `Avatar` was using a helper method `StyleHelper.ToCssPixel()`. It seems to me that `CssSizeLength` does same + much more so I also replaced it. I think this helper could be retired in favour of `CssSizeLength` in all occurrences. 

### DIFF
--- a/components/avatar/Avatar.razor
+++ b/components/avatar/Avatar.razor
@@ -3,7 +3,7 @@
 
 @if (Position == null || (Position == "shown" && !Overflow) || (Position == "hidden" && Overflow))
 {
-    <span class="@ClassMapper.Class" style="@Style" id="@Id" @ref="Ref">
+    <span class="@ClassMapper.Class" style="@_sizeStyles @Style" id="@Id" @ref="Ref">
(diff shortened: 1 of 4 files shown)


ElderJames (approved):
The unit test is excellent!😀

Issue #1384, opened 2021-04-18
Table: column header sorter not refresh after ClearSorter is called


anranruye:
### Describe the bug

Column header sorter not refresh after ClearSorter is called if all parameters are value type or string.

### Steps to reproduce (please include code)

The following code is the classic WeatherForecast demo. All columns are changed to use DataIndex and remove all paramters which are not value type or string. Also, we are not using multiple sorter.

```C#
@using System.ComponentModel
@using AntDesign.TableModels

<Table @ref="table"
       TItem="WeatherForecast"
       DataSource="@forecasts"
       Total="_total"
       @bind-PageIndex="_pageIndex"
       @bind-PageSize="_pageSize"
       @bind-SelectedRows="selectedRows"
       OnChange="OnChange">
    <Selection Key="@(context.Id.ToString())" />
    <Column Field="@context.Id" DataIndex="Id" Sortable />
    <Column Field="@context.Date" Format="yyyy-MM-dd" DataIndex="Date" Sortable />
    <Column Field="@context.TemperatureC" DataIndex="TemperatureC" Sortable />
    <Column Title="Temp. (F)" TData="int?" DataIndex="TemperatureF" Sortable />
    <Column Title="Hot" Field="@context.Hot" DataIndex="Hot" Sortable></Column>
    <Column Field="@context.Summary" DataIndex="Summary" Sortable />
    <ActionColumn>
        <Space>
            <SpaceItem><Button Danger OnClick="()=>Delete(context.Id)">Delete</Button></SpaceItem>
        </Space>
    </ActionColumn>
</Table>

<br />
<p>PageIndex: @_pageIndex | PageSize: @_pageSize | Total: @_total</p>

<br />
<h5>selections:</h5>
@if (selectedRows != null && selectedRows.Any())
{
    <Button Danger Size="small" OnClick="@(e => { selectedRows = null; })">Clear</Button>

    @foreach (var selected in selectedRows)
    {
        <Tag @key="selected.Id" Closable OnClose="e=>RemoveSelection(selected.Id)">@selected.Id - @selected.Summary</Tag>
    }
}

<Button OnClick="()=> { _pageIndex--; }">Previous page</Button>
<Button OnClick="()=> { _pageIndex++; }">Next Page</Button>

@using System.Text.Json;
@code {

    WeatherForecast[] forecasts;

    IEnumerable<WeatherForecast> selectedRows;
    ITable table;

    int _pageIndex = 1;
[...]

PR #1385, opened 2021-04-18
fix(module:table): column header sorter not refresh after ClearSorter is called

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

fixes https://github.com/ant-design-blazor/ant-design-blazor/issues/1384

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix column header sorter not refresh after ClearSorter is called |
| 🇨🇳 Chinese | 修复清除排序方向后列头排序状态不刷新的问题 |

### DIFF
--- a/components/table/Column.razor.cs
+++ b/components/table/Column.razor.cs
@@ -114,7 +114,11 @@ namespace AntDesign
 
         public LambdaExpression GetFieldExpression { get; private set; }
 
-        void IFieldColumn.ClearSorter() => SetSorter(SortDirection.None);
+        void IFieldColumn.ClearSorter()
+        {
+            SetSorter(SortDirection.None);
+            StateHasChanged();
+        }
 
         private static readonly EventCallbackFactory _callbackFactory = new EventCallbackFactory();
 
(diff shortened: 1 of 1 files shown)


components/table/Column.razor.cs
```diff
@@ -114,7 +114,11 @@ public TData Field
 
         public LambdaExpression GetFieldExpression { get; private set; }
 
-        void IFieldColumn.ClearSorter() => SetSorter(SortDirection.None);
+        void IFieldColumn.ClearSorter()
+        {
+            SetSorter(SortDirection.None);
+            StateHasChanged();
```
ElderJames, line 120: Does this problem only occur when using DataIndex? If so, please add a judgment condition.

ElderJames (approved):

PR #1392, opened 2021-04-20
feat: use ResizeObserver Api instead of window.resize

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Slow refresh when window is resized. Lack of API to react to element size changes.

### 💡 Background and solution
I wanted to add ability to observe a specific element for size changes. Currently this was achieved only by using js event `window.resize`. But what if element needs to be observed for size changes not originating from resize event? On top of that, `window.resize` is slow and I wanted to have something that is much faster. That is why I added support for [Resize Observer API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver).

One can observe reaction time improvement by observing reactive select component. First gif will show how it was behaving with old approach, using `window.resize`. Notice, that redrawing of the component is a bit behind. When component is getting smaller, it becomes 2 lines for a moment until recalculation is done and more tag items are aggregated into a single tag. Same is happening the other way around - there is a moment where `Select` component can easily fit all tags, but it still shows `4+...`.

selectContainer_windowsResizeSlowReaction

Second gif is showing the same component but using Resize Observer API. The reaction speed issues are gone.

selectContainer_resizeObserverFastReaction

This API should probably replace all instances where `window.resize` is used to evaluate element size change.
The observer api has more observers (intersection, mutation) that should be quite easy to implement following what I did here. 
A question: any reason why we are not on ES6 but on ES2015?
A suggestion: maybe it would be a good idea to reorganize interop to be more structured. It is starting to be quite hard to maintain. 
A request: please pay attention to my approach to `JSInteropConstants`. I think this organization style is clearer. 
A disclaimer: I heavily based the solution on what I found in [blazor.matchbox](https://github.com/blacktau/blazor.matchbox) repository.
### 📝 Changelog
There are no changes exposed to the end user. This is purely for internal use.


ElderJames:
> A question: any reason why we are not on ES6 but on ES2015?

We may need to support IE11, which was supported by the 3.1 LTS.

> A suggestion: maybe it would be a good idea to reorganize interop to be more structured. It is starting to be quite hard to maintain.

Good suggestion. And I also want to wrap all the JS interop methods as the extensions methods for the `IJSRuntime`.

anddrzejb:
> > A question: any reason why we are not on ES6 but on ES2015?
> 
> We may need to support IE11, which was supported by the 3.1 LTS.

Oh...so that basically means we need to write our own `Virtualize` method (we won't be able to use the one in net5 without compiler preprocessor directives). Anyway, that is for future considerations.

MutatePat:
> > > A question: any reason why we are not on ES6 but on ES2015?
> > 
> > 
> > We may need to support IE11, which was supported by the 3.1 LTS.
> 
> Oh...so that basically means we need to write our own `Virtualize` method (we won't be able to use the one in net5 without compiler preprocessor directives). Anyway, that is for future considerations.

This is quiet upsetting as I believe the Virtualisation component does a lot of work behind the scenes, so rewriting and supporting it could prove a hassle. Is it possible to use directives and just disable Virtualize on clients which don't have the supported features? Then using "error boundaries" in .net 6 to show that their browser isn't supported?

anddrzejb:
Until now I only once tried Virtualisation, but failed. I still plan to, but I do hope someone will do it before me :wink:. Anyway, we probably could revert to rendering whole list for not supported versions...It will be terribly slow, but that is the trade-off when agreeing to technical debt. On the other hand, due to complexity of the `Select` or `Table` components, it may be necessary to recreate virtualisation (I really hope not...)

ElderJames:
I don't think it's necessary to implement virtualization features for users of 3.1. We can reserve `RenderFragement` slots for users to use `Virtualize` component themselves.

MutatePat:
> I don't think it's necessary to implement virtualization features for users of 3.1. We can reserve `RenderFragement` slots for users to use `Virtualize` component themselves.

Could this be done in the current version of AntDesign or are code changes required?

ElderJames:
@MutatePat  Not yet. At least Table has some work to do, providing users with the ability to organize their own rows.

ElderJames (dismissed):

ElderJames:
@anddrzejb  I find that the ResizeObserver API was not supported by IE. So can we fallback to `window.resize` for IE users?

anddrzejb:
Sure. Do we have a built-in recognition of IE?

ElderJames:
Maybe😂

anddrzejb:
Hmmm. Ok, what about a polyfill? I had no dealings with one, but maybe it is the right time to start...

MutatePat:
> Hmmm. Ok, what about a polyfill? I had no dealings with one, but maybe it is the right time to start...

Could you get the browser agent and use that to check? Then we could have a Feature Flag manager which can disable/enable based on this or user settings?

anddrzejb:
I guess this could be done as well. But isn't this more involved? I mean polyfill is meant to kick in if particular functionality does not exist. Otherwise, just move on. And also it still provides similar or the same functionality thus keeping the code simpler.
Don't get me wrong, I am not against feature flag manager, just looking for the most optimal solution.

MutatePat:
> I guess this could be done as well. But isn't this more involved? I mean polyfill is meant to kick in if particular functionality does not exist. Otherwise, just move on. And also it still provides similar or the same functionality thus keeping the code simpler.
> Don't get me wrong, I am not against feature flag manager, just looking for the most optimal solution.

This is true, but I assume the Feature Flag Manager could serve other purposes too, so I believe code reusability would be higher than polyfill( though I am not familiar with them)

anddrzejb:
> This is true, but I assume the Feature Flag Manager could serve other purposes too, so I believe code reusability would be higher than polyfill( though I am not familiar with them)

Well in this case polyfill would still provide same functionality. My understanding is that feature would either exclude particular functionality or switch to a different implementation (didn't use it yet either). Please correct me if I am wrong.

anddrzejb:
@ElderJames  Well, polyfill did not work... Seems like it wanted es6, but not sure. I reverted back to `window.resize` but on a global level. 
The funny part is that my tests of responsive select still do not work in IE. This time this is not due to `window.resize` but due to the fact that blazor get 0 values from `getBoundingClientRect`. The curious part is that JS is returning valid values, but `IJSRuntime` I think is failing. Look at the screenshot:
image

The values are there. But when I inspect the returned `DomRect` object from `IJSRuntime.InvokeAsync<DomRect>()`, all I see is zeros. I did install [ie polyfill lib](https://github.com/Daddoon/Blazor.Polyfill). The app is starting and does not raise any errors. But for example `Select` dropdown is not shown at all. I think there are many more issues with AntBlazor in IE11 than just this. 

...
I just had a look and boy, half of the functionality is dead... I mean none of the overlays are working (select, autocomplete, pickers, notification, popups, tooltips, etc), `InputNumber`, `Select`, `Slider` (in my case completely crashes the app), some Table `versions` (the one with long columns)...I guess I could continue checking, but the whole thing is kind-of useless in IE11...

ElderJames:
Well, it seems that we have to give up IE11 😂

anddrzejb:
I can only applaud that idea 👏🏻 🎊 .
So...shall I revert the commit 24f1300?

ElderJames:
This seems to be needed... Sorry to waste your hard work

anddrzejb:
I was thinking...maybe we leave this, but change the code from `IsIE11()` to `IsResizeObserverSupported()`? This API is available for some time now but there are old browser versions that do not support it. What do you think?

ElderJames:
I see. But it need one more interop...

anddrzejb:
Not really, because we need to switch to a different event. But notice what I did in `DomEventService` - the IsIE11 is only called once throughout application lifetime. So whenever again observer is requested, the app already knows if it can use observer api or needs to fall back to window resize event.

ElderJames:
That's fine.

anddrzejb:
@ElderJames Commit 43f3b45 switched to `IsResizeObserverSupported()`.

components/core/JsInterop/DomEventService.cs
```diff
@@ -82,6 +87,71 @@ public void AddEventListenerToFirstChild<T>(object dom, string eventName, Action
             });
         }
 
+        public async ValueTask AddResizeObserver(ElementReference dom, Action<List<ResizeObserverEntry>> callback)
+        {
+            string key = FormatKey(dom.Id, nameof(JSInteropConstants.ObserverConstants.Resize));
+            if (!(await IsResizeObserverSupported()))
+            {
+                Console.WriteLine("Registering window.resize");
```
ElderJames, line 95: Please remove the `Console.WriteLine`

ElderJames (changes requested):

ElderJames (approved):

Issue #1041, opened 2021-01-25
Switch Checked property always returns 'false'


ClumsyPenguin:
### Describe the bug
When you use the Checked property.
![image](https://user-images.githubusercontent.com/22469147/105697063-6097c880-5f04-11eb-8787-844c68a3d94b.png)
It always return false whether the switch is true or false...
![image](https://user-images.githubusercontent.com/22469147/105697051-57a6f700-5f04-11eb-81a0-d82b91555644.png)

**Note: The switch in this case is not being used in a form.**
### Workaround

Use @bind-Value instead of Checked
```html
<Switch style="float:right" @bind-Value="colorSetting" OnChange="OnChangeSetting">
            </Switch>
```
### Steps to reproduce (please include code)
```html
<Switch style="float:right" Checked="colorSetting" OnChange="OnChangeSetting"></Switch>
```
```csharp
private bool colorSetting { get; set; }
private void OnChangeSetting()
{
    Console.WriteLine(colorSetting);
}
```

### Further technical details
- AntDesign Nuget Package version: 0.5.3
```
.NET SDK (reflecting any global.json):
 Version:   5.0.102
 Commit:    71365b4d42

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19041
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.102\

Host (useful for support):
  Version: 5.0.2
  Commit:  cb5f173b96

.NET SDKs installed:
  1.1.8 [C:\Program Files\dotnet\sdk]
  1.1.14 [C:\Program Files\dotnet\sdk]
  2.0.0 [C:\Program Files\dotnet\sdk]
  2.1.104 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.403 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]
  2.1.502 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]
  2.1.504 [C:\Program Files\dotnet\sdk]
  2.1.505 [C:\Program Files\dotnet\sdk]
  2.1.507 [C:\Program Files\dotnet\sdk]
  2.1.509 [C:\Program Files\dotnet\sdk]
  2.1.520 [C:\Program Files\dotnet\sdk]
[...]

PR #1394, opened 2021-04-20
fix(module: checkbox & switch) Checked parameter binding

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Refers to #1041 and to [this comment](https://github.com/ant-design-blazor/ant-design-blazor/issues/1041#issuecomment-795615166)

### 💡 Background and solution
My understanding is that `Checked` and `Value` for these 2 components serve exactly the same purpose. So I made them behave the same way.
It seems to me that `Checkbox` and `Switch` have base functionality exactly the same, but for some reason they are written as 2 separate components. How about if we try to use one as the base for the other? If you agree, please do not merge yet. Let me know and I will try to apply my suggestion.

### 📝 Changelog
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  `Checkbox` and `Switch` allow now binding to `Changed` property. `Value` and `Changed` properties can be used interchangeably.          |
| 🇨🇳 Chinese |           |


ElderJames:
Sure, you can try to merge them in a one base class.

anddrzejb:
@ElderJames Commit aa888d8 contains the merge. Please have a look and let me know if things are ok. In the meantime I will start doing tests for the 2 components and once you sign off on them, I will also fix the docs. There are no breaking changes, but I would like to change the docs to indicate ability to bind to the `Checked` property.

ElderJames:
It looks fine.

anddrzejb:
@ElderJames  Ok, I added tests and changed the docs. I also forgot to mention that I removed from `Switch` property `Control`. I could not find any use for it in the component. It also was not mentioned in the docs (neither in antD docs). If it should be included, please let me know (and I would appreciate if you could tell me what is the purpose of that parameter).

ElderJames:
The `Control` is from ng-zorro, https://ng.ant.design/components/switch/zh#components-switch-demo-control.

anddrzejb:
@ElderJames Commit 5755ffa contains the changes that reintroduce the `Control` property. Previously it was rather useless without accompanying `OnClick` event (at least it seems so to me - based on the ng-zorro docs you provided). 

I also have a question/request regarding tests - would you be ok with using [Fluent Assertions library](https://fluentassertions.com)? I find it very useful during testing.
I also do not fully understand the tests naming convention. Is there any? I myself am in favour of [this approach](https://stackoverflow.com/a/7398606/3239998), but in the long run I will adapt to any existing rule is there is such.

anddrzejb:
Since I was doing this overhaul, I also decided to fix `CheckboxGroup` layout docs. This also requires some changes in `CheckboxGroup`. Currently it seems that there was an intention for the consumer to mix manually rendered checkboxes with passed as options:
```razor
    <CheckboxGroup Style="width:100%"  Options="@plainOptions">

        <Row>
            <Col Span="8">
                <Checkbox Value="A" >A</Checkbox>
            </Col>
            <Col Span="8">
                <Checkbox Value="B"  >B</Checkbox>
            </Col>
        </Row>
    </CheckboxGroup>
@code {
    string[] plainOptions = { "Apple", "Pear", "Orange" };
}
```

This is currently obviously not working. But is this assumption correct? This is much more involved than simple allowing user to either use `ChildContent` or `Options`. So what was the intention here?

ElderJames:
Thank you for the work about the `Control`.

I'm not sure it's really beneficial to use Fluent Assertions, my thought is ”Entities should not be multiplied unnecessarily“. But we are open for discussion.

For the naming of test cases, my opinion is to follow the Microsoft official specification.

I don't think we should consider mixing Options and ChildContent.. React also didn't..

anddrzejb:
- I don't even know what is microsoft's approach. But will have a look.
- `FluentAssertions` - what do you mean `Entities`? In my opinion in this case there is a huge gain, becase `FluentAssertions` offers a ton of functionality that can seriously speed-up tests (for example in collections area). Obviously it will not replace `Assert`, it can still be used. I am also not hell-bent on that lib. If you don't want it, then no problem. I am just trying to limit the amount of work tests require to run. 
- Mixing in `CheckboxGroup` - I almost finished 🤣 . I will post it with docs (without tests though) and you will have a look. If you won't like it, I will remove it. I decided to do it because I was thinking about a use case when a user would like to have a constant set of checkboxes available + anything rendered dynamically.

ElderJames:
If it does speed up writing tests, then of course I'm happy to use it.

Now that if the mixing work is done, keep it. Hah, it is actually nice feature. Thanks.

anddrzejb:
@ElderJames Done. Still without tests. Please have a look at the top of `CheckboxGroup.razor.cs` where I have defined the enum. I have a feeling this is not the place you would like to have that enum. If that is the case, where shall I put it? Core? 
If you are ok with the way it looks, I will create tests for this. 
Also, what is lacking here is switching the mode. I mean if after rendering a consumer wishes to switch the modes (so for example now the `Options` checkboxes should be rendered first), the `CheckboxGroup` will not rerender. So changes to mode after initialization have no effect.

components/checkbox/CheckboxGroup.razor.cs
```diff
@@ -58,32 +114,85 @@ protected override void OnInitialized()
...
+                                                    .ToArray(),
+                        arr => _currentValue = () => _selectedValues);
+                }
+                else
+                {
+                    var workWith = (_isOptionDefined ? Options : _constructedOptions);
+                    workWith.Match(
+                        opt => _currentValue = () => opt.Where(x => x.Checked).Select(x => x.Value).ToArray(),
+                        arr => _currentValue = () => _selectedValues);
+                }
```
ElderJames, line 150: Could you please put these logic code into a method? There are many if-else here.

components/checkbox/CheckboxGroup.razor.cs
```diff
@@ -6,20 +6,42 @@
 
 namespace AntDesign
 {
+    public enum CheckboxGroupMixedMode
+    {
+        ChildContentFirst,
+        OptionsFirst
+    };
```
ElderJames, line 13: You can put it in the checkbox folder.

components/checkbox/CheckboxGroup.razor.cs
```diff
@@ -35,15 +57,49 @@ internal void AddItem(Checkbox checkbox)
...
+        {
+            if (Options.Value is not null)
+            {
+                if (ChildContent is not null)
+                {
+                    bool result = false;
+                    Options.Switch(
+                        opt => result = opt.Any(o => o.Label.Equals(checkbox.Label)),
+                        arr => result = arr.Contains(checkbox.Label));
+                    return result;
```
ElderJames, line 82: You maybe could use the `result = Options.Match()` .

components/checkbox/CheckboxGroup.razor.cs
```diff
@@ -35,15 +57,49 @@ internal void AddItem(Checkbox checkbox)
...
+            {
+                checkbox.SetValue(_selectedValues.Contains(checkbox.Label));
+                if (_indexConstructedOptionsOffset == -1)
+                    _indexConstructedOptionsOffset = _checkboxItems.Count - 1;
+            }
+            else if (checkbox.IsFromOptions && _indexSetOptionsOffset == -1)
+                _indexSetOptionsOffset = _checkboxItems.Count - 1;
+        }
+
+        private bool IsIsFromOptions(Checkbox checkbox)
```
ElderJames, line 72: This naming is a bit strange. I think `IsCheckboxFromOptions` is better.

components/core/Base/AntInputBoolComponentBase.cs
```diff
@@ -0,0 +1,66 @@
...
+        public bool Disabled { get; set; }
+
+        internal virtual string PrefixCls { get; }
+
+        protected virtual void SetClass()
+        {
+            ClassMapper.Clear()
+                .Add(PrefixCls)
+                .If($"{PrefixCls}-rtl", () => RTL);
+        }
```
ElderJames, line 44: I think it's no need to set class in base class.

components/checkbox/CheckboxGroup.razor.cs
anddrzejb: Sure. I was planning to do that but rushed to get this out. I would have probably fixed it during test creation.

components/checkbox/CheckboxGroup.razor.cs
anddrzejb: Ha. Forgot about that 😉

components/core/Base/AntInputBoolComponentBase.cs
anddrzejb: Do you want me to move `PrefixCls` out of base class as well?

ElderJames: Yes, it doesn't seem to make much sense to separate it from the specific component.

site/AntDesign.Docs/Demos/Components/Checkbox/doc/index.en-US.md
```diff
@@ -24,14 +24,15 @@ Checkbox
...
-| CheckedChange |The callback function that is triggered when the state changes| function(e)|-     |
+| OnChange |The callback function that is triggered when the state changes| function(e)|-     |
 
 Checkbox Group
 
 | Property | Description | Type | Default Value |
 | --- | --- | --- | --- |
 | CheckboxItems | Check box items                             | IList<AntCheckbox>        | -         |
 | Disable | Disable all checkbox                               | boolean        | false         |
+| MixedMode            | Applies only when bot `Options` and `ChildContent` is used. Allows to select which section (`Options` or `ChildContent`) will be rendered first.           | (enum)CheckboxGroupMixedMode         |ChildContentFirst       |
```
ElderJames, line 35: The parameter is naming `MixMode`.

anddrzejb: Hmm...It was supposed to be `MixedMode`. Wrong parameter name.

ElderJames:
@anddrzejb  In the mixed mode demo, I added a button to switch the mixed mode. But it throw a exception `Index was outside the bounds of the array` when I check the `Kiwi`. Could you have any idea to fix it?

anddrzejb:
Can you post your demo code please?

ElderJames:
Just in the mixed demo I have modified in https://github.com/ant-design-blazor/ant-design-blazor/pull/1394/commits/3f272ed0eeb2989eca9a901d04a2994198845b1c#diff-9ab515a947eb47e627193c71e58e7d5178816396ad225d000b6068aebd77260dR32

anddrzejb:
I mentioned [here](https://github.com/ant-design-blazor/ant-design-blazor/pull/1394#issuecomment-824694668) that switching live is not yet supported. However, if you think it is important to have it, I can try to do it.

ElderJames (approved):

ElderJames:
Great!

Issue #1435, opened 2021-04-28
Drawer OffsetX not working


MutatePat:
### Describe the bug

When using DrawerService to create a Drawer, offsetX does not work in left or right placement mode.

Interestingly, there is no reference to OffsetX in ant.design react

### Steps to reproduce (please include code)
```
DrawerOptions options = new DrawerOptions
{
    Closable = true,
    Width = 350,
    Title = "Settings",
    Placement = "left",
    OffsetX = 500
};

DrawerService.CreateAsync<MyForm, string, string>(options, "hello world");
```

### Further technical details
- AntDesign Nuget Package version: 0.8.0
- Include the output of `dotnet --info`:
```
.NET SDK (reflecting any global.json):
 Version:   5.0.200
 Commit:    70b3e65d53

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.200\

Host (useful for support):
  Version: 5.0.3
  Commit:  c636bbdc8a

.NET SDKs installed:
  5.0.100-preview.5.20279.10 [C:\Program Files\dotnet\sdk]
  5.0.200 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.23 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.25 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.23 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.25 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.10 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.0-preview.5.20279.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #1448, opened 2021-05-01
fix(module: drawer): OffsetX and offsetY do not work (#1435)

zxyao145:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#1435 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix OffsetX and offsetY do not work of Drawer, and update the documents how to use DrawerService by the way. |
| 🇨🇳 Chinese |  修复 Drawer OffsetX 和offsetY 不起作用，并更新使用 DrawerService 的文档。|


components/core/JsInterop/ObservableApi/resizeObserver.ts
```diff
@@ -1,8 +1,10 @@
 ﻿import { getDom } from '../interop';
 
+// @ts-ignore
```
anddrzejb, line 3: May I ask what errors were you getting?

zxyao145: > Error TS2304: Cannot find name 'ResizeObserver'. 

Maybe because it's experimental? (In fact, all modern browsers support it except IE)

anddrzejb: I don't think it is experimental any longer. I cannot find any reference to it being experimental... We do have a detection in place if it does not exist and a fallback to window.resize. On the other hand, I think IE will [not be supported by AntBlazor anylonger.](https://github.com/ant-design-blazor/ant-design-blazor/pull/1392#issuecomment-825404976)

To be honest I never got that error. But I guess there is no harm to have it here though 👍🏻 .

zxyao145: Yes, the word "experimental" does not exist in the English document of MDN. Maybe the Chinese document of MDN has not been updated.

anddrzejb:
@zxyao145 There is a rendering problem with drawer. Please have a look at the gif. Notice that first time `Left` is clicked, the drawer seems to appear out of place. Second time is ok. The same applies to `Right`. If `Left` is clicked again after clicking on `Right`, whole flickering process repeats.
drawerOffset_appear

To be honest it looks awfully similar to a problem that I was dealing with `DatePicker` & `Overlay` in PR #1452. If you check the first comment, it is the point 2. I don't know if this is going to be helpful, but to be honest it does look like css class "leftover".

zxyao145:
> @zxyao145 There is a rendering problem with drawer. Please have a look at the gif. Notice that first time `Left` is clicked, the drawer seems to appear out of place. Second time is ok. The same applies to `Right`. If `Left` is clicked again after clicking on `Right`, whole flickering process repeats.
> drawerOffset_appear
> 
> To be honest it looks awfully similar to a problem that I was dealing with `DatePicker` & `Overlay` in PR #1452. If you check the first comment, it is the point 2. I don't know if this is going to be helpful, but to be honest it does look like css class "leftover".

Yes, that's the problem with demo. I'm repairing it locally

anddrzejb:
@zxyao145 I think you need to update your branch for the build errors to go away.

anddrzejb (approved):

ElderJames (approved):

ElderJames:
@zxyao145  I found that the closing animation was lost.

zxyao145:
> @zxyao145 I found that the closing animation was lost.

fixed

ElderJames:
The drawer opening animation has a sense of paragraph, not smooth enough.

ElderJames (approved):

Issue #1429, opened 2021-04-28
DatePicker: week count, date&time picker focus, quarter range & placeholder


anddrzejb:
### Describe the bug
I found several bugs in date/range picker.
~1. Week picker - week count for beginning of the year is showing incorrectly. For example first week in 2020 is actually starting from the second~
![image](https://user-images.githubusercontent.com/6518006/116353249-183bbf80-a7ff-11eb-87cc-e305e10982cc.png)
~and it should show like this:~
![image](https://user-images.githubusercontent.com/6518006/116353437-6d77d100-a7ff-11eb-91a9-2ae8107ab552.png)
Solved by #1452 

2. Date & Time picker - time is not focused and when changing time using keyboard, it does not focus either:
![image](https://user-images.githubusercontent.com/6518006/116353568-aca62200-a7ff-11eb-8bf0-aaac1618f12e.png)

3. Date & Time range picker - after selecting the date and clicking `Ok` button, the focus should move to the second date. Instead, the picker is closed and focus is taken off the input:
![dateTimePickerRange_2ndPickerFocus](https://user-images.githubusercontent.com/6518006/116354069-892fa700-a800-11eb-9ee5-0f2f3cf28763.gif)
Expected behaviour:
![dateTimePickerRange_2ndPickerFocus_Expected](https://user-images.githubusercontent.com/6518006/116354149-a8c6cf80-a800-11eb-9b2b-4298bace731a.gif)

4. Quarter range: When I set the end quarter, the same quarter of the start quarter is disabled, it should be enabled.
![image](https://user-images.githubusercontent.com/10374056/116251711-818ce580-a7a1-11eb-9e3d-20774be17a40.png)

5. Placeholder issue. 
  - the attribute `Placeholder` has no effect unless it is placed after `PickerType`
  - if it is visible, it is only visible after initialization of the component. If the value is set and the cleared, default placeholder is shown
![datePickerRange_vanishingPlaceholder](https://user-images.githubusercontent.com/6518006/116356879-81720180-a804-11eb-9f9c-9d2cdfca978e.gif)

### Steps to reproduce (please include code)
Except for the 5th point, all the other are reproduceable using the docs.
To reproduce the 5th, use this code (last 2 pickers will have a custom placeholder and it will vanish if cleared):
```razor
<DatePicker TValue="DateTime?" Placeholder="@("Will vanish on clear")" AllowClear Picker="@DatePickerType.Date" />
[...]

PR #1452, opened 2021-05-02
fix(module:datepicker): Date picker panel click closing + some issues from #1431

### LINKED ISSUE #1431: DatePicker: The overlay panel closes when click any place
### Describe the bug

After update to 0.8.1-nightly-2104280412, the DatePicker overlay panel closes when click any place. If we want to use the header buttons, this causes troubles. 
image

### Further technical details
- AntDesign Nuget Package version: 0.8.1-nightly-2104280412

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
1. Fixes #1431 
2. Addresses parts of #1429 ([this comment](https://github.com/ant-design-blazor/ant-design-blazor/issues/1429#issuecomment-828196477)):
   a. When first clicked, the RangePicker panel will first display on the left side of the page.
   b. The height of the RangePicker is refreshed twice when it is opened.
3. When panel was opening, `OnBlur` event is being fired.
4. When Tab+Shift on `RangePicker` was leaving the input, it is not loosing focus.
5. Initial render of `RangePicker` if rendered above the input is placed too low when panel renedered above input element:
image
Closing and reopening the panel would fix that to expected behavior:
image

### 💡 Background and solution
1. `DatePicker` in `OnBlur` event handler was calling `Close()` method and should not.
2. a. This was an issue with `Overlay` manifesting only in server-side. I guess it wasn't manifesting anywhere else except `RangePicker` because of how complex the `RangePicker` is. Anyway, the issue there was that the `RangePicker` was starting with style set to `display: inline-flex` and then its position was calculated and adjusted to match input element. I changed that initial style to `display: inline-flex; visibility: hidden;` - we still need this to render and take space but not to be visible. 
2. b. It wasn't really refreshing height 2 times. The problem was that when panel should hide, it receives class `ant-slide-up-leave-active`. That class stays there. So when it is shown again, it is made first visible with that class, and then it gets proper `ant-slide-up-enter-active`. So basically on any consecutive open the picker was first being animated with leave and then with enter css class. My solution is to reset the class to enter class after it is closed.
3. Removed `Close()` method call.
4. Missing `AutoFocus = false` was added.
5. `RangePicker` was supposed to get `OverlayClassName="ant-picker-dropdown-range`. Otherwise first render was placing the pointing arrow incorrectly. It was only affecting the panel if it was rendered above the input.

Feature: added support for `Escape` button - matches the behaviour of react antD.

### DIFF
--- a/components/core/Component/Overlay/Overlay.razor.cs
+++ b/components/core/Component/Overlay/Overlay.razor.cs
@@ -85,6 +85,12 @@ namespace AntDesign.Internal
 
         private int _overlayClientWidth = 0;
 
+        protected override void OnInitialized()
+        {
+            _overlayCls = Trigger.GetOverlayHiddenClass();
+            base.OnInitialized();
+        }
+
         protected override async Task OnParametersSetAsync()
         {
             if (!_isOverlayShow && Trigger.Visible && !_preVisible)
@@ -156,9 +162,8 @@ namespace AntDesign.Internal
                 return;
             }
             await Task.Yield();
-
             HtmlElement trigger;
-            if (Trigger.ChildContent != null)
+            if (Trigger.ChildContent is not null)
             {
                 trigger = await JsInvokeAsync<HtmlElement>(JSInteropConstants.GetFirstChildDomInfo, Trigger.Ref);
                 // fix bug in submenu: Overlay show when OvelayTrigger is not rendered complete.
@@ -583,6 +588,7 @@ namespace AntDesign.Internal
             if (!_isOverlayShow && !_isWaitForOverlayFirstRender)
             {
                 overlayCls = Trigger.GetOverlayHiddenClass();
+                _overlayCls = Trigger.GetOverlayEnterClass();
             }
             else
             {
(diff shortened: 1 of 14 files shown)


anddrzejb:
@Zonciu Could you review my commit 934acca? I was trying to push it as a new PR but for some reason cherry-pick was still picking up commit from this PR...

Commit fixes [this part](https://github.com/ant-design-blazor/ant-design-blazor/issues/1429#issuecomment-828208070) of #1429.

What I found was that `DatePickerPanelBase.Value` that was not nullable. If `Value` passed from `DatePickerBase` was null, then `DateTime.Now` was assumed. Changed it, so value in `DatePickerPanelBase` is a nullable.

Zonciu:
Sorry I'm too busy recently.

anddrzejb:
Commit 7ad46ed addresses issues associated with week mode:
> 1. Week picker - week count for beginning of the year is showing incorrectly. For example first week in 2020 is actually starting from the second
>    image
>    and it should show like this:
>    image

mutouzdl:
@anddrzejb 
I looked it over and found nothing wrong.

By the way, I found a problem, but this problem exists in the current version(0.8.0).
RangePicker(date mode with timer) cannot select the correct end date, nor can you select months forward.
Steps:
1. Choose a start date
2. Select an end date, and you will find that the end date is always one month longer than the selected date, and you cannot switch to the previous month

-----------translate-----------
我检测了一遍，没有问题。

顺便提一下，我发现一个问题，但这个问题在当前版本(0.8.0)中就存在了。
RangePicker（带时间的日期模式）不能选择正确的结束日期，也无法向前选择月份。

步骤：
1. 选择一个开始日期
2. 选择一个结束日期，此时会发现，结束日期会永远比选择的日期多1个月，并且无法切换到上一个月份

issue

mutouzdl (approved):

anddrzejb:
@mutouzdl Thanks. Regarding the issue - I will add it to aggregated issue list #1429 we created with @Zonciu . I noticed there are several issues with `RangePicker` with timer, so this adds another one...

ElderJames (approved):

PR #1464, opened 2021-05-04
feat(module:form): expose EditContext

### LINKED ISSUE #1128: Subscribe to OnFieldChanged when using the Form component
I am trying to subscribe to the `EditContext`'s `OnFieldChanged` event while using the `Form` component. As I understand it, this is not possible since `_editContext` is private and is created within the component ([Form.razor.cs](https://github.com/ant-design-blazor/ant-design-blazor/blob/master/components/form/Form.razor.cs))

Potentially allow an `EditContext` to be passed to a `Form` reference? Or create methods to expose the `OnFieldChanged` event? Or expose `Editcontext`?

### Additional context
Potential alternative solutions are welcomed.

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1128 

~I would like to expose `EditContext.GetValidationMessages()` method. Currently it is not available. My use case is that I have 3 components (Parent=>Child=>Grandchild). Each of the components is basically a form with one-to-many relationship and one-to-one on a dynamic component relationship. To the end user I want to show all validation messages in a single place. Without direct access to validation messages, it is impossible (or I cannot find it).~

#### Edit:

> I would like to expose `EditContext`. Currently it is not available. My use case is that I have 3 components (Parent=>Child=>Grandchild). Each of the components is basically a form with one-to-many relationship and one-to-one on a dynamic component relationship. To the end user I want to show all validation messages in a single place. Without direct access to validation messages, it is impossible (or I cannot find it). 

For the time being I am accessing `EditContext` using reflection, but this is just an ugly workaround.

### 💡 Background and solution
~Expose `EditContext.GetValidationMessages()` in `Form` component as a public method.~

If accepted, I will add it to API documentation (basically I guess I will create it as it seems it does not exist).

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Expose `EditContext` in `Form` component.         |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/Form.razor.cs
+++ b/components/form/Form.razor.cs
@@ -211,6 +211,6 @@ namespace AntDesign
             _editContext = new EditContext(Model);
         }
 
-        public IEnumerable<string> GetValidationMessages() => _editContext.GetValidationMessages();
+        public EditContext EditContext => _editContext;
     }
 }
--- a/components/form/Form.razor.cs
+++ b/components/form/Form.razor.cs
@@ -210,5 +210,7 @@ namespace AntDesign
         {
             _editContext = new EditContext(Model);
         }
+
+        public IEnumerable<string> GetValidationMessages() => _editContext.GetValidationMessages();
     }
 }


ElderJames:
There is a similar #1128 , do you think there are any side effects of exporting `EditContext` directly?

anddrzejb:
We could expose it only as a getter. I had a look at [`EditContext`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.forms.editcontext?view=aspnetcore-5.0) and there are a number of methods I have never used and I cannot really say how can they affect the `Form`/`FormItem`. I could probably run some test and play around with this to see what is what. I do have a feeling though we should allow the devs to access it.

anddrzejb:
@ElderJames I had a much more thorough look at exposing `EditContext` and to be honest, I cannot really see anything that will have side effects. The only thing that possibly could mess things up if people start playing with extension methods of `EditContext`, but I think these are advanced scenarios and anyway more advanced developers should have access to these methods (for example a dev can change what classes will be returned for modiflied/valid/invalid field).

PR #1469, opened 2021-05-05
fix(module:select): fix dropdown width for IE

### LINKED ISSUE #1468: Select: width of dropdown for IE is 0
### Describe the bug

image
image

### Further technical details
- AntDesign Nuget Package version: 0.8.0

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

fixes #1468 
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix dropdown width for IE |
| 🇨🇳 Chinese | 修复IE浏览器下拉选项宽度为0的问题 |


anddrzejb:
And this actually makes a difference? What IE version?

anranruye:
@anddrzejb 
IE 11 and Windows 10 embedded version. To run a blazor server app in IE, you need this polyfill: https://github.com/Daddoon/Blazor.Polyfill

anddrzejb:
Yeah, I know, I had it set up. But I cannot really see any difference coming from your change, especially from the language point of view. I cannot see any difference between the original object and the one you create...
On a side note I actually [had a look](https://github.com/ant-design-blazor/ant-design-blazor/pull/1392#issuecomment-824819383) at how AntBlazor is doing in IE and my overall experience was dreadful. You actually had a success here?

anranruye:
@anddrzejb 

Actually, I don't know how other parts of AntBlazor work in IE. I myself have no plan to support IE in my own apps. I just solve this one issue.

The story is, a group friend from Ding Talk asks about this issue today. He says, everything in his app work fine with the polyfill, except the `Select` component. I do the same research as you and guess the problem is that `DomRect` in IE is not serialized properly. Then I change the code to this one. It works in Windows 10 embedded version on my machine. And the one who asks for help says this also works for his IE 11.

anddrzejb:
This is truly baffling, as I really cannot see any difference from the language standpoint. However I am not a js guru. I guess @ElderJames will have to state his opinion on this. On the other hand, this IE situation should be investigated. Maybe we were too quick to drop it. Don't get me wrong, I do not want to support IE either. But if people are using it and find it functional, we should make sure we can support it if possible. 

I was thinking - do you have a minimal repro, that I could clone and try it myself? Maybe my previous tests had some kind of fundamental issue...

anranruye:
@anddrzejb 
I use this demo: https://antblazor.com/en-US/components/select#components-select-demo-basic.

Here's a repro: https://github.com/anranruye/BlazorIERepro

You can add this snippet to the _Host.cshtml file of an existed project:
```
    <script>
        if (/MSIE \d|Trident.*rv:/.test(navigator.userAgent)) {
            window.AntDesign.interop.getBoundingClientRect = function (element) {
                let dom = window.AntDesign.interop.getDom(element);
                if (dom && dom.getBoundingClientRect) {
                    let domRect = dom.getBoundingClientRect();
                    return {
                        width: domRect.width,
                        height: domRect.height,
                        top: domRect.top,
                        right: domRect.right,
                        bottom: domRect.bottom,
                        left: domRect.left,
                        x: domRect.x,
                        y: domRect.y
                    }
                }
                return null;
            }
        }
    </script>
```

anddrzejb:
@anranruye Thanks for preparing the repro. Now I see that I had exactly the same problem in my [investigation of IE](https://github.com/ant-design-blazor/ant-design-blazor/pull/1392#issuecomment-824819383) but I wasn't that thorough 😞 . 
What I found out is that IE, when `getBoundingClientRect()` is called, returns an object that is different when compared to other modern browsers (IE returns [TextRectangle](https://docs.microsoft.com/en-us/previous-versions/ms536433(v=vs.85))). So JSRuntime cannot match it to `DomRect` object, so it just returns empty. Your solution, even though for all other browser is completely repetitive, actually is a necessity for IE. Could you add a comment explaining why this repetition is there? Without it, I am sure it will be removed by someone in the future...

anranruye:
@anddrzejb 
feel free for the object name, in the Chinese version of this page https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect , the translator comments that `DOMRect` is the standard name for `TextRectangle` or `ClientRect`.

The most likely cause is:

> The returned DOMRect object can be modified in modern browsers. This was not true with older versions which effectively returned DOMRectReadOnly. With IE and Edge, not being able to add missing properties to their returned ClientRect, object prevents backfilling x and y.
> 
> Due to compatibility problems (see below), it is safest to rely on only properties left, top, right, and bottom.
> 
> Properties in the returned DOMRect object are not own properties. While the in operator and for...in will find returned properties, other APIs such as Object.keys() will fail. Moreover, and unexpectedly, the ES2015 and newer features such as Object.assign() and object rest/spread will fail to copy returned properties.

Can you help to research whether there is a use for `Object.assign()` or `object rest/spread` inside the blazor js interop? Or this issue is caused by missing properties(`X` and `Y`)?

anddrzejb:
@anranruye When I did the tests yesterday, I had a look at exactly what IE is returning. It not only did not contain `X` & `Y` but also contained other structures, while in other browsers it is just 8 properties. I actually tried removing `X` & `Y` from POCO `DomRect` and it still did not work. So I guess the other properties returned in IE are in the way.

anddrzejb (approved):

ElderJames (approved):

PR #1507, opened 2021-05-11
feat(module: form): support set validation rules on FormItem

mutouzdl:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

在某些情况下，我们需要复用model类，但同一个字段在不同的页面有可能出现不同的验证规则，目前的验证模式（给model字段附加特性）只能针对不同的页面创建新的model类。
所以，我加上了这个功能。默认仍然使用原来的验证模式，通过Form的ValidateMode参数切换验证模式，允许两种模式混用。

------------------------------translate------------------------------

In some cases, we need to reuse the Model class, but it is possible to have different validation rules for the same field on different pages. The current validation mode (adding validationAttribute to the model's field) can only create new model classes for different pages.
So, I added this feature.By default, the original validation mode is still used, and the validation mode is switched through the validateMode parameter of the Form, allowing a mix of the two modes.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     support set validation rules on FormItem      |
| 🇨🇳 Chinese |      支持在FormItem上添加验证规则     |

### DIFF
--- a/components/form/Form.razor.cs
+++ b/components/form/Form.razor.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using AntDesign.Forms;
 using AntDesign.Internal;
@@ -94,6 +95,9 @@ namespace AntDesign
         [Parameter]
         public bool ValidateOnChange { get; set; }
 
+        [Parameter]
+        public FormValidateMode ValidateMode { get; set; } = FormValidateMode.Default;
+
         private static readonly RenderFragment _defaultValidator = builder =>
         {
             builder.OpenComponent<ObjectGraphDataAnnotationsValidator>(0);
@@ -110,6 +114,8 @@ namespace AntDesign
         private IList<IControlValueAccessor> _controls = new List<IControlValueAccessor>();
         private TModel _model;
 
+        private RulesValidator _rulesValidator;
+
         ColLayoutParam IForm.WrapperCol => WrapperCol;
 
         ColLayoutParam IForm.LabelCol => LabelCol;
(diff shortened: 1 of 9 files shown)


components/form/Form.razor.cs
```diff
@@ -166,6 +179,52 @@ private async Task OnInvalidSubmit(EditContext editContext)
             await OnFinishFailed.InvokeAsync(editContext);
         }
 
+        private void OnFieldChanged(object sender, FieldChangedEventArgs args)
```
anddrzejb, line 182: @mutouzdl We are going to have some interesting merge issues when Feature branch gets merged. I am referring to my PR #1504 & for example [this line](https://github.com/ant-design-blazor/ant-design-blazor/blob/8942b61af093cdf0c3b4f83251cecc223916f490/components/form/Form.razor.cs#L161). 
Since we are here this is a good moment to discuss this future merge issue.
In that PR `EditContext` is exposed to the consumer. There the event is attached only to marshal it back to EventCallback. Here though it does much more (very interesting I must add). Can we actually do both here? It seems to me we probably shouldn't? I am not sure to be honest... I think that we should have a detection here - if EventCallback has a delegate, invoke it and skip the body of this method. Otherwise, run it. My reasoning behind it is - if a dev wants to use EventCallback, that means he/she wants to have full control over the component's behaviour. Then we should not force default behaviour on him/her.

mutouzdl: @anddrzejb 
I'll try to rearrange my code based on the Feature branch.

I'm going to resubmit the PR to the Feature branch.
If there are any new questions, let's discuss them in the new PR.

So I'm going to close this.

PR #1516, opened 2021-05-11
feat(module: form): support set validation rules on FormItem

mutouzdl:
### 🤔 This is a ...

- [x] New feature
- [x] Demo update
- [x] Test Case

### 💡 Background and solution

添加了以下功能：
1. FormItem支持Rules参数，和Ant Design React的FormItem类似。
大部分Rule都和Ant Design React的[Rule](https://ant.design/components/form-cn/#Rule)一致，"whitespace"和"validateTrigger"我没有实现，"enum"我使用“OneOf”代替了（功能上有点不一样）

2. Form支持ValidateMessages参数，可自定义验证错误信息（仅支持在Rules模式下使用）。我不确定这么做是不是最好的方式，如果有更优的方式，希望能提示我

3. 增加了相应的文档、单元测试

------------------------------translation------------------------------

The following features have been added:
1. FormItem supports Rules parameters, similar to Ant Design React's FormItem.
Most of the Rule are consistent with  Ant Design the React ’s [Rule](https://ant.design/components/form/#Rule) , but without "whitespace" and "validateTrigger", and  I used "OneOf" instead of "Enum" (It's a little different)

2. Form supports ValidateMessages parameter, which can be customized to validate error messages (only in Rules mode).I'm not sure if this is the best way to do it, but if there's a better way, hopefully you'll remind me

3. Added relevant documents and unit tests

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    support set validation rules on FormItem       |
| 🇨🇳 Chinese |     支持在FormItem上添加验证规则      |


mutouzdl:
我想把Rules改成可配置的，类似Ant Design React的做法，所以我把PR的状态改为Word In Process了

---------------------------translate-----------------------------

I want to change Rules to configurable, similar to Ant Design React, so I changed PR status to Word In Process

anddrzejb (changes requested):
Currently it does not seem to work:
form_multivalidation

Also, it is not obvious what is the difference between each mode. Only once I read the enum `FormValidateMode` comments, I understood. Future consumers will not understand until they try to use it. Maybe when the demo starts working it will be more obvious, but I think it wouldn't hurt to include a very basic explanation in the demo description. 

Aside that, really cool feature 🚀 .

components/form/Form.razor.cs
```diff
@@ -202,6 +216,52 @@ private async Task OnInvalidSubmit(EditContext editContext)
...
         }
 
+        private void RulesModeOnFieldChanged(object sender, FieldChangedEventArgs args)
+        {
+            if (!ValidateMode.IsIn(FormValidateMode.Rules, FormValidateMode.Complex))
+            {
+                return;
+            }
+
+            _rulesValidator.ClearErrors();
```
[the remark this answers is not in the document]
mutouzdl: It should work like this(It works fine on my browser):
image

Something must be wrong, I'ill check it.

Thanks for your review!
I'll refactor the code, because I'am trying to change Rules to configurable, similar to Ant Design React:
image


So I changed PR status to Word In Process.
If possible, please help me review it again after I finish it.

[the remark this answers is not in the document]
anddrzejb: Oh, I missed that somehow...sorry to bother you. With demo I used the generated preview. I will test it again once you finish.

mutouzdl:
@anddrzejb I have modified it. If you are free, please help me review it (I have modified the description of PR).
I tested the Edge browser and Chrome browser and it worked fine.(used the generated preview)

components/form/Validate/Rule/Rule.cs
```diff
@@ -0,0 +1,40 @@
...
+        Done: min	            必须设置 type：string 类型为字符串最小长度；number 类型时为最小值；array 类型时为数组最小长度	number
+        Done: pattern	        正则表达式匹配	                                                                        RegExp
+        Done: required	        是否为必选字段	                                                                        boolean
+        Done: transform	        将字段值转换成目标值后进行校验	                                                            (value) => any
+        Done: type	            类型，常见有 string |number |boolean |url | email。更多请参考此处	                        string
+        TODO: validateTrigger	设置触发验证时机，必须是 Form.Item 的 validateTrigger 的子集	                            string | string[]
+        Done: validator	        自定义校验，接收 Promise 作为返回值。示例参考	                                            (rule, value) => Promise
+        would not support: whitespace	    如果字段仅包含空格则校验不通过	                                                            boolean
+     */
+    public class Rule
```
ElderJames, line 25: I think it would be better to name it `FormValidationRule`.

mutouzdl: I've renamed it

components/form/Internal/Validate/Rule/RuleValidateHelper.cs
```diff
@@ -0,0 +1,407 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace AntDesign.Internal
+{
+    internal class RuleValidateHelper
```
anddrzejb, line 6: I cannot find any non-static methods, so probably the class should also be static?

components/form/Internal/Validate/Rule/RuleValidateHelper.cs
```diff
@@ -0,0 +1,407 @@
...
+                Console.WriteLine($"isvalid validationContext.FieldName:{validationContext.FieldName}");
+                return false;
+            }
+
+            result = null;
+
+            return true;
+        }
+
+        private static string GetMinValueString(Type type)
```
anddrzejb, line 373: I cannot find any references to this method. Seems like it is not used.

components/form/Internal/Validate/Rule/RuleValidateHelper.cs
```diff
@@ -0,0 +1,407 @@
...
+            if (type == typeof(sbyte)) return sbyte.MinValue.ToString();
+            if (type == typeof(ushort)) return ushort.MinValue.ToString();
+            if (type == typeof(uint)) return uint.MinValue.ToString();
+            if (type == typeof(ulong)) return ulong.MinValue.ToString();
+            if (type == typeof(decimal)) return decimal.MinValue.ToString();
+
+            return null;
+        }
+
+        private static string GetMaxValueString(Type type)
```
anddrzejb, line 390: I cannot find any references to this method. Seems like it is not used.

components/form/FormItem.razor.cs
```diff
@@ -223,15 +254,54 @@ void IFormItem.AddControl<TValue>(AntInputComponentBase<TValue> control)
...
+        {
+            if (Rules == null)
             {
-                _labelCls = $"{_prefixCls}-required";
+                return Array.Empty<ValidationResult>();
             }
+
+            var results = new List<ValidationResult>();
+
+            var propertyInfo = _fieldIdentifier.Model.GetType().GetProperty(_fieldIdentifier.FieldName);
```
anddrzejb, line 273: A micro-optimization, but maybe this `propertyInfo` should be a private field? Then initialize it only once, if used. You are going to avoid reflection in here on every validation.

components/form/FormItem.razor.cs
```diff
@@ -223,15 +254,54 @@ void IFormItem.AddControl<TValue>(AntInputComponentBase<TValue> control)
...
+            var displayName = string.IsNullOrEmpty(Label) ? _fieldIdentifier.FieldName : Label;
+
+            if (propertyInfo != null)
+            {
+                var validateMessages = Form.ValidateMessages ?? ConfigProvider?.Form?.ValidateMessages ?? new ValidateMessages();
+
+                foreach (var rule in Rules)
+                {
+                    var propertyValue = propertyInfo.GetValue(_fieldIdentifier.Model);
```
anddrzejb, line 282: If I understand this correctly, the value cannot really be different for each rule. If that is the case, then `propertyValue` could be obtained before `foreach` loop.

anddrzejb (changes requested):
My first more-involved read-through of your code. I intend to run some tests to see if I understand all/most what is going on in here. So probably I will get back to you with further review.
This looks like a really great feature!

components/form/FormItem.razor.cs
mutouzdl: Thank you for your review. All the situations you mentioned are correct. I have modified the code and adjusted some class names (using "Form" instead of "Rule").

ElderJames (approved):

PR #1530, opened 2021-05-14
fix & feat & docs & tests: Input general fixes

### LINKED ISSUE #1471: Support FocusAsync on "Search" component

I want to focus a Search input every time when user active a web UI.

I think it should be as below, it is the same as normal input as blazor componenet:

```csharp

input.FocusAsync();

```

I think it should be support to custom what will happen when user invoke FocusAsync(), there should be some options:

1. Keep content and select that all
1. Clean content
1. Keep content and move cursor at the end of input
1. Keep content and move cursor at the head of input

### LINKED ISSUE #1479: TextArea does not support readonly
### Describe the bug

TextArea does not allow you to set the html readonly attribute. It actively fights any attempt to work around it using @Attributes. 

We want to do this: 
`<TextArea id="blah-content" @bind-Value="@blah" MinRows="20" MaxRows="20" readonly />`

This works with the Microsoft controls: 
```
<EditForm Model="bogusModel">
<InputTextArea id="blah-content" @bind-Value="@blah" rows="20" cols="20" readonly class="ant-input"/>
</EditForm>
```

It's possible that this also occurs on the input controls.

### Further technical details
- AntDesign Nuget Package version 0.7.3

### LINKED ISSUE #1525: Setting AntDesign.Input component font color via the Style property is invalid
当AntDesign.Input 组件设置AddOnBefore后，通过Style属性设置颜色、字体大小等，不发生作用，去除<AddOnBefore>后正常

When the Input component is set `AddOnBefore`, setting the color, font size, etc. through the Style property does not work. But if remove `AddOnBefore`, it works normally.

image

image

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1471
Fixes #1479
Fixes #1525

### 💡 Background and solution
This is a general update of `Input` components. My goal here was to bring the components to the same functional state as the react version.

Changes:
1. `Input` gets
   - public methods: `Focus()` and `Blur()`
   - new `[Parameter] Bordered` 
   - new `[Parameter] ReadOnly` 
   - new `[Parameter] InputElementSuffixClass` - check out `InputGroup` demo to see the usecase
   - clear icon fixed to look as react
2. `TextArea`
   - new `[Parameter] ShowCount`
   - clear icon fixed to look as react
3. `Search`
   - look matches react version
   - previous look can still be accessed by using new `[Parameter] ClassicSearchIcon`
4. `InputGroup`:
   - override of style for `Compact` version. React uses `display:block` but that generates whitespace ([source](https://css-tricks.com/fighting-the-space-between-inline-block-elements/)). So `flexbox` is used instead
5. `InputPassword`
   - new `[Parameter] ShowPassword` 
   - new `[Parameter] IconRender` - allows to customize the icon. I believe this probably could be done better. Current implementation requires a lot of effort from a consumer. I was unable to do it purely with `RenderFragment`. Becuase of that, I had to expose `ShowPassword` property. Please refer to the demo.
6. Api & Demos update to match react version as closely as possible
7. All public properties & methods got xml comments.
8. Most of the changes have tests. Except for custom icon for `InputPassword`, as I do not know if this is going to be kept. Maybe someone has a better idea?

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  New parameters for `Input`:  `Bordered`, `ReadOnly`, `InputElementSuffixClass` & methods: `Focus()`, `Blur()`, fixed clear icon. <br/> New parameter for `TextArea` `ShowCount`, fixed clear icon. <br/> `Search` gets new look and paramter `ClassicSearchIcon` for fallback to old look. <br/> `InputGroup` whitespace removed. <br/> New parameters for `InputPassword`: `ShowPassword` & `IconRender`.           |
| 🇨🇳 Chinese |           |


components/input/Input.cs
```diff
@@ -286,6 +405,7 @@ private void ToggleClearBtn()
             {
                 builder.OpenComponent<Icon>(31);
                 builder.AddAttribute(32, "Type", "close-circle");
+                builder.AddAttribute(32, "Theme", "fill");
```
ElderJames, line 408: Are you sure you would using the same sequenceId?

components/input/Input.cs
```diff
@@ -500,6 +618,7 @@ protected override void BuildRenderTree(RenderTreeBuilder builder)
                 builder.AddAttribute(60, "placeholder", Placeholder);
                 builder.AddAttribute(61, "value", CurrentValue);
                 builder.AddAttribute(62, "disabled", needsDisabled);
+                builder.AddAttribute(62, "readonly", ReadOnly);
```
ElderJames, line 621: Same sequenceId as well?

components/input/TextArea.razor.cs
```diff
@@ -32,56 +32,85 @@ public partial class TextArea : Input<string>
...
+                if (value >= DEFAULT_MIN_ROWS && value <= MaxRows)
                 {
-                    _maxRows = value;
+                    _minRows = value;
                     AutoSize = true;
                 }
                 else
                 {
-                    throw new ArgumentOutOfRangeException($"Please enter a value between {MinRows} and {uint.MaxValue}");
+                    throw new ArgumentOutOfRangeException(nameof(MinRows), $"Please enter a value between {DEFAULT_MIN_ROWS} and {MaxRows}");
```
ElderJames, line 91: Can we fallback to the minimum or maximum value?

components/input/Input.cs
anddrzejb: Good catch. Obviously not 😄 . After so many changes I stopped seeing obvious mistakes...

components/input/TextArea.razor.cs
anddrzejb: Sure. Probably safer to fall back to minimum? Also, maybe we do a `Console.Debug` here then?

ElderJames: That's fine. And the maximum for `MaxRows`.

components/input/Input.cs
ElderJames: 😂

components/core/Base/AntComponentBase.cs
```diff
@@ -104,7 +104,31 @@ protected async Task JsInvokeAsync(string code, params object[] args)
...
+        /// HTMLInputElement or HTMLTextAreaElement. Otherwise will only focus.
+        /// </summary>
+        /// <param name="target">Element that will receive focus.</param>
+        /// <param name="behavior">Behavior of focused element</param>
+        /// <param name="preventScroll">Whether to scroll to focused element</param>
+        protected async Task FocusAsync(ElementReference target, FocusBehavior behavior, bool preventScroll = false)
+            => await Js.FocusAsync(target, behavior, preventScroll);
+
+        /// <summary>
+        /// Standard Blur. Uses JS introp.
```
ElderJames, line 127:
```suggestion
        /// Standard Blur. Uses JS interop.
```

ElderJames:
I notice that there are missing a showcase with select addons.

image

ElderJames:
The format tooltip demo is not correct.
image

ElderJames:
The icon would make input password blur.
image

anddrzejb:
@ElderJames Please have a look at commit 32e904e. 
I did not really understand the issue with the `Tooltip`. I tried on my machine, and both wasm & server-side were working. I did some changes, to work almost like react version. The only difference is that if you type `0.000` only `0` will be shown, at least until you type `0.0001`. But this is coming from `Convert` and I don't think it is worth it to spend more time on this - the idea is presented, and that should be enough I guess?
Please have a look at the `InputPassword` focus issue solution. This is like the most crappy solution I have submitted, but honestly, I don't know how else to solve it. I think JS must be doing its own thing, but only in this weird situation. Do you have any suggestions there? This problem is only in wasm, server-side works as expected.

ElderJames:
Still have a binding issue with the clear all button.

ant-input-issue

anddrzejb:
@ElderJames I am away now, so I will look at this as soon as I come back.

ElderJames:
Welcome back! I found the clear button still has an issue. 

input-issue

anddrzejb:
Yeah, I noticed that. If you set `DebounceMiliseconds` to 0 then this problem does not appear. Or if you press a key after debouce period passes. I think we probably should ignore `DebounceMiliseconds` in wasm (or at least default it to 0). What do you think?

ElderJames:
I don't think so 😂
We need `DebounceMiliseconds` to delay the model binding, but can ignore it with the showing of the clear button.

anddrzejb:
@ElderJames Thanks for pushing me. You were right.

ElderJames:
@anddrzejb Good suggestion. We can try to fix it.

ElderJames (approved):
LGTM. Thank you for your time and hard work, it has improved us a lot.

PR #1541, opened 2021-05-16
feat(module: select): support simple data source

### LINKED ISSUE #1454: (Feature) Select: Support for not using ValueName and LabelName with data source
`ValueName` and `LabelName` are required when work with data source. As a result, we can not use collection of `TItem` directly as the data source, a collection of wrapper class is required. For example, `List<string>` can not be used as data source.

If 'ValueName' is not assigned or empty, should use `TItem` instance itself as the item value. In this case, TItem and TItemValue should be same type.

If 'LabelName' is not assigned or empty, should use the result of `TItem.ToString()` as the label text.

anranruye:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

fixes #1454
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | When the item in the data source and the value property of select use the same type, it is not necessary to specify ValueName; When LabelName is not specified, the return value of the ToString() method of the item in the data source is used as the label |
| 🇨🇳 Chinese | 当数据源中的项和Select的Value属性使用相同类型时，无需指定ValueName；当不指定LabelName时，将使用数据源中的项的ToString()方法的返回值作为Label |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -511,6 +511,11 @@ namespace AntDesign
         internal bool IsDropdownShown() => _dropDown.IsOverlayShow();
         protected override void OnInitialized()
         {
+            if (typeof(TItemValue) != typeof(TItem) && string.IsNullOrWhiteSpace(ValueName))
+            {
+                throw new ArgumentNullException(nameof(ValueName));
+            }
+
             SetClassMap();
 
             if (string.IsNullOrWhiteSpace(Style))
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -85,10 +85,10 @@ namespace AntDesign
             get => _labelName;
             set
             {
-                _getLabel = PathHelper.GetDelegate<TItem, string>(value);
+                _getLabel = string.IsNullOrWhiteSpace(value) ? null : PathHelper.GetDelegate<TItem, string>(value);
                 if (SelectMode == SelectMode.Tags)
                 {
-                    _setLabel = PathHelper.SetDelegate<TItem, string>(value);
+                    _setLabel = string.IsNullOrWhiteSpace(value) ? null : PathHelper.SetDelegate<TItem, string>(value);
                 }
                 _labelName = value;
             }
@@ -179,8 +179,8 @@ namespace AntDesign
             get => _valueName;
             set
             {
-                _getValue = PathHelper.GetDelegate<TItem, TItemValue>(value);
-                _setValue = PathHelper.SetDelegate<TItem, TItemValue>(value);
+                _getValue = string.IsNullOrWhiteSpace(value) ? null : PathHelper.GetDelegate<TItem, TItemValue>(value);
+                _setValue = string.IsNullOrWhiteSpace(value) ? null : PathHelper.SetDelegate<TItem, TItemValue>(value);
                 _valueName = value;
             }
         }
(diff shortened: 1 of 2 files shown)


components/select/Select.razor.cs
```diff
@@ -490,6 +490,13 @@ internal bool IsGroupingEnabled
 
         #endregion Properties
 
+        private static readonly bool _isItemAndValueSameType;
+
+        static Select()
```
anddrzejb, line 495: Why is this static?

components/select/Select.razor.cs
```diff
@@ -637,7 +643,7 @@ private void CreateDeleteSelectOptions()
 
             foreach (var item in _datasource)
             {
-                TItemValue value = _getValue(item);
+                TItemValue value = _getValue == null ? THelper.ChangeType<TItemValue>(item) : _getValue(item);
```
anddrzejb, line 646: @Zonciu Could you look at this? Could this be incorporated somehow into expression tree idea? Maybe it is not worth it, or too complex to actually invest time in it.

anddrzejb (commented):
Sorry for my delay. I am away now and cannot really spend much time. These here are my first thoughts after looking at you code.

components/select/Select.razor.cs
anranruye: Of course this can be static. All instances of `Select<T1, T1>` will share the same `_isItemAndValueSameType`. But instances of `Select<T2, T2>` will use another `_isItemAndValueSameType`. And for generic class static fields, the runtime will distribute memory when first time use the specific generic class .

anranruye: Of course this can be static. All instances of `Select<T1, T1>` will share the same `_isItemAndValueSameType`. But instances of `Select<T2, T2>` will use another `_isItemAndValueSameType`. And for generic class static fields, the runtime will distribute memory when first time use the specific generic class .

components/select/Select.razor.cs
anranruye: I also thought about this. I think it's nice to get `TItem x => (TItemValue)x` for string.Empty and whitespace. But maybe it's better to return null or throw an exception for null.

components/select/Select.razor.cs
anranruye: Now I want to just remove this judgement for `!_isItemAndValueSameType && string.IsNullOrWhiteSpace(ValueName)`. As you said, it is possible to use expression tree to build the delegate. So `null` or `string.Empty` or whitespace should be valid value for `ValueName` .

anddrzejb (approved):
LGTM

Zonciu:
The semantics of `delegate == null ` like `setValue == null` is unclear. There is a field `_isPrimitive` is similar to this PR, I think we can reuse it.

My suggestion: Wrap get/set value/label into four private methods, then check `_isPrimitive` to use delegate or use the item directly. When `_isPrimitive` is false, `getValue`, `setValue`, `getLabel`, `setLabel` must not be null, so we can check it in `GetValue`, `SetValue`, `GetLabel`, `SetLabel` method and throw our exceptions to remind users to set `LabelName` and `ValueName`, instead of unclear exception message: `InvalidCastException: Object must implement IConvertible`.

ElderJames (approved):

ElderJames (approved):

PR #1571, opened 2021-05-25
fix(module:table): improve build-in filter

YMohd:
### 🤔 This is a ...

- [ ] New feature
- [X ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [X ] Performance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

1-reduce number of table reload , this change will stop calling reload on overlay MouseUp event which is not required
2-Fix issue when change filter condition (And\Or) where it is close automatically. 
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     X      |
| 🇨🇳 Chinese |           |

- [X ] Doc is updated/provided or not needed
- [ X] Demo is updated/provided or not needed


ElderJames:
I think we should keep the behavior that clicking on the outside of the filter will make it close and submit automatically. This is the behavior in Ant Design React, and it makes sense to me.

YMohd:
OK agreed, I revert to the previous behavior and now this request include 2 things:

1. Fix issue of closing drop down when condition(And/Or) is changed.
2. trigger onMaskClick event only when the overlay is visible.

components/core/Component/Overlay/OverlayTrigger.razor.cs
```diff
@@ -338,7 +338,7 @@ protected virtual async Task OnTriggerContextmenu(MouseEventArgs args)
 
         protected virtual void OnMouseUp(JsonElement element)
         {
-            if (_mouseInOverlay == false && _mouseInTrigger == false)
+            if (_mouseInOverlay == false && _mouseInTrigger == false && Visible)
```
anranruye, line 341: If you use `Visible` as a condition, `DatePicker` component may be affected so that the date selector can not be closed.

ElderJames:
Please fix the conflics.

PR #1588, opened 2021-06-01
fix(module:dropdownbutton): align with react

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Current behaviour is this:
dropdownButton-issues

Expected is this:
dropdownButton-expected
1. Hovering over left button should not open the dropdown.
2. When loading, dropdown should be disabled.
3. Overlay should be position to the right edge of the right button.

### 💡 Background and solution
1. The issue here was that overlay trigger was set on the wrapping div instead of on the right button. 
2. Exposed `Loading` parameter. This is different than in react, but actually is the same as in a number of other components.
3. Hardcoded the position of the overlay in `DropdownButton`

Other changes:
1. Fixed docs for the `Button` in area of `Multiple` and fixed styling for block buttons.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Added `Loading` parameter to `DropdownButton`         |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/Component/Overlay/OverlayTrigger.razor.cs
+++ b/components/core/Component/Overlay/OverlayTrigger.razor.cs
@@ -139,7 +139,7 @@ namespace AntDesign.Internal
         }
 
         [Inject]
-        private DomEventService DomEventService { get; set; }
+        protected DomEventService DomEventService { get; set; }
 
         private bool _mouseInTrigger = false;
         private bool _mouseInOverlay = false;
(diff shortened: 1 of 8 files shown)


ElderJames (approved):
LGTM

Issue #1495, opened 2021-05-09
(Epic) Add xml comments for the whole library


anranruye:
### Is your feature request related to a problem? 

Without xml comments,  it's hard to get enough information from the library metadata.

anddrzejb:
We should at least try to add to every `[Parameter]` & public method.

knight1219:
Would we want to pull the documentation from what is in the API section of the current docs for each property?

![image](https://user-images.githubusercontent.com/6771697/137592543-0b647140-6ba7-4149-9a22-b6f583920db8.png)


It would be interesting if once the XML docs are updated that API section could be generated _from_ the XML docs so they stay in sync with each other.

ElderJames:
@knight1219  Yes, we wish we could do that. I would also like to support English and Chinese. Do you have any ideas, please?

anddrzejb:
Cool! But we are going to loose all the xml attributes (like `<see cref=... `), aren't we?

knight1219:
Depends on what you mean: if you mean references to objects, I don't think so. Those should still work. If you mean links to the docs site, we will need to do some find and replace magic to get the correct language in the route

knight1219:
It seems that this isn't a built in way to do localized comments. The roslyn team has had this as an open issue since 2015 with no real signs of being adopted by them. [https://github.com/dotnet/roslyn/issues/3371](Source:dotnet/roslyn). 

I'm still trying to think of the best way to do this so we can support multiple languages, but this is what I've found so far.

knight1219:
Wanted to give an update on this. I have found a [library ](https://github.com/devlooped/NuDoq) that allows your to read and manipulate the XML document file and I am testing it out. The idea is that inside the documentation sections your will add a new `<localize>` node in the XML with the appropriate `lang` attribute:

```
        /// <summary>
        /// <localize lang="en-US">
        /// Offset from the bottom of the viewport (in pixels)
        /// </localize>
        /// <localize lang="zh-CN">
        /// 从视口底部偏移（以像素为单位）
        /// </localize>
        /// </summary>
```

We can then use the NuDoc library to parse out and create as many doc files as we need for the different languages that are part of the documentation. I'm still working on this and hope to have something testable soon. Seeing as this is .net code I'm running, how do we want to get this into the build process? Just a console app that runs and generates everything? I'm open to ideas.

knight1219:
[...]

PR #584, opened 2020-09-09
refactor(module: select): add overlay, fix some serious bugs

mutouzdl:
[[中文版模板 / Chinese template](https://github.com/ant-design-blazor/ant-design-blazor/blob/master/.github/PULL_REQUEST_TEMPLATE/pr_cn.md)]

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |      add overlay, fix some serious bugs     |
| 🇨🇳 Chinese |     将下拉弹出功能改为使用overlay，修复部分严重的bug    |

### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #1609, opened 2021-06-06
fix: DropdownButton missing properties & docs fix for Dropdown & Button

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
`DropdownButton` is missing `Button` standard behaviours like `Ghost`, `Danger`, etc.

### 💡 Background and solution
This is a follow-up to my PR #1588.
You will see a lot of changes here, because I was also try to follow the suggestion in #1495. 
So this PR adds:
1. All missing functionality of a regular `Button` to `DropdownButton`:
   - `Block`
   - `Danger`
   - `Ghost`
   - `Loading`
   - `Type` - was already existing but this PR adds ability to set type separately for left and right button
   - `ButtonsRender` - I decided to hide this in `Dropdown` by making it `internal` and expose it only in `DropdownButton` - since this is strictly related to `DropdownButton`, I do not think it should be visible from regular `Dropdown`.
2. Adds new demos and API docs
3. Adds new button type - `text` to `Button` component
4. Fixes the `Button` demos to include `text` button type
5. Adds API docs to the `Button` component

My left-out issues:
1. In `Overlay` there is a property `HiddenMode` - I could not find any real used for it. I tried using it but I failed to see any result. It was added in PR #584. If you find time @mutouzdl, can you let me know the purpose of this parameter? 
2.  In `Button` there is the paramter `Search`. It looks to me that its only purpose is to add a css class `ant-input-search-button`. That css class seems to be used only `Search` component, that anyway sets it manually. It seems to me it is actually useless here, but I left it, so maybe @ElderJames could let me know what other purposes it may server (it was added by PR #82). 

Question:
How can we embed a style in docs in similar fashion as it is done in antd? For example, if you got to [antd button demo](https://ant.design/components/button/), you will notice that any button has also class 
```css
[id^=components-button-demo-] .ant-btn {
  margin-right: 8px;
  margin-bottom: 12px;
}
```
(and other) that are giving proper margins, so when the components are wrapped, they look much better.
### 📝 Changelog
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add typical `Button` propertied to `DropdownButton`. Include demo & API docs for `Dropdown` API and `Button` API.         |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/Component/Overlay/Overlay.razor.cs
+++ b/components/core/Component/Overlay/Overlay.razor.cs
@@ -54,6 +54,14 @@ namespace AntDesign.Internal
         [Parameter]
         public int HorizontalOffset { get; set; } = 4;
 
+        /// <summary>
+        /// By default Overlay does not render its content if Overlay hasn't been
+        /// activated (shown at least once). Setting HiddenMode = true will 
+        /// go through rendering process.
+        /// Use case: Select component, when using <SimpleOption> or <SelectOption>
+        /// needs HiddenMode = true, so the select options are initialized and 
+        /// potential defaults can be rendered properly.
+        /// </summary>
         [Parameter]
         public bool HiddenMode { get; set; } = false;
 
--- a/components/button/Button.razor.cs
+++ b/components/button/Button.razor.cs
@@ -7,42 +7,6 @@ namespace AntDesign
 {
     public partial class Button : AntDomComponentBase
     {
-        [Parameter]
-        public bool Block { get; set; } = false;
-
-        [Parameter]
-        public bool Ghost { get; set; } = false;
-
-        [Parameter]
-        public bool Search { get; set; } = false;
-
-        [Parameter]
-        public bool Loading
-        {
-            get => _loading;
-            set
-            {
-                if (_loading != value)
-                {
-                    _loading = value;
-                    UpdateIconDisplay(_loading);
-                }
-            }
-        }
-
-        [Parameter]
-        public string Type { get; set; } = ButtonType.Default;
-
-        [Parameter]
-        public string HtmlType { get; set; } = "button";
-
-        [Parameter]
-        public string Shape { get; set; } = null;
-
-        private bool _animating = false;
-
-        private string _btnWave = "--antd-wave-shadow-color: rgb(255, 120, 117);";
-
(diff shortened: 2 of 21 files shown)


mutouzdl:
Select component should render the overlay children on first time(but should not show). So "HiddenMode" is used to do that.

You may try to set "HiddenMode" to false in Select component's "OverlayTrigger", and check the "Simple encapsulation" demo.You should find someting wrong on it.

anddrzejb:
@mutouzdl Thank you for the explanation. It seems this is needed only when `RenderFragment` is used to fill up the select. I will try to give it an appropriate description.

@ElderJames Something is wrong with codecov. It caused the build to fail for at least couple of last PRs.

ElderJames:
> How can we embed a style in docs in similar fashion as it is done in antd? For example, if you got to antd button demo, you will notice that any button has also class

@anddrzejb  You can add the <style> tag in doc/*.md files

anddrzejb:
@ElderJames This I know. It is exactly what I did for button block demo. I was hoping for a more unified solution and less verbose in the demo section, but if `<style>` section is the recommended way, then `<style>` it is.
Any advice regarding the `Search` parameter in `Button` component?

components/core/Component/Overlay/Overlay.razor.cs
```diff
@@ -54,6 +54,14 @@ public sealed partial class Overlay : AntDomComponentBase
         [Parameter]
         public int HorizontalOffset { get; set; } = 4;
 
+        /// <summary>
```
anddrzejb, line 57: @mutouzdl for your consideration.

mutouzdl: I think that's a pretty clear explanation

ElderJames:
I noticed that we have added `ant-input-search-button` to the button in the search component,  so we can delete the Search parameter.

ElderJames (approved):

Issue #1590, opened 2021-06-03
DatePicker cannot input using keyboard


shirleen1234:
### Describe the bug

The date doesn't change when input by using keyboard. but it works by selecting date from the picker using mouse.

### Steps to reproduce (please include code)

```
<FormItem Label="Birth Date">
  <DatePicker @bind-Value="@context.BirthDate" DefaultValue="DateTime.Now" Format="yyyy-MM-dd" />
</FormItem>
```

### Further technical details
- AntDesign Nuget Package version 0.8.1

anranruye:
@shirleen1234  Thanks for contacting us.

Do you meet the same problem described in #1568?

shirleen1234:
I'm using the same format (yyyy-MM-dd) for both displaying and inputting.
I found that @bind-Value might be the problem as `<DatePicker TValue="DateTime?" DefaultValue="new DateTime(2015, 1, 1)" Format="yyyy/MM/dd" />` (from AntDesign docs) does work.

anranruye:
Confirmed this is a bug. @anddrzejb Do you have any idea why this happens?

anddrzejb:
Unfortunately I will be able to have a look at this on Monday. Currently I am away. Sorry...

anranruye:
Also found that this only happens when the DatePicker is inside a Form.

PR #1617, opened 2021-06-08
fix(module:datepicker): in form behavior

### LINKED ISSUE #1615: DatePickerType.Year ErrorMessege
### Describe the bug
When i select year on YearPicker, i get error messege 
image

image
Model
image

### Further technical details

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1615 

### 💡 Background and solution
Both pickers had a method `UpdateCurrentValueAsString` which was called on data change and was only doing something if the component was attached to a form. The problem with it is that it was trying to apply the string value of the input to `CurrentValue`. But that code is a general code that sits in `AntInputComponentBase`. It does not handle well the datepicker specific data format. So it was failing a lot and was resetting also the data. I decided to remove it. 
Also, I noticed that for `RangePicker`, when in `Form`, `EditContext.NotifyFieldChanged` was never called. In regular `DatePicker` this call is done when `CurrentValue` is updated. But in `RangePicker` `CurrentValue` is never updated as a whole, only array items are. So I added the appropriate code. Also, maybe it would be a good idea to provide `Required` attribute for `RangePicker`. Regular `Required` will not work, because the array with `DateTime?` will exist and meet the built-in attribute, even when its items are nulls. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/DatePicker.Razor.cs
+++ b/components/date-picker/DatePicker.Razor.cs
@@ -99,7 +99,7 @@ namespace AntDesign
             }
             if (FormatAnalyzer.TryPickerStringConvert(args.Value.ToString(), out TValue changeValue, IsNullable))
             {
-                Value = changeValue;
+                CurrentValue = changeValue;
                 GetIfNotNull(changeValue, (notNullValue) =>
                 {
                     PickerValues[0] = notNullValue;
@@ -107,8 +107,6 @@ namespace AntDesign
 
                 StateHasChanged();
             }
-
-            UpdateCurrentValueAsString();
         }
 
         protected override Task OnBlur(int index)
@@ -244,8 +242,6 @@ namespace AntDesign
             }
 
             _pickerStatus[0]._hadSelectValue = true;
-
-            UpdateCurrentValueAsString();
 
             if (!IsShowTime && Picker != DatePickerType.Time)
             {
--- a/components/date-picker/RangePicker.razor.cs
+++ b/components/date-picker/RangePicker.razor.cs
@@ -149,11 +149,13 @@ namespace AntDesign
 
                 ChangePickerValue(changeValue, index);
 
+                if (_isNotifyFieldChanged && (Form?.ValidateOnChange == true))
+                {
+                    EditContext?.NotifyFieldChanged(FieldIdentifier);
+                }
 
                 StateHasChanged();
             }
-
-            UpdateCurrentValueAsString();
         }
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


anddrzejb:
@anranruye This PR may also fix #1590, but honestly, I cannot recreate that issue on my machine (neither server-side nor wasm). Do you think it would be possible for you to get this PR and check if the problem still persists?
On my machine it looks like this:
datepicker-form-blur

anranruye:
@anddrzejb  
I didn't check this pr yet. To repro #1590, you should press enter after inputing. I can repro #1590 by this demo: https://ant-design-blazor.gitee.io/en-US/components/form#components-form-demo-size. For both a regular datepicker(not in a form) and an in-form datepicker, the value doesn't get updated when just clicking outside the component(OnBlur), not sure whether this is also a bug. Pressing `arrow up` or `tab` is ok. But pressing `enter` is not working for an in-form datepicker. In the gif you provided, it seems this issue is already solved.

anddrzejb:
When value is changed using keyboard but not confirmed by pressing tab/enter key, then clicking outside of the `DatePicker` (blur) will cause the `DatePicker` to revert to previous value. This is by design and follows react antD.

I also checked again on the docs generated with this PR and indeed it seems #1590 is fixed by this change as well. @ElderJames could you add it to the linked issues? My permissions do no allow me to add it.

anranruye (approved):

ElderJames (approved):

anranruye:
@anddrzejb @ElderJames  I just found that there is an issue when blur. For both datepickers inside and outside a form, after input a valid date string, the bound values are updated immediately. If we click somewhere outside the datepicker at this time, the datepicker ui value changes to the last valid value, but the bound value doesn't change.
image
image

anddrzejb:
@anranruye Thanks for tracking that down. I do not think it is related to this PR, but nevertheless it needs to be fixed ASAP. I will try to have a look at this.

anranruye:
@anddrzejb  I think it is caused by the new behavior introduced in this pr. Anyway, this doesn't matter.

datnguyen89:
Dear team!
How to apply this change of this pull request?  Should I re-install Ant Blazor or wait version 0.8.2. I'm sorry cause may be it's idiot question

anranruye:
@datnguyen89 You can use nightly-build version which contains the latest changes. Unfortunately, the nightly-build mechanism crashed, so there is no nightly-build versions these days. We will fix it as soon as possible. If you can't wait, you can download the source code and build the project in your own machine. Node.js is not necessary, it is used for generating css and js file, you can remove the Gulp tasks from the project file and get the css and js files from a release version.

ElderJames:
Nightly build has been fixed. Please follow the guides https://antblazor.com/docs/nightly-build

PR #1658, opened 2021-06-19
Update doc for `Select`

Banyc:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

Not everyone understands Chinese. Enhance development experience.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Update doc for `Select` |
| 🇨🇳 Chinese |           |

### DIFF
--- a/site/AntDesign.Docs/Demos/Components/Select/doc/index.en-US.md
+++ b/site/AntDesign.Docs/Demos/Components/Select/doc/index.en-US.md
@@ -14,8 +14,7 @@ Select component to select value from options.
 
 ## API
 
-
-### Select props
+### Select Properties
 
 | Property | Description | Type | Default | Version |
 | --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


anddrzejb (changes requested):
As far as I know, the idea was to create documentation as close to the [original AntD](https://ant.design/components/select/#API). Of course it is not always possible, but whenever possible, we try to do it. Also, this approach allows us to actually reuse both translations. Could you possibly align your docs with the original (when possible) and also add Chinese version using original as a source and English for missing parts?

Banyc:
@anddrzejb In this commit, I have translated the English doc based on the corresponding Chinese part in this doc. Was not realizing the coordination between the two projects. XD

anddrzejb:
@Banycm I may be wrong in here, but this is the approach I have been employing since I started contributing to this project and it was never questioned. When in doubt, I always ask the creator & owner of the repository. However, since I have been referred to original multiple times, I do believe this is the correct approach.

PR #1681, opened 2021-06-27
fix(module: datepicker) Date picker blur/focus & regression tests

### LINKED ISSUE #1627: RangePicker: RangePickers are always highlight after set the value
### Describe the bug

image

### Further technical details
- AntDesign Nuget Package version: 0.8.1

### LINKED ISSUE #1630: DatePicker - blur persisting value but not in the picker
### Describe the bug
Coming from PR #1617, credited to @anranruye:

> I just found that there is an issue when blur. For both datepickers inside and outside a form, after input a valid date string, the bound values are updated immediately. If we click somewhere outside the datepicker at this time, the datepicker ui value changes to the last valid value, but the bound value doesn't change.
> image
> -------then blur--------
> image

I added this issue, so it can be easily tracked.

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1627 
Fixes #1630 
Fixes incorrect `CultureInfo` being picked up - for example when `en-GB` culture was set, the `CultureInfo` was `en` which was defaulting to `en-US`. 
Fixes blur when clicked on non-functional panel part:
datePickerBlurOnNonFunctionalPanel

### 💡 Background and solution
I could not make all the tests I wanted for the date picker, because there are some issues with multiple re-renderings coming from `Overlay` and `bUnit` has a problem with picking them up. Instead of forcing the `bUnit` to work with the changes, I intend to actually fix the `Overlay` first. 

### 📝 Changelog
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/DatePicker.Razor.cs
+++ b/components/date-picker/DatePicker.Razor.cs
@@ -34,6 +34,10 @@ namespace AntDesign
                 else if ((IsNullable && DefaultValue != null) || (!IsNullable && !DefaultValue.Equals(default(TValue))))
                 {
                     DefaultPickerValue = DefaultValue;
+                    if (!DefaultValue.Equals(Value))
+                    {
+                        CurrentValue = DefaultValue;
+                    }
                 }
                 else if (!IsNullable && Value.Equals(default(TValue)))
                 {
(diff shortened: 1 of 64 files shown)


ElderJames (approved):
Thank you!

Issue #1655, opened 2021-06-19
The 'Upload' component will treat the 201 (created) HTTP status code as a failure result


hongyeli620:
### Describe the bug

The 'Upload' component will treat the 201 (created) HTTP status code as a failure result

#### Page Code:
```xml
<Upload Action="@uploadUrl" Name="files" FileList="_fileList"
        OnSingleCompleted="OnSingleCompletedAsync" BeforeAllUploadAsync="OnBeforeUploadAsync">
    <Button>导入</Button>
</Upload>
```
#### C# Code:
When the HTTP status code returns **201 (created)**, the **fileinfo.File.State** property returns **Uploadstate.Fail**, which makes it impossible to get the actual return result through **fileinfo.File.Response**
```csharp
async Task OnSingleCompletedAsync(UploadInfo fileinfo)
{
    if (fileinfo.File.State == UploadState.Success)
    {
        var jObject = JObject.Parse(fileinfo.File.Response);

        var result = fileinfo.File.GetResponse<UploadResult>();
        await this.ImportAsync(Path.GetFileName(result.StoredFileName));
    }
}
```

To solve this problem, you can only replace the **CreatedResult** with the **JsonResult** in the file save controller
```csharp
[HttpPost]
public async Task<ActionResult<IList<UploadResult>>> PostFile(
            [FromForm] IEnumerable<IFormFile> files)
{
    //...
    //return new CreatedResult(resourcePath, uploadResults);
    return new JsonResult(uploadResults);
}
```

### Further technical details
- AntDesign Nuget Package version: 0.8.1
- Dotnet 5.0
- The IDE : VS 2019
- Blazor Server

ElderJames:
Thanks for contacting us @hongyeli620 

You're right that we have a restriction in our component that only 200 is marked as success, but I think we can support all status codes that indicate success.

https://github.com/ant-design-blazor/ant-design-blazor/blob/2ac444bb241d8a8cbd66c5775a9fc2aa1bb83059/components/core/JsInterop/interop.ts#L128

Would you be interested in helping to revise it?

henrikwidlund:
I can do a PR for this

ElderJames:
@henrikwidlund Cool, thank you!

PR #1705, opened 2021-07-04
fix(module: upload): treat 2xx response codes success

henrikwidlund:
Treat all 2xx status codes as success in the upload module.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#1655 (The 'Upload' component will treat the 201 (created) HTTP status code as a failure result)

### 💡 Background and solution
Depending on how the server receiving to the uploaded file is configured, successful requests might not be responded with 200 as a status code. Fix is to be less strict and treat any 2xx status code as successful.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Treat all 2xx status codes in responses in the upload module as successful |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/JsInterop/interop.ts
+++ b/components/core/JsInterop/interop.ts
@@ -125,7 +125,8 @@ export function uploadFile(element, index, data, headers, fileId, url, name, ins
   const req = new XMLHttpRequest()
   req.onreadystatechange = function () {
     if (req.readyState === 4) {
-      if (req.status != 200) {
+      // #1655 Any 2xx response code is okay
+      if (req.status < 200 || req.status > 299) {
         instance.invokeMethodAsync(errorMethod, fileId, `{"status": ${req.status}}`);
         return;
       }


components/core/JsInterop/interop.ts
```diff
@@ -125,7 +125,8 @@ export function uploadFile(element, index, data, headers, fileId, url, name, ins
   const req = new XMLHttpRequest()
   req.onreadystatechange = function () {
     if (req.readyState === 4) {
-      if (req.status != 200) {
+      // #1655 Any 2xx response code is okay
+      if (req.status < 200 || req.status > 299) {
```
ElderJames, line 129: It should be less than 300.

henrikwidlund: You mean to say that 1xx is fine too? Because the code i modified would trigger error method only if status is less than 200 or greater than 299.

ElderJames: Ok, you are right.

ElderJames (approved):
LGTM

Issue #1557, opened 2021-05-19
<Select> Very slow


mrcameron999:
### Describe the bug
Select component takes a very long time to open and feels very unresponsive

### Steps to reproduce (please include code)
Create a multi-select component and then publish the site
When you click on the select section it takes a good second or two to open. I don't get this issue on localhost but only on hosted sites. Is it to do with the latency and is there a way I can speed this up?

anddrzejb:
I have several questions:
1. Wasm or server-side?
2. What is the size of the datasource (how many items)? If you limit the size of the datasource, do you see any improvement?
3. Does this happen on every select open or only on first time opening?
  - if it happens on every select open, do you see a difference between first time open and any consecutive open?

mrcameron999:
> I have several questions:
> 
> 1. Wasm or server-side?
> 2. What is the size of the data source (how many items)? If you limit the size of the data source, do you see any improvement?
> 3. Does this happen on every select open or only on the first time opening?
> 
> * if it happens on every select open, do you see a difference between first time open and any consecutive open?


- I'm using Blazor Server
- The data source is no more than 4 items.
- Happens with every click

Here is a gif showing -I am clicking as soon as the cursor is over the box
Host: https://gyazo.com/90164d30244c6d3eba808d04691cf830

Local Host: https://gyazo.com/04a77d69aed78627726ed435b250bb76

Thanks for your help

anddrzejb:
Well...I can see the problem. In my app it takes roughly 2sec to show the overlay regardless if I have 5 items or 450. So I do not think this is related to the number of items. If I run the same from localhost, then it is much faster (0.5sec?). This might be related to multiple re-renderings on open, but I am guessing now. I will try to have a look at this, but I do not have any time frame for this (currently overloaded with work).

mrcameron999:
I have done some more experimenting and as I am using Cloudflare this is increased. It appears to me that when you click on the select it's going to the server and then a response is given back. What I'm confused about and forgive me if I am wrong I'm new to blazer surely the select could be stored in the user's browser instead of going to the server to ask to open the select?
Again I might be wrong and missing how blazer works when it comes to components like these

anddrzejb:
[...]

PR #1711, opened 2021-07-05
fix(module:interop): reorganization into classes

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
I tried here to reorganize the javascript into modules/classes. My approach was mostly to use static classes instead of group of functions in order to organize them into more coherent groups. The naming is arbitrary and of course should be discussed. 
I removed some of the functions without any references or where I found a functional duplication. In obvious places I added types (so to actually use the TypeScript main feature). I haven't really touched `Caret` & `Dragable` classes, except moving them to different folders. I noticed that there are still a lot of functions that could be improved, but I figured this is not the scope of this PR.
I also added a singleton class with js state - in hope it will be used as such. So if in the future there is a need to add a structure to hold some js specific data - the state should be the place to hold it. It is already used in that capacity (keeps dictionaries). 
Also, `JSInteropConstants` was changed - the structure of `*.ts` files is mostly reflected in static helper classes. Because I did not want to change all components, that is why I left the calls in `JSInteropConstants` with redirects to respective static class. I guess, with time, we could replace them with direct calls to static helpers. But also, I remember that @ElderJames mentioned something about moving the calls to extension methods? 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames:
This is great! I have two more thoughts on this part.
1. For . NET 5 and later versions, use JS isolation to import  JS modules(compiled from ts), and also support <script>.
2. Make all interop calls as extension methods of `IJsRuntime` or modular services. Then users can use the interop methods easily.

ElderJames:
This is great! I have two more thoughts on this part.
1. For . NET 5 and later versions, use JS isolation to import  JS modules(compiled from ts), and also support <script>.
2. Make all interop calls as extension methods of `IJsRuntime` or modular services. Then users can use the interop methods easily.

anddrzejb:
1. I have to do more investigation, but I think that only component-related js should be imported through isolation functionality. I mean, common js (like `infoHelper.ts` or `manipulationHelper.ts`) probably should be always loaded regardless. 
2. Do you want to make the `IJsRuntime` extension methods part of this PR? If yes, should I try to apply similar grouping as in current `JSInteropConstants`?

Unfortunately, I have very little time now to work on `AntBlazor` 😞 . If at all, I can usually spend an hour, maybe 2 a day working on it. It would probably be easier for me if this PR was merged. Then I can create another PR following through with your requests. Otherwise, I may be faced with loosing a day or 2 trying to merge any upcoming changes to `interop`. All the work here is meant to lead me to overhaul of overlay in order to fix the issue I mentioned in #1557 (and further in making proper tests for `DatePicker`).

ElderJames:
Thanks for your time @anddrzejb , and of course you can create another PR for them. They are not very emergency.

anddrzejb:
@ElderJames My latest couple of commits are adding a new project that contains tests for typescript. Currently it has only 3 tests - a proof of concept. I handle there basic tests including also samples on how to test dom related functions/elements. 
Another change here is that there is new gulp task that except bundled js, will also create separated js files (matching 1-to-1 typescript file structure). This is probably the first step to requested js isolation. 
Current testing approach is not my favourite. My issue with it is that I cannot run the tests in VS test runner. I could not find how to do it for typescript tests. The runner can run js tests, but I could not run the tests for bundled js (`ant-design-blazor.js`) - `mocha` (the js testing framework) apparently does not handle well the way of bundling we use. Besides, I think we should be testing in typescript since we write in typescript. So I added the call to tests during `AntDesing.Tests.Js` build. I also added in test section of `pull-request.yml`, but I guess it is redundant, since they are run anyway during build... Hopefully, Microsoft will add direct typescript tests from test runner. Unless you know how to do it, then please let me know.

ElderJames:
There was an unexpected commit
image

anddrzejb:
@ElderJames Getting rid of the commit failed spectacularly, so I am closing this PR and opening a new one. Cleaning this up is taking too much of my time and I find it faster and easier to just get the commit I need and put them in a new PR.

PR #1727, opened 2021-07-11
fix: click dialog cause the mask click event triggered

zxyao145:
The last fix was merged into the feature branch, so this fix wants to be merged into the feature branch

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  fix: click dialog cause the mask click event triggered   |
| 🇨🇳 Chinese |     修复dialog点击导致mask的点击事件被触发的bug      |

### DIFF
--- a/components/modal/core/Dialog.razor.cs
+++ b/components/modal/core/Dialog.razor.cs
@@ -138,20 +138,20 @@ namespace AntDesign
             _dialogMouseDown = true;
         }
 
-        private async Task OnMaskMouseUp()
-        {
-            if (Config.MaskClosable && _dialogMouseDown)
-            {
-                await Task.Delay(4);
-                _dialogMouseDown = false;
-            }
-        }
-
         private async Task OnMaskClick()
         {
-            if (Config.MaskClosable && !_dialogMouseDown)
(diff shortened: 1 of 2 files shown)


components/modal/core/Dialog.razor.cs
```diff
@@ -138,20 +138,20 @@ private void OnDialogMouseDown()
             _dialogMouseDown = true;
         }
 
-        private async Task OnMaskMouseUp()
```
zxyao145, line 141: I'm not sure why Antd listens to the mouseup event to determine if dialog was clicked.

It seems that it's OK to reset _dialogMouseDown in MaskClick, no matter whether the sub DOM is clicked or not, the mask click event is always triggered, and it is behind mouseup event. So I remove the MouseUp event of dialog-wrap DOM.

ElderJames (approved):

PR #1766, opened 2021-07-23
fix(module:select): duplicated tags

### LINKED ISSUE #1645: Select component in 'tags' mode is duplicating tags
### Describe the bug

The Select component in 'tags' mode is duplicating tags and producing unpredictable behavior. This seems to occur when at least two custom tags are inserted, the second one causing the first to get duplicated. Inserting additional tags does not cause any duplication, but trying to delete the duplicated tag causes the other tags to get duplicated also. This does not seem to affect the Values variable, which stores the correct data, without duplicates, altough it is presented incorrectly.

This can be easily reproduced in the 'Select' demo (https://antblazor.com/en-US/components/select), on the 'Tags' section. Initially, there is only one tag present (0f15). If one enters a new tag ('A', for example) and presses enter. Then a new tag ('B', for example), pressing enter again. This will cause the 'A' tag to be duplicated. Inserting other tags will not produce this behavior again, but trying to delete the second 'A' will produce unexpected results on the remaining tags.

image

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
1. Fixes #1645 
2. Any mouse button click on "x" in selected item is removing it.

### 💡 Background and solution
1. `SelectOption` was being removed in the `Dispose()` method irrespective of it being selected. `Dispose` method now checks if `SelectOption` is selected and prevents removing.
2. Selected item will be removed only if the "x" receives left mouse button click.
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -12,6 +12,7 @@ using AntDesign.Select.Internal;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
 using OneOf;
+using OneOf.Types;
 
 #pragma warning disable 1591 // Disable missing XML comment
 #pragma warning disable CA1716 // Disable Select name warning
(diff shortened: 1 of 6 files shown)


ElderJames (approved):
LGTM

PR #1786, opened 2021-07-28
fix(module: select): fix null value

### LINKED ISSUE #1783: Select component will throw exception if the value is null
I am on version 0.84.  My guess was one of my select option item's value field is null. This is kind of my fallback option.
Right now, I work around it by enabling AllowClear and Placeholder. 

In order to verify my assumption, I modified Select.OnDataSourceChangeTests and changed the Person's Id type to string and set it as null. When I ran the test, exception was thrown at line 684. I guess we need to handle null value comparison.

image

image

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Fixes #1783

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | accept `null` as select option value |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -681,7 +681,7 @@ namespace AntDesign
                 if (processedSelectedCount > 0)
                 {
                     if (SelectMode == SelectMode.Default)
-                        isSelected = value.Equals(_selectedValue);
+                        isSelected = ReferenceEquals(value, _selectedValue) || value?.Equals(_selectedValue) == true;
                     else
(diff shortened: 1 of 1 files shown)


anddrzejb (approved):
LGTM

ElderJames (approved):

PR #1788, opened 2021-07-28
fix(module:rangePicker): ShowTime issues addressed

### LINKED ISSUE #1482: RangePicker 右侧日期选择有很多问题
### Describe the bug

GIF

1、右侧选择时间后会自动跳到下一个月且无法返回上个月
2、选择下一个月时候会跳过一个月
3、选择月份会自动加一个月，如选择3月份会跳到4月份，选择4月份会跳到5月份

```` c#
<RangePicker TValue="DateTime?[]" ShowTime="@true" />

````

### Further technical details
- AntDesign Nuget Package version：0.8.0

### LINKED ISSUE #1677: RangePicker: not revert to valid value after input an invalid value in the first textbox
### Describe the bug
ant design blazor behavior:
动画05
ant design react behavior:
动画06

https://antblazor.com/en-US/components/datepicker#components-datepicker-demo-range-picker

### Further technical details
- AntDesign Nuget Package version 0.8.2

### LINKED ISSUE #1678: RangePicker: Clicking "Ok" button doesn't transfer the cursor to the second textbox
### Describe the bug

动画07

react antD behavior:
动画08

https://antblazor.com/en-US/components/datepicker#components-datepicker-demo-range-picker

### Further technical details
- AntDesign Nuget Package version 0.8.2

### LINKED ISSUE #1680: RangePicker: when `ShowTime` is enabled, for the second picker, the month grows by 1 when clicking any button
### Describe the bug

动画04

and when `ShowTime` = false, there is a similar bug for the next month and next year buttons:
动画05

动画06

https://antblazor.com/en-US/components/datepicker#components-datepicker-demo-range-picker

### Further technical details
- AntDesign Nuget Package version 0.8.2

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
1. Fixes #1482
Fixes #1680 
2. Fixes #1677 
3. Fixes #1678 
4. Focus issues
5. Visual issue for 2nd picker - extra space:
image

### 💡 Background and solution
1. When next month/year was clicked, base year/month (that would be incremented) was taken from the second picker instead of the first, but applied to the first. Solution applied: for the second picker, the value is decreased by predefined value to achieve expected increments.
2. After my changes, this is no longer recreatable. Either this PR fixed this or was fixed in previous PRs but not mentioned.
3. Ok click was only closing. Added logic to move to next/previous input.
4. Addresses focus remaining after selection has been made. There are still a number of issues with focus indication. I have not addressed them here as they are extremely hard to test, due to overlay issues and DatePicker spawning myriad of refresh events. Hopefully these will be easier to solve once at least overlay issues are solved (should at least be possible to write automated tests).
5. Special scenario for 2nd picker when `ShowTime` is `true` - the first picker was rendered as empty div and was causing this extra space. Removed rendering of the first picker div when 2nd is visible.
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


anranruye (changes requested):
1. #1677 still exists. To reproduce this issue, you should press "enter" after input.
动画08

2. In react antD, the cursor is moved to the prev/next input after clicking "Ok" only one time, then the dropdown is closed. The logic of this PR is different. The cursor is moved to the prev/next input every time "Ok" is clicked unless the picker value is changed(after the picker value is changed, the cursor will still move one time, then the dropdown will be closed).
动画09
动画10

~3. Focus remaining still happens when not select both date values and the picker wrongly loses focus when switch to the prev/next input(exsiting issue). In the second gif, the week range picker can not lose focus, not sure the condition to reproduce.~
动画11
动画12

anddrzejb:
1. That is really awkward 😄 . I tested this on 3 different browsers on my machine and everywhere I could not recreate it. And I was trying with `Enter` key... I will have another look.  
2. It is not a part of this PR. Otherwise I would include the issue where you mention it. My primary goal was to solve the major issue that was making `RangePicker` with `ShowTime` unusable. Other issues - will be addressed in future PRs. 
3. So... ok here? I am aware about this issue. Focus issues here gave me a headache and always consume too much of my time. I want to address them only once testing is solved, so I can actually do a fast red-green path.

anranruye:
1. That's strange, I will also have a look.
2. Yes, this doesn't matter. I didn't mean we should have the same behavior as react ant design at this moment. But I think it's better to have a consistent behavior. Either to always move the cursor, or always close the dropdown. It's ok to make it in the future.
3. Eventually I think these issues are not parts of this PR.

anranruye:
@anddrzejb confirmed that #1677 only happens on blazor wasm, blazor server is ok.

anddrzejb:
I really was sure I checked wasm. I will have a look once I am avaliable.

anddrzejb:
I found out what was it. I didn't get it on mine originally, due to the fact that for ant development, I created my own playground project where I usually work on a single component. The issue here is showing only when there are more components and jsinterop is competing with blazor. The problem here is that input value was not yet passed to `DatePickerInput` and already `OnKeyDown` event was triggered that was actually checking that value (and clearing it if it was incorrect). I do not know how else to solve this problem, except by using `await Task.Delay()` method, so this is how I solved it. This solution also fixed a different bug in server side - when incorrect value was entered, the input was cleared but focus was being moved to the next input. Now it stays on the same input - same as react antD.

anranruye (approved):

ElderJames (approved):

PR #1790, opened 2021-07-29
fix(module:tree): amend documentation of check/uncheck all

lukblazewicz:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Misleading description of the methods.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | amend code-documentation of check/uncheck all |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tree/Tree.razor.cs
+++ b/components/tree/Tree.razor.cs
@@ -314,7 +314,7 @@ namespace AntDesign
         public EventCallback<string[]> CheckedKeysChanged { get; set; }
 
         /// <summary>
-        /// Dechecked all selected items
+        /// Checks all nodes.
         /// </summary>
         public void CheckedAll()
         {
@@ -324,7 +324,9 @@ namespace AntDesign
             }
         }
 
-        // Decheck all of the checked nodes
+        /// <summary>
+        /// Unchecks all nodes.
+        /// </summary>
         public void DecheckedAll()
         {
             foreach (var item in ChildNodes)


lukblazewicz:
I think the next step should be renaming both methods to CheckAll() and UncheckAll() respectively but obviously, as a breaking change, it should be planned for the next major release. What do you think?

components/tree/Tree.razor.cs
```diff
@@ -324,7 +324,9 @@ public void CheckedAll()
             }
         }
 
-        // Decheck all of the checked nodes
+        /// <summary>
+        /// Unchecks all nodes.
+        /// </summary>
         public void DecheckedAll()
```
ElderJames, line 330: Should we rename this method?

lukblazewicz: I believe so:
https://github.com/ant-design-blazor/ant-design-blazor/pull/1790#issuecomment-888817711

ElderJames:
> I think the next step should be renaming both methods to CheckAll() and UncheckAll() respectively but obviously, as a breaking change, it should be planned for the next major release. What do you think?

I think we can bring all to `feature` branch and will ship in next major release.

lukblazewicz:
Okay, I'll then abandon this PR and include the abovementioned changes along with methods renaming in a separate PR to the feature branch.

PR #1791, opened 2021-07-29
fix(module:interop): reorganization into classes and typescript tests

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
This is a redo of PR #1711. 

> I tried here to reorganize the javascript into modules/classes. My approach was mostly to use static classes instead of group of functions in order to organize them into more coherent groups. The naming is arbitrary and of course should be discussed.
> I removed some of the functions without any references or where I found a functional duplication. In obvious places I added types (so to actually use the TypeScript main feature). I haven't really touched Caret & Dragable classes, except moving them to different folders. I noticed that there are still a lot of functions that could be improved, but I figured this is not the scope of this PR.
> I also added a singleton class with js state - in hope it will be used as such. So if in the future there is a need to add a structure to hold some js specific data - the state should be the place to hold it. It is already used in that capacity (keeps dictionaries).
> Also, JSInteropConstants was changed - the structure of *.ts files is mostly reflected in static helper classes. Because I did not want to change all components, that is why I left the calls in JSInteropConstants with redirects to respective static class. I guess, with time, we could replace them with direct calls to static helpers. 

> Last couple of commits are adding a new project that contains tests for typescript. Currently it has only 3 tests - a proof of concept. I handle there basic tests including also samples on how to test dom related functions/elements.
> Another change here is that there is new gulp task that except bundled js, will also create separated js files (matching 1-to-1 typescript file structure). This is probably the first step to requested js isolation.
> Current testing approach is not my favourite. My issue with it is that I cannot run the tests in VS test runner. I could not find how to do it for typescript tests. The runner can run js tests, but I could not run the tests for bundled js (ant-design-blazor.js) - mocha (the js testing framework) apparently does not handle well the way of bundling we use. Besides, I think we should be testing in typescript since we write in typescript. So I added the call to tests during AntDesing.Tests.Js build.  Hopefully, Microsoft will add direct typescript tests from test runner. Unless you know how to do it, then please let me know.

I also added in test section of pull-request.yml. They may seem redundant but this is a different test script that also runs test coverage. I am not sure if the coverage is being uploaded to codecov though.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames:
It seems that test coverage reports cannot be submitted..

https://github.com/ant-design-blazor/ant-design-blazor/pull/1791/checks?check_run_id=3189369501#step:8:37

anddrzejb:
Yeah, I noticed that. I am not very skilled in github actions. I will try to fix it.

anddrzejb:
@ElderJames My last commit seem to actually hit the sweet spot. Fixes:
1. Now the test coverage has both cs & ts (at least looking at that report it seems so). Overall coverage has decreased but that was expected since a number of new typescript files had been added.
2. It seems that my changes to the `dotnet test` script caused the test action to run much faster (my previous slow test was over 18 minutes, now it is below 3 minutes - that looks like a really nice optimization. To achieve this I:
    - updated nuget package for `coverlet.collector` to the newest version, 
    - point test directly to the test project (this way the `TestKit` is omitted but since no tests are there, there is no need to run the coverage there) 
    - run tests in `Release` configuration (that was the source of biggest gain).  
3. Last change - I switched `codecov/codecov-action` to `v2` because:
   - I had a problem and for some reason only last coverage report file was being taken, 
   - `v1` is depreciated (please check the [`codecov-action` repo readome](https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1).

ElderJames (approved):
This helps a lot.

ElderJames:
We also need to test the cases only for .net 6

anddrzejb:
> We also need to test the cases only for .net 6

I am sorry, I do not understand what do you mean. We already have preprocessor directives in our tests. You would like to have a separate test project just for net6?

ElderJames:
No, I mean some test is only complie in .net 6 target. Because we use `#if NET6` to separate them.

anddrzejb:
Well...here the issue - in some cases these are full tests that can only be executed in net6. But there are also tests that are working in net5 but a line or 2 have the net6 preprocessor directive added. Regardless, I still do not fully understand what you have in mind - how would you like to separate them if not in a separate project? Anyway, all this cannot be done until `AntDesign` moves to net6.

ElderJames:
@anddrzejb I found that the upload component is broken... Could you please take a look if you have time?

anddrzejb:
I will have a look. Any particular functionality you have in mind or basic?

ElderJames:
There is no response when you click the upload button. But the 0.9.0 is work well.

Issue #992, opened 2021-01-17
Performance best practices


anddrzejb:
While doing my work for this and other projekt, I came back to [blazor best practices page](https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0). There is a lot of interesting points there, that probably could and should be applied. Maybe this issue could be a way of writing down all best practices, but now I mostly want to emphasize the performance gain coming from [this paragraph](https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0#ensure-cascading-parameters-are-fixed). It looks like a very quick and easy fix. I am already going through all components as a part of my works on PR #937 and I guess I could apply this quick change there.

ElderJames:
Good suggestion.
Maybe we can list a checklist to setup each step to improve the performance.

anddrzejb:
## Performance best pracitces
- when dealing with `[Parameters]`:
  - whenever possible use primitive types ([source](https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0#avoid-unnecessary-rendering-of-component-subtrees))
  - if primitive types are not possible, configure `ShouldRender` to return `false` for components that never change after initial render ([source](https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0#avoid-unnecessary-rendering-of-component-subtrees))
  - avoid attribute splatting - if possible avoid using [`CaptureUnmatchedValues`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.parameterattribute.captureunmatchedvalues?view=aspnetcore-5.0#Microsoft_AspNetCore_Components_ParameterAttribute_CaptureUnmatchedValues) flag ([source](https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0#avoid-attribute-splatting-with-captureunmatchedvalues))
- when dealing with `[CascadingValue]`:
  - set `IsFixed` attribute to `true` when passing `this`  ([source](https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices?view=aspnetcore-5.0#ensure-cascading-parameters-are-fixed))
[...]

PR #1848, opened 2021-08-17
fix(module: overlay): positioning moved to js

### LINKED ISSUE #886: [select] in the form the selected content is slow to respond
The selected content is slow and particularly slow to respond.
If it's not in the Form, it's a little faster to select it

@ElDiddi In the new Select use, do you have this situation

### LINKED ISSUE #1075: Menu drop-down lists does not close
### Describe the bug

image

The Menu component is really good, but there is one small thing that leaves a bad impression on the user. It is about hiding the options. This problem with hiding seems to exist in general, because you have it also in ToolTip (you can see it in the picture).

Specifically, in the given image, we dragged the mouse from left to right in the horizontal menu bar normally (not too fast, not too slow) and the drop-down lists gradually opened but did not close again.

We have seen that two parameters are provided in the API for the menu :
- SubMenuCloseDelay
- SubMenuOpenDelay

We hope that with the next version it will be possible to solve the problem with these two parameters.

Thanks

Translated with www.DeepL.com/Translator (free version)

we dragged the mouse from left to right in the horizontal menu bar normally (not too fast, not too slow) and the drop-down lists gradually opened but did not close again.

### Further technical details
- AntDesign Nuget Package version 0.6
- Community 2019

### LINKED ISSUE #1388: RangePickerComponent doesn't fit when the page size changes
### Describe the bug

When the page size changes, the RangePicker component wouldn't fit into the page, which causes some of the dates to be outside the page.

This snapshot should decribe the bug clearly:
The page size is changed by pulling out the debug section of the browser.
企业微信截图_16188128808073

### Further technical details
- AntDesign Nuget Package version: 0.8.0
- IDE : VS 2019 version 16.9.4

### LINKED ISSUE #1414: Modal: sometimes modal is closed by double clicking
### Describe the bug

Modal will be closed by double clicking an inner date picker if MaskClosable=true. 
Update: Also reproduce this with `Select` component.
动画

```C#
<Button OnClick="()=> { openModal = true; }">Click</Button>

<Modal Title="Modal" DestroyOnClose="true" Visible="openModal" OnCancel="() => { openModal = false; }">
    <div style="height:500px;">
        <DatePicker @bind-Value="dateTime"></DatePicker>
    </div>
</Modal>

@code{
    bool openModal;
    DateTime dateTime = DateTime.Now;
}
```
### Further technical details
- AntDesign Nuget Package version: 0.8.0

### LINKED ISSUE #1557: <Select> Very slow

### LINKED ISSUE #1565: Modal closes when Form inside
Trying to display a form inside a Modal:

```
@using System.ComponentModel.DataAnnotations
@inject Services.CompanyService _companyService;

<Button OnClick="@ShowModal">New order</Button>
<Modal
    Visible=@_visible
    OnOk=@OnOk 
    OnCancel=@CloseModal 
    ConfirmLoading=@_confirmLoading
>
    <Title>
        Create new order
    </Title>
    <Content>
        <Form 
            @ref=@form
            OnFinish=@OnSubmit
            Model=@FormModel LabelColSpan="4"
            WrapperColSpan="16"
            Size="@AntSizeLDSType.Small"
        >
            <FormItem Label="Number">
                <Input @bind-Value=@context.OrderNumber />
            </FormItem>
            <FormItem Label="Date">
                <DatePicker Picker="@DatePickerType.Date" @bind-Value="@context.Date" />
            </FormItem>
        </Form>
    </Content>
</Modal>

@code {
    bool _visible = false;
    bool _confirmLoading = false;
    Form<Model> form;

    private class Model
    {
        public int CompanyId { get; set; }
        [Required]
        public int ContractId { get; set; }
        [Required]
        public string OrderNumber { get; set; }
        [Required]
        public DateTime Date { get; set; }
    }

    Model FormModel = new Model();

    private void ShowModal()
    {
        Console.WriteLine("Show modal");
        _visible = true;
    }

    private void CloseModal()
    {
        _visible = false;
    }

    private void OnCompanyChanged(Company company)
    {
        ListOfContracts = company.Contracts;
        FormModel.ContractId = 1;
    }

    private void OnOk(MouseEventArgs e)
    {
        form.Submit();
    }

    private async Task OnSubmit()
    {
        _confirmLoading = true;

        StateHasChanged();

        await Task.Delay(2000);

        _confirmLoading = false;
    }
}
```

Issues:
- Click on `Ok` button closes the modal
- Resetting the date in the DatePicker also closes the Modal

Expected behavior:
- Modal remains visible as long as the `_visible` is not set to `false` explicitly.

-----------

P.S.: https://antblazor.com/en-US/components/modal#components-modal-demo-position in the API section > ConfirmOptions `false` is spelled as `fasle`

### LINKED ISSUE #1613: use serverside, datepicker component  cause  the page flash when first time to select date
serverside 模式下，页面中使用了datepicker控件，加载页面后，首次选择日期时页面会闪烁。
好像tooltip也有这个问题。

### LINKED ISSUE #1672: Tooltip triggers a scroll bar the first time it is displayed near the right side of the screen.
1. When you use the ToolTip component near the right side of the screen, the first time ToolTip is displayed, the screen appears with an instant scroll bar.

ps . when expand row data, trigger the row click event. 

sample code.
[demo.zip](https://github.com/ant-design-blazor/ant-design-blazor/files/6713323/demo.zip)

### LINKED ISSUE #1836: Overlay is not in the correct position when the menu component is in the contracted state
### Describe the bug

Overlay is not in the correct position when the menu component is in the contracted state 

You need to slide the mouse over the shrink menu
### Exceptions (if any)

Snipaste_2021-08-13_11-11-16
Snipaste_2021-08-13_11-16-18

### Further technical details
- AntDesign Nuget Package 0.9

### LINKED ISSUE #1880: The sub menu on quick hover gets stuck
As you can see from the following when the sider is collapsed and you move the mouse quickly to trigger the hover effect the menus get stuck in some cases. On slower machines the effect is much more viewable.

image

```
<Sider Class="bg-white" Width="280" Collapsible Collapsed="CollapseNavMenu" NoTrigger>
    <NavLink href="/">
        <img class="logo d-flex mx-auto" src="logo.png" alt="Logo" />
    </NavLink>
    <Menu Theme="MenuTheme.Light" Mode="MenuMode.Inline" InlineCollapsed="CollapseNavMenu">
        <SubMenu>
            <TitleTemplate>
                <AntMenuTitle Text="Dashboards" Icon="dashboard" Theme="outline" />
            </TitleTemplate>
            <ChildContent>
                @foreach (DashboardDto dashboard in _dashboardList)
                {
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                }
            </ChildContent>
        </SubMenu>
        <SubMenu>
            <TitleTemplate>
                <AntMenuTitle Text="Realtime" Icon="clock-circle" Theme="outline" />
            </TitleTemplate>
            <ChildContent>
                <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
            </ChildContent>
        </SubMenu>
        <SubMenu>
            <TitleTemplate>
                <AntMenuTitle Text="Reports" Icon="fund" Theme="outline" />
            </TitleTemplate>
            <ChildContent>
                <SubMenu>
                    <TitleTemplate>
                        <AntMenuTitle Text="Reports" Icon="table" />
                    </TitleTemplate>
                    <ChildContent>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                </SubMenu>
                <SubMenu>
                    <TitleTemplate>
                        <AntMenuTitle Text="Reports" Icon="area-chart" />
                    </TitleTemplate>
                    <ChildContent>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    </ChildContent>
                </SubMenu>
            </ChildContent>
        </SubMenu>
        <SubMenu>
            <TitleTemplate>
                <AntMenuTitle Text="Reports" Icon="control" Theme="outline" />
            </TitleTemplate>
            <ChildContent>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
            </ChildContent>
        </SubMenu>
            <SubMenu>
                <TitleTemplate>
                    <AntMenuTitle Text="Reports" Icon="global" Theme="outline" />
                </TitleTemplate>
                <ChildContent>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                </ChildContent>
            </SubMenu>
            <SubMenu>
                <TitleTemplate>
                    <AntMenuTitle Text="Reports" Icon="api" Theme="outline" />
                </TitleTemplate>
                <ChildContent>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                </ChildContent>
            </SubMenu>
            <SubMenu>
                <TitleTemplate>
                    <AntMenuTitle Text="Reports" Icon="sound" Theme="outline" />
                </TitleTemplate>
                <ChildContent>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                    <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
                </ChildContent>
            </SubMenu>
        <AntDesign.MenuItem RouterMatch="NavLinkMatch.All" RouterLink="test">Test</AntDesign.MenuItem>
    </Menu>
</Sider>

```

### Further technical details

### LINKED ISSUE #1928: The datePicker will extend beyond the edge of the screen
### Describe the bug
The date picker goes beyond the edge of the screen when it is in the extra part of the card
A5(LS5F0HGVNV7%E7BA$MRC
This should be the normal style-by antd pro:
(Z{)25T )Q8HNUW%V6QOC{2

- AntDesign Nuget Package version : 0.93

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1557
Fixes #1836
Fixes #1880
Fixes #1672
Fixes #1613
Fixes #1388
Fixes #1075
Fixes #1565
Fixes #1414
Fixes #886
Fixes #1928

### 💡 Background and solution
The goal of this PR is to improve performance for ServerSide without affecting Wasm. Major change has happened in the `Overlay`. All code that was responsible for overlay positioning, size change detection & scroll has been removed. Instead, a new class `Overlay` has been created in the file `overlay.ts`. That class handles:
- calculation of the position
- recalculation on container size change (uses `ResizeObserver`)
- recalculation on trigger mutation: either size or position (by monitoring `style` attribute, uses `MutationObserver`) - direct example is shown in the slider demo, when tooltip has to "follow" the slider's edge (it should be much more performant now)
- recalculation on scroll - for now any scroll causes recalculation, but this could be optimized to actually detect if recalculation is needed
The calculation logic has been changed. Originally, all calculations were done from the perspective of `top` and `left` coordinates. The `Overlay` now is using all coordinates for its position calculation (no all at once, but depending on the selected placement, a pair is selected, ex: `top` & `right`, `bottom & `left`, `top` & `left`, etc). This addresses the changes proposed by @anranruye in PR #1799 (I also remebered about #1797).  

The logic of the new calculation is also explained in [this miro diagram(title "overlay calculation logic") - you may need to zoom-out](https://miro.com/app/board/o9J_l9OYUeA=/).
A big challenge for me was also to include typescript tests with dom support, mocking & theories. This can serve as an example as well for future contributors on how ts tests might be done. I am sure the tests could be optimized, but I decided to leave them as such - maybe this simpler approach will be more understandable to someone who just wants to start with tests. 

There are 2 breaking changes - I switched from `PlacementType` (class) to `Placement` enum in `[Parameter]` usage, while still using the `PlacementType` class and its logic. I also switched the `TriggerType` class to `Trigger` enum in `[Parameter]` usage. There are 2 reasons for changes: 
1. According to #992, for performance purposes we should use primitive types to avoid unnecessary rendering of component subtrees
2. It is easier to move an enum between C# and JS. This is however a breaking change that will affect a lot of people. 

What is missing:
- currently the adjustment strategy is the same as previously - if overlay does not fit, switch sides; the difference is that the sides are switched independently of a direction; what possibly is missing is to attempt to fit the overlay by disconnecting it from the trigger (in a way that it can even obstruct the trigger)
- more adjustment scenarios - when simple switching sides is not working, continue with all other possible placements
- scroll optimization - run only when it is needed; a sample is done in the ts file (commented out code in `onScroll` function)
- for `RangePicker` there probably should be more functionality available that handles situation (for example on a mobile device) when `RangePicker` just does not fit (width is larger than device screen width);

Testing experience - I found that Visual Studio Code is actually best to develop & test typescript. The test runner there actually picks up the tests and they can be run individually (and most important, debugged - no need to deal with god-awful debugging in browsers). It even picks up separate theory scenarios:
image
However Visual Studio Code does not come with all the goodies to run the tests. Maybe it would be beneficial to create a wiki entry on how to configure dev env for ts tests.

Other stuff:
1. `Overlay` & `OverlayTrigger` were optimized to reduce number of rerenderings
2. `MenuItem` had been changed not to render `Tooltip` every time but only on a compact & root menu (previously it was always rendering it)
3. `Select`, `DatePicker` & `RangePicker` have now available `[Parameter] BoundaryAdjustmentMode`
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  `TriggerType` switched to `AntDesign.Trigger` enum. `PlacementType` switched to `AntDesign.Placement` enum. `Autocomplete`, `Cascader`, `PaginationOptions`, `Select`, `DatePicker` & `RangePicker` have now available  `[Parameter] BoundaryAdjustmentMode`        |
| 🇨🇳 Chinese |           |


ElderJames:
Awesome @anddrzejb ! It helps a lot!

anddrzejb:
@ElderJames I forgot to mention in the PR comment: I checked with the same VPN setup as described in #1557 [this comment](https://github.com/ant-design-blazor/ant-design-blazor/issues/1557#issuecomment-869031425). The execution time dropped to 273ms in comparison to 39ms locally. So that is significant improvement to previous 1.65sec (which was already optimized by my other attempts) and around 3-4 sec on the unaltered version.

ElderJames:
That's great improvement!

Is there any chance that those adjustment strategies will be added?

> currently the adjustment strategy is the same as previously - if overlay does not fit, switch sides; the difference is that the sides are switched independently of a direction; what possibly is missing is to attempt to fit the overlay by disconnecting it from the trigger (in a way that it can even obstruct the trigger)
more adjustment scenarios - when simple switching sides is not working, continue with all other possible placements
scroll optimization - run only when it is needed; a sample is done in the ts file (commented out code in onScroll function)
for RangePicker there probably should be more functionality available that handles situation (for example on a mobile device) when RangePicker just does not fit (width is larger than device screen width);

anddrzejb:
> That's great improvement!
> 
> Is there any chance that those adjustment strategies will be added?

A chance exists, but we need to discuss which strategies should be used and in what order. And maybe we could be smart about it and not just try every possible scenaio. Also, some components should not apply disconnect scenarios (for example tooltip). I will try to prepare some kind of description of what I foresee as a problem and maybe we could discuss what is the preferred solution. Do you want to make these solution part of this PR?

ElderJames:
@anddrzejb Of course not, you can bring them to another PR.

I think we could refer other libraries, such as:

- https://material.angular.io/cdk/overlay/overview
- https://popper.js.org/

anddrzejb:
I am on a leave now, but I will think about the improvements.
I did have a look at popper before. I liked it approach to scrolling behavior when the arrow is involved, but I am not sure if this is what we want. It seems to me that we already surpass the react antD when it comes to overlay behaviour on boundaries detection. And I have other performance improvements on my mind that I think should be addressed before any other new visual features. I know the performance changes are not as flashy as visual  and really hard to

ElderJames:
Totally agree.
Let's review this PR first and I hope to merge it this weekend if it goes well。

ElderJames:
I think it also needs to follow the resize of the trigger to adjust the positioning.

select-dropdown

anddrzejb:
> I think it also needs to follow the resize of the trigger to adjust the positioning.

@ElderJames you are absolutely right. Somehow I missed that. Please check my commit 0cecc3c. It is really easy with `ResizeObserver`.

ElderJames:
The dropdown of submenu will be repeatedly opened in some cases

menu-dropdown

anddrzejb:
> The dropdown of submenu will be repeatedly opened in some cases

I think this is not repeatedly shown, but sometimes overlay stays open instead of closing. I just tried to recreate it and failed. This is a bug that is coming from original implementation. `Overlay` hiding is triggered either by JS call or by overlay trigger (as far as I remember) and I did not change that. I already noticed that when there is a lot of things happening, some js events seem to be lost - I guess this may happen, since js is only single threaded. I don't know - maybe a web worker could solve us this problem? This is something I never worked with (there is always a first time though...). What do you think?

anddrzejb:
> The dropdown of submenu will be repeatedly opened in some cases

So I had a closer look and indeed I can see that this problem is coming from trigger event `onMouseLeave`. Actually, js event fires properly. The problem happens in the `Overlay`. I believe the underlaying problem is indeed single threaded nature of js. Anyway, the issue is that sometimes the `Show` method of the `Overlay` does not manage to finish (it has few awaits inside of it and the js itself can sometimes take a bit of time to execute) when the `onMouseLeave` event triggers `Hide` method of the `Overlay`. During `Hide` method, a check is done if `_isOverlayShow == true`. The problem is that sometimes it is still `false` because `Show` method hasn't finished processing and did not set it to `true`. The code to handle this scenario is basically a loop (up to 100 times) that waits for `_isOverlayShow` to become true. I probably would prefer using a `while` loop, but I am afraid there is something I overlooked and we will end up with an endless one. Please have a look at the gif and the code snippets I used during the gif generation. I added this piece of code in the begining of `Show` method:
```csharp
        internal async Task Show(int? overlayLeft = null, int? overlayTop = null)
        {
            Console.WriteLine($"Showing {Trigger.Id}");
...
```
and this code in the beginning of `Hide`:
```csharp
        internal async Task Hide(bool force = false)
        {
            Console.WriteLine($"Hiding {Trigger.Id}");
            if (_isOverlayDuringShowing)
            {
                for (int i = 0; i < 100; i++)
                {
                    Console.WriteLine($"Hiding {Trigger.Id}: waiting {i+1} time...");
                    await Task.Delay(10);
                    if (_isOverlayShow)
                    {
                        break;
                    }
                }
            }
            if (!_isOverlayShow)
            {
                Console.WriteLine($"Hiding {Trigger.Id}: stopped due to _isOverlayShow: {_isOverlayShow}");
                return;
            }
...
```
And the result:
overlayMouseLeave


As you can see on the gif, sometimes the loop repeats up to 16 times. In my other tests, I had it up to 26 times. 100 times x 10ms gives a moderate 1 second wait time. At this point I am waiting for suggestions.

I applied the above changes to commit 708797a.

anddrzejb:
Commit 670c829 is extracting the `for` loop into a `WaitFor` function. I found out that in case of `Slider` with visible tooltip on the ServerSide the same wait is needed - I guess because of the overlay is being rendered before trigger gets back the element reference. Without the wait, overlay throws an exception, because `Trigger.Ref.Id` is null.

mutouzdl:
@anddrzejb It's amazing.
I'm looking at all the components that involve overlays one by one (this may take some time).
And just found out there's a problem with Dropdown.

In "Cascading menu" demo, when dropdown is hidden, it appears in the top left corner for a while, then disappears

ElderJames:
I also find an issue on the widget fixed at the document.

widget-dropdown-issue

anddrzejb:
@ElderJames Commit 52ce937 contains some scrolling adjustments. Unfortunately I was wrong - intersection observer cannot really be used as an `onScroll` replacement. Also - `onScroll` is not really that slow, it actually fires every 20ms. The problem here is that it just fires so often that more serious calculations cannot be finished by the time next event is fired. The scenario you presented on the gif is actually problematic only when trigger (or any of its ancestors) has a `position:fixed`. That scenario also allows to greatly simplify new position calculation. So this is what I did. I think it behaves much better, but still sometimes a flicker is visible. The only other way I can think to solve this problem is to actually for triggers with `position:fixed` also make the overlay `position:fixed`. This would require either us to detect the position of the trigger using js (from the overlay) or add a parameter so the consumers can actually define that the trigger is going to be fixed. With that, overlay's position would also have to be calculated differently.

mutouzdl:
@anddrzejb I was looking at the DatePicker component, and I found a new problem  

In the "Base" Demo:
step1: Click on the DatePicker
step2: Scroll the browser until the DatePicker's Panel is out of view
step3: Scroll the browser to its original location

You should find that the DatePicker's Panel disappears and cannot be opened again

anddrzejb:
@mutouzdl Thanks for this great catch! I am really surprised I actually missed it. There was a problem with reverting adjustment calculation if the adjustment was not bringing better results. 
I added a test case to cover this scroll scenario.

anddrzejb:
@zxyao145 I am doing some debugging in the `Drawer` area (demo DropdownInDrawer). I need to use `PopupContainerSelecter` on all components that use `Overlay`. It needs to point to `Drawer`'s id. The problem is that currently the `Id` is assigned to `div` with `class="ant-drawer ant-drawer-open"`:
> Drawer.razor (line 4)
```razor
<div class="@ClassMapper.Class" @ref="@Ref" style="@_drawerStyle @InnerZIndexStyle @Style" id="@Id">
```
It seems to me that this particular `div` is not good as a container:
image

I do not think it is useful for any styling either. I would like to propose to move the `Id` to the next `div`:
> Drawer.razor (line 9)
```razor
<div class="ant-drawer-content-wrapper @WrapClassName " style="@WrapperStyle" id="@Id">
```

Do you think it would create any issues? It does solve my problem :smirk: (one of, anyway).

mutouzdl (approved):
All relevant components are working fine. I didn't find any other problems

ElderJames (dismissed):

anddrzejb:
@ElderJames please wait with merge. I found some hidden issues and have pending changes.

anddrzejb:
@ElderJames, @mutouzdl my latest commit fixes some weird behaviors in modal & drawer. Anyway, here are the fixes:
1. I added `BoundaryAdjustMode` to`AutoComplete`, `Cascader`  & `PaginationOptions`.
2. I made changes to `Dialog` & `Drawer` (@zxyao145 can you have a look at them and confirm the changes are not braking anything?)
    - `Dialog` - I moved the `Id` from the wrapping div to the dialog content div. `Overlay` uses the div with that `Id` as the default container when it is placed in `Dialog` (through `CascadingValue`). The issue here was that wrapping div does not really represent real dimensions of the dialog that the `Overlay` needs (it needs to know the boundaries of the container and wrapping div is basically full viewport):
image
    
     I also added `@onclick:stopPropagation` - I think this addition actually fixes #1414 & #1565 (and possibly removes the need for `@onmouseup` in wrap & `@onmousedown` in the document div but @zxyao145 has to actually confirm that). Anyway, my tests with `Modal` are showing that the aforementioned issues do not appear any longer (they were especially visible in server-side).
    - `Drawer` - I applied the same solution as in `Dialog` in regards to `PopupContainer` being passed as `CascadingValue`. This way all overlays placed as child elements in the `Drawer` will automatically pick it up as their container.
3.  In case of `Select` I noticed an unnecessary scroll event. This was happening in server-side when a `Select` component was low enough that to get to it, the page had to be scrolled. Then clicking on the `Select` was causing the page to scroll (sometimes just a bit and other times to the top of the page):
scroll
I noticed this is due to scrolling to the selected/first item in the list. The scroll was done using javascript and it was done against the body. I changed that - now scroll is done against the scrollable selection area.
4. Fixes to `overaly.ts` - there was an issue when positioning an overlay in a containers like `Modal` & `Drawer`. Current approach when asserting when to switch position of the `Overlay` is to validate if requested placement fully fits in the container. If it does, then no adjustment is needed. If it does not, then placement is switched and new position is calculated. Next step is to compare how much of the overlay was visible before adjustment and after adjustment. If more of the overlay was visible before adjustment - adjustment is abandoned. If more of the overlay is visible after adjustment - the adjustment stays. This however has a hidden trap - what if the overlay is in a container that is in a modal? Like in this example:
image
So my change here is to try to detect if the container is in a state above the `body`. I detect if overlay's container or any of its ancestors has `z-index > 0`. If this condition is met, then during adjustment calculation I also take into account the visible area of the overlay in relation to the body. I am not sure though if maybe the `z-index` value comparison should be something different than 0. Our modal/drawer gets `z-index: 1000`. But maybe I should put here `z-index > 100`? Not sure... I don't think there are full proof solutions here. Maybe except if we add a parameter to `OverlayTrigger` - something to indicate to also check `Overlay` calculation against its visibility in relation to the body.
5. Some minor optimizations in typescript.
6. TS tests to cover the discovered bug scenarios.
7. Docs update

@mutouzdl, @ElderJames  I am really sorry I did not find these issues earlier. I know you spent a lot of time looking for the bugs. Unless new bugs are found, I think it is ready for merging.

zxyao145:
> Dialog - moved the Id from the wrapping div to the dialog content div. 

This may not be a problem. This is added by @mutouzdl and needs to be evaluated.

>  @onclick:stopPropagation

This is great! I never thought that it could just add `@onclick: stoppropagation` without adding `@onclick` events.

In my test, the above two modifications are all right.

anddrzejb:
Fixes #1880 
Fixes #1672
Fixes #1613
Fixes #1388
Fixes #1075
also confirmed by @zxyao145 : 
Fixes #1565 
Fixes #1414

ElderJames:
>  I am really sorry I did not find these issues earlier. I know you spent a lot of time looking for the bugs. Unless new bugs are found, I think it is ready for merging.

Take it easy, you've gone above and beyond what I thought you could do.

ElderJames:
@anddrzejb I've just merged a new component tree select into the feature branch and there are some changes to the select component in there that I'm worried will conflict with this pr. Do you have any suggestions for this?

anddrzejb:
@ElderJames I made very few changes to `Select` components. I know about the PR you merged (I have been watching it since it was submitted). I noticed some substantial changes in that PR, but my changes are so small and from what I can tell they will not create any merge conflicts. But even if they are, I will be watching for the feature merge to fix them.

zxyao145 (approved):

ElderJames:
@anddrzejb Sorry about the conflicts.

anddrzejb:
@ElderJames no worries, I was waiting for that PR to be merged. Now I need to figure out how it works, so I can apply it properly 😄 but I think it is a great improvement. Kudos to @tonyyip1969!

mutouzdl (approved):

ElderJames (approved):
Thank you @anddrzejb ! This is a huge improvement!

anddrzejb:
@ElderJames I am very happy it was approved. Now I will be waiting for devs complaints about breaking changes 😨

PR #1859, opened 2021-08-22
Breadcrumb updates

CAPCHIK:
### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [x] Demo update
- [x] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Breadcrumb can have dropdown with Overlay property, that feature has been added. Additionaly, in Dropdown component ChildContent was wrapped to `ant-dropdown-trigger` according to AntDesign, for applying to icon style font-width = 10 instead of 14.

BreadcrumbItem can have Href property to automatically create NavLink.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/breadcrumb/BreadcrumbItem.razor
+++ b/components/breadcrumb/BreadcrumbItem.razor
@@ -3,52 +3,54 @@
 
 <span @ref="Ref">
 
-    @if (Overlay != null)
-    {
-        <Dropdown>
-            <ChildContent>
-                <span class="ant-breadcrumb-link">
-                    @FormattedChildContent
-                </span>
-                <Icon Type="down"></Icon>
-            </ChildContent>
-            <Overlay>
-                @Overlay
-            </Overlay>
-        </Dropdown>
-    }
-    else
-    {
-        <span class="ant-breadcrumb-link">
-            @FormattedChildContent
-        </span>
-    }
+	@if (Overlay != null)
+	{
+		<Dropdown Placement="@PlacementType.BottomCenter">
+			<Unbound>
+				<span @ref="context.Current" class="ant-breadcrumb-overlay-link">
+					<span class="ant-breadcrumb-link">
+						@FormattedChildContent
+					</span>
+					<Icon Type="down"></Icon>
+				</span>
+			</Unbound>
+			<Overlay>
+				@Overlay
+			</Overlay>
+		</Dropdown>
+	}
+	else
+	{
+		<span class="ant-breadcrumb-link">
+			@FormattedChildContent
+		</span>
+	}
 
-    @if (BreadCrumb?.Separator != null)
-    {
-        <span class="ant-breadcrumb-separator">
-            @BreadCrumb.Separator
-        </span>
-    }
+	@if (BreadCrumb?.Separator != null)
+	{
+		<span class="ant-breadcrumb-separator">
+			@BreadCrumb.Separator
+		</span>
+	}
 
 </span>
 
 @code {
 
-    [Parameter]
-    public RenderFragment ChildContent { get; set; }
+	[Parameter]
+	public RenderFragment ChildContent { get; set; }
 
-    [Parameter]
-    public RenderFragment Overlay { get; set; }
+	[Parameter]
+	public RenderFragment Overlay { get; set; }
 
-    [Parameter]
-    public string Href { get; set; }
+	[Parameter]
+	public string Href { get; set; }
 
-    [CascadingParameter]
-    public Breadcrumb BreadCrumb { get; set; }
+	[CascadingParameter]
+	public Breadcrumb BreadCrumb { get; set; }
 
 
-    private RenderFragment FormattedChildContent => Href is null ?
-        ChildContent :
-    @<Microsoft.AspNetCore.Components.Routing.NavLink href="@Href">@ChildContent</Microsoft.AspNetCore.Components.Routing.NavLink>;
+	private RenderFragment FormattedChildContent => Href is null ?
+		ChildContent :@<a href="@Href">@ChildContent</a>
+	;
 }
--- a/components/breadcrumb/BreadcrumbItem.razor
+++ b/components/breadcrumb/BreadcrumbItem.razor
@@ -8,7 +8,7 @@
         <Dropdown>
             <ChildContent>
                 <span class="ant-breadcrumb-link">
-                    @ChildContent
+                    @FormattedChildContent
                 </span>
                 <Icon Type="down"></Icon>
             </ChildContent>
@@ -20,7 +20,7 @@
     else
     {
         <span class="ant-breadcrumb-link">
-            @ChildContent
+            @FormattedChildContent
         </span>
     }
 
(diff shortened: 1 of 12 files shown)


components/breadcrumb/BreadcrumbItem.razor
```diff
@@ -5,17 +5,22 @@
...
     @if (Overlay != null)
     {
-        <span class="ant-breadcrumb-overlay-link" nz-dropdown [nzDropdownMenu]="nzOverlay">
-            <span class="ant-breadcrumb-link">
-                @ChildContent
-            </span>
-            <Icon Type="down"></Icon>
-        </span>
+        <Dropdown>
+            <ChildContent>
```
ElderJames, line 9: We recommend using the unbound method to bind the trigger. And when we use it internally we need to use unbuound.

components/breadcrumb/BreadcrumbItem.razor
```diff
@@ -34,8 +39,16 @@
...
+    [Parameter]
+    public string Href { get; set; }
 
     [CascadingParameter]
     public Breadcrumb BreadCrumb { get; set; }
+
+
+    private RenderFragment FormattedChildContent => Href is null ?
+        ChildContent :
+    @<Microsoft.AspNetCore.Components.Routing.NavLink href="@Href">@ChildContent</Microsoft.AspNetCore.Components.Routing.NavLink>;
```
ElderJames, line 53: We can also try to implement route matching, similar to the implementation in the menu.

https://ant.design/components/breadcrumb-cn/#components-breadcrumb-demo-router-4

components/breadcrumb/BreadcrumbItem.razor
CAPCHIK: Can you show an example with using `Unbound` property? I can't find a working example.

components/breadcrumb/BreadcrumbItem.razor
CAPCHIK: Ok, I will try to create the same experience with `_breadcrumbs` field

 https://github.com/ant-design-blazor/ant-design-blazor/blob/473868547541c6acae6c8ea33caea71a4ba8d523/components/breadcrumb/Breadcrumb.razor.cs#L24

https://github.com/ant-design-blazor/ant-design-blazor/blob/473868547541c6acae6c8ea33caea71a4ba8d523/components/breadcrumb/Breadcrumb.razor#L7-L15

components/breadcrumb/BreadcrumbItem.razor
ElderJames:
```
<Dropdown>
    <Overlay>
     ...
    </Overlay>
    <Unbound>
        <div @ref="@context.Current">
         Trigger
        </a>
    </Unbound>
</Dropdown>
```

components/breadcrumb/BreadcrumbItem.razor
ElderJames: Here's what I mean, when the route changes, the matching item is automatically displayed.

breadcrumb-router

CAPCHIK: I understand, but in React example that behavior was realized using re-rendering component and parsing URL. In Blazor, that logic can be placed in Layout, I think.

So, using Options parameter can be implemented in the following way 
```razor
    <Breadcrumb AutoGenerate Options="@(new BreadcrumbOption[] {
                new BreadcrumbOption("Applications", "apps" ),
                new BreadcrumbOption("Application 1", "apps/1" ),
                new BreadcrumbOption("Detail", "apps/1/detail" ),
                })" />
```
With that option, we can generate array when need
```razor
@implements IDisposable
@inject NavigationManager NavigationManager
<div>
    <Breadcrumb AutoGenerate Options="@options" />
</div>

@code {

    private BreadcrumbOption[] options;

    private Dictionary<string, string> pathNames = new Dictionary<string, string>
    {
        { "components", "Components" },
        { "components/breadcrumb", "Breadcrumb" },
    };

    {
        base.OnInitialized();
        NavigationManager.LocationChanged += HandleLocationChanged;
        UpdateBreadcrumb(NavigationManager.ToBaseRelativePath(NavigationManager.Uri));
    }

    private void HandleLocationChanged(object sender, LocationChangedEventArgs args)
    {
        UpdateBreadcrumb(NavigationManager.ToBaseRelativePath(args.Location));
    }

    private void UpdateBreadcrumb(string location)
    {
        var pathSnippets = location.Split('/', StringSplitOptions.RemoveEmptyEntries);
        options = pathSnippets
            .Skip(1) // Lang
            .Select((_, index) => string.Join('/', pathSnippets.Skip(1).Take(index + 1)))
            .Select(r => new BreadcrumbOption(pathNames[r], $"{pathSnippets[0]}/{r}"))
            .ToArray();
        StateHasChanged();
    }


    public void Dispose()
    {
        NavigationManager.LocationChanged -= HandleLocationChanged;
    }
}
```

But if you want to see same code as in React example - we need to render BreadcrumbItem content from parse location logic:
```razor
@implements IDisposable
@inject NavigationManager NavigationManager
<div>
    <Breadcrumb>
        @foreach (var (label, href) in options)
        {
            <BreadcrumbItem>
                <NavLink href="@href">@label</NavLink>
            </BreadcrumbItem>
        }
        </Breadcrumb>
</div>

@code {

    private (string label, string href)[] options = new (string label, string href)[0];

    private Dictionary<string, string> pathNames = new Dictionary<string, string>
    {
        { "components", "Components" },
        { "components/breadcrumb", "Breadcrumb" },
    };

    {
        base.OnInitialized();
        NavigationManager.LocationChanged += HandleLocationChanged;
        UpdateBreadcrumb(NavigationManager.ToBaseRelativePath(NavigationManager.Uri));
    }

    private void HandleLocationChanged(object sender, LocationChangedEventArgs args)
    {
        UpdateBreadcrumb(NavigationManager.ToBaseRelativePath(args.Location));
    }

    private void UpdateBreadcrumb(string location)
    {
        var pathSnippets = location.Split('/', StringSplitOptions.RemoveEmptyEntries);
        options = pathSnippets
            .Skip(1) // Lang
            .Select((_, index) => string.Join('/', pathSnippets.Skip(1).Take(index + 1)))
            .Select(r => (pathNames[r], $"{pathSnippets[0]}/{r}"))
            .ToArray();
        StateHasChanged();
    }


    public void Dispose()
    {
        NavigationManager.LocationChanged -= HandleLocationChanged;
    }
}
```

Both examples can be shown on the documentation page, but the first example with the `Options` parameter looks too complicated. 

For integrating with router, we can create "path resolver" logic and place work with NavigationManager inside the Breadcrumb, and usage might look like this :

```razor
    <Breadcrumb RegexPathResolvers="@(new (Regex pattern, string label)[]{
                                      (new Regex("components$"), "Components"),
                                      (new Regex("components/breadcrumb$"), "Breadcrumb"),
                                      })"
                PathResolver="@(path => "Unexpected path")">
```

Which implementation method do you prefer?

PR #1860, opened 2021-08-23
fix(module:autocomplete): overlay is showing for AutoCompleteSearch

### LINKED ISSUE #1813: Wrong autocomplete dropdown placement
### Describe the bug

The dropdown for the lookup patterns is appended to the root element instead of the parent of the dropdown component and thus placed in the top left corner of the pages.

This bug can be seen in your documentation: [](https://ant.design/components/auto-complete/)
image

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1813

### 💡 Background and solution
1. `AutoCompleteSearch` wasn't properly communicating its reference to `Overlay`. 
2. Unification of approach for `AutoCompleteSearch` and `AutoCompleteInput`.
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/auto-complete/AutoComplete.razor
+++ b/components/auto-complete/AutoComplete.razor
@@ -53,16 +53,16 @@
         </Overlay>
         <Unbound>
             <CascadingValue Value="this" IsFixed="@true">
-                @if (ChildContent == null)
-                {
-                    <AutoCompleteInput RefBack="context" @bind-Value="@CurrentValue" Placeholder="@Placeholder" Disabled="@Disabled" Style="@Style" Id="@Id" />
-                }
-                else
-                {
-                    <CascadingValue Value="context" Name="OverlayTriggerContext">
-                        @ChildContent
-                    </CascadingValue>
-                }                    
+                <CascadingValue Value="context" Name="OverlayTriggerContext">
+                    @if (ChildContent == null)
+                    {
+                        <AutoCompleteInput @bind-Value="@CurrentValue" Placeholder="@Placeholder" Disabled="@Disabled" Style="@Style" Id="@Id" />
+                    }
+                    else
+                    {
+                        @ChildContent                    
+                    }
+                </CascadingValue>
             </CascadingValue>
         </Unbound>
     </OverlayTrigger>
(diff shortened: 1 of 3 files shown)


ElderJames:
The dropdown can't be closed in this demo.

auto-complete-cant-close

anddrzejb:
@ElderJames From what I see this bug was there before I made my change (demo site has it). I will try to have a look at this.

anddrzejb:
@ElderJames I had a look and it is another problem coming from `[Parameter]` not being a primitive. So in this case there is this `OptionDataItem` parameter which is of type `IEnumerable<AutoCompleteDataItem<string>>` used to generate list of autocomplete options. In `AutoComplete` component in component lifecycle event `OnParametersSet()` there is a call to a method `ResetActiveItem()` which shows the overlay. After the overlay is closed, the `OnParametersSet()` is called, because of [this (framework always treats the parameter set as changed when one or more complex-typed parameters are present)](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-5.0#after-parameters-are-set-onparameterssetasync-1). Which in turn forces the overlay to reopen. I honestly do not understand the `AutoComplete` to start playing with that part. To be honest I do not really understand what is the difference between `AutoComplete` and `Select` component (conceptually). 

One of the solutions is the one I mentioned in one of my previous comments regarding performance optimization - replace the parameter with a class that will be able to detect any changes to itself. Then add a check in whatever lifecycle method is used - either `OnParametersSet()` or `SetParameters()` to detect if any change really happened. And then act upon that detection.
Another solution - expose a change method and allow option list change only through that method. Ignore any changes to `[Parameter]` after initialization. This is less intuitive thus I do not like it. But it is simpler to implement.

PR #1863, opened 2021-08-23
fix(module:checkboxGroup): option change does not lock checkboxes

### LINKED ISSUE #1716: When CheckboxGroup Options changed，ValueChanged don't work
### Describe the bug

It works fine when CheckboxGroup Options keep not change。But if I Changed Options in code，ValueChanged don't work。

IF you click button，OnCheckboxChanged

@page "/"

<Button OnClick="@Test">test</Button>
<br />
<AntDesign.CheckboxGroup Options="@_weekOptions" ValueChanged="@OnCheckboxChanged" />
<Text>@TXT</Text>

        private CheckboxOption[] _weekOptions = new CheckboxOption[] {  new CheckboxOption {Label = "one", Value = "1"},
        new CheckboxOption { Label = "two", Value = "2" },

    };

        private string TXT
        {
            get
            {
                return string.Join(",", _ids.ToArray());
            }
        }

        List<int> _ids = new List<int>();
        /// <summary>
        /// 获取选中的日期
        /// </summary>
        /// <param name="weeks"></param>
        private void OnCheckboxChanged(string[] weeks)
        {
            _ids.Clear();
            if (weeks != null)
            {
                foreach (string s in weeks) _ids.Add(int.Parse(s));
            }
        }

        private void Test(Microsoft.AspNetCore.Components.Web.MouseEventArgs e)
        {
            _weekOptions = new CheckboxOption[] {
                new CheckboxOption { Label = "one", Value = "1" },
                new CheckboxOption { Label = "two", Value = "2" },
                new CheckboxOption { Label = "three", Value = "3" },
                new CheckboxOption { Label = "four", Value = "4" }
            };
        }
    }

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1716

### 💡 Background and solution
When `Options` is changing, an attached function that calculates `CheckboxGroup` current value should also be changed. This PR is forcing the change.

#### However
I see another issue here which is tightly connected to performance. Currently. because `Options` are `OneOf`, the reset of its value is done on basically any action done on the component (so clicking on a single checkbox in the group triggers the `Options` to be set again). This I think is due to the fact that `Options` is not a primitive type (as stated in #992). In this case though I do not think we can use a primitive type. What we could use is a class that will internally hold `CheckboxOption` array and will also internally monitor for the changes. My idea here is that such class could very quickly know if it changed and a simple comparison could be done to find out if a change really happen. Then, if it did, appropriate action could be called (like clearing current value of the options and reset the aforementioned function responsible for calculation).

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/checkbox/CheckboxGroup.razor.cs
+++ b/components/checkbox/CheckboxGroup.razor.cs
@@ -19,6 +19,10 @@ namespace AntDesign
             {
                 _options = value;
                 _isOptionDefined = true;
+                if (_afterFirstRender)
+                {
+                    _currentValue = GetCurrentValueFunc();
+                }
             }
         }
 


ElderJames (approved):

ElderJames:
> I see another issue here which is tightly connected to performance. 

Yes, we have many components that have this issue. Maybe we can solve them later.

anddrzejb:
> > I see another issue here which is tightly connected to performance.
> 
> Yes, we have many components that have this issue. Maybe we can solve them later.

We should aim at solving them before v1, as these changes will most of the times (if not always) be breaking changes.

ElderJames:
Yes, you are right.

Issue #1894, opened 2021-08-29
The default DateTime filter in Table conflicts with some third party library 


iamSmallY:
### Describe the bug

The new feature in #1864 conflicts with the third party library `Pomelo.EntityFrameworkCore.MySql`, since it can't be translated to the MySql statement.

### Steps to reproduce (please include code)

Bind the `DataSource` of the table to the DBSet and use the default filter.

### Exceptions (if any)

```
2021/08/29 19:15:05.669|WARN|::1|Microsoft.AspNetCore.Components.Server.Circuits.RemoteRenderer|Unhandled exception rendering component: The LINQ expression 'DbSet<SerialNumber>()
    .OrderBy(s => s.Sn)
    .Where(s => s.Expires - TimeSpan.FromMilliseconds((double)s.Expires.Millisecond) == __p_0)' could not be translated. Additional information: Translation of method 'System.TimeSpan.FromMilliseconds' failed. If this m
ethod can be mapped to your custom function, see https://go.microsoft.com/fwlink/?linkid=2132413 for more information. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly
 by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
2021/08/29 19:15:05.669|ERROR|::1|Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost|Unhandled exception in circuit 'HvUodM1SDEpQ-YrcWTpuspHYssThblABPbOKYSiQ9Jk'.
```

### Further technical details
- AntDesign Nuget Package version 0.9.3
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version

PR #1895, opened 2021-08-29
Change the implement of the compare operator of DateTime

iamSmallY:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#1894 

### 💡 Background and solution

The way of remove millisecond in #1864 is not supported by `Pomelo.EntityFrameworkCore.MySql` (maybe even more third party orm libraries), so I changed the implement of compare operator.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix the bug that the code of remove millisecond in #1864 is not supported by third party orm library.          |
| 🇨🇳 Chinese | 修复了#1864 中移除毫秒语句不被第三方 orm 库支持的 bug. |


anranruye:
AddMilliseconds method can be translated by ef core, you can use this.

iamSmallY:
I've tried, but it can't be translated by `Pomelo.EntityFrameworkCore.MySql` which I'm currently using.

I am not sure if we should be compatible with it

anranruye:
I have tried sql server and sqlite provider and this works. It's not a good news that 'Pomelo.EntityFrameworkCore.MySql' doesn't support this translation. Can you share your usage of 'AddMilliseconds'? And there are multiple mysql providers. Maybe you can also have a try with 'MySql.EntityFrameworkCore'

iamSmallY:
My usage is

```C#
s = _db.s.Where(x => x.DateTime.AddMilliseconds(-x.DateTime.Millisecond) == DateTime.Now)
```

and the error is

```
The error was: The LINQ expression 'DbSet<x>()
    .Where(s => s.DateTime.AddMilliseconds((double)-s.DateTime.Millisecond
    ) == DateTime.Now)' could not be translated. Additional information: Translation of method 'System.DateTime.AddMilliseconds' failed.
```

iamSmallY:
Since PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#1497 is merged, the new implement won't cause error with Pomelo.EntityFrameworkCore.MySql

ElderJames:
Nevertheless, we need to be compatible with the older version.

anranruye:
现在 Pomelo.EntityFrameworkCore.MySql 那边属于是合并了但是没发版吧，我觉得只要那边发版了就行，兼容旧版本只能说尽量吧，但是现在不排除其他数据库/orm还有兼容问题。

PR #1720, opened 2021-07-08
fix(module:select): allow change both data source and value in one render period

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Fixes #1719
Fixes #1536 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | allow change both data source and value in one render period |
| 🇨🇳 Chinese |           |

### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #1886, opened 2021-08-28
fix(module:select): fix clearing behavior, reduce ValueChanged triggered, change value after datasource changes, follow the datasource element order

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Bundle size optimization
- [x] Performance optimization
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

When a `Select` component:
    1. has `DefaultActiveFirstOption` or `DefaultValue` attribute
    2. inital value is not `null`
Then clearing is not not working as expected. When the first time clear, the selected value is set to `DefaultValue`/first option value

动画11

code:
```
<Select DataSource="@_list"
        @bind-Value="@_selectedValue1"
        DefaultActiveFirstOption
        ValueName="@nameof(Person.Value)"
        LabelName="@nameof(Person.Name)"
        DisabledName="@nameof(Person.IsDisabled)"
        Style="width:120px"
        AllowClear
        OnSelectedItemChanged="OnSelectedItemChangedHandler">
</Select>

<Button OnClick="()=>_selectedValue1 = null">Set value to null</Button>
@code
{
    List<Person> _list;
    string _selectedValue1 = "lucy";
    class Person
    {
        public string Value { get; set; }
        public string Name { get; set; }
        public bool IsDisabled { get; set; }
    }

    protected override void OnInitialized()
    {
        _list = new List<Person>
        {
            new Person {Value = "jack", Name = "Jack"},
            new Person {Value = "lucy", Name = "Lucy"},
            new Person {Value = "disabled", Name = "Disabled", IsDisabled = true},
            new Person {Value = "yaoming", Name = "YaoMing"}
        };
    }

    private void OnSelectedItemChangedHandler(Person value)
    {
        Console.WriteLine($"selected: ${value?.Name}");
    }
}
```
-------------------------------------------------------------------------------------------
2. ValueChanged is triggered multiple times, which is unnecessary.
动画111
code:
```
<Select DataSource="@_list"
        TItemValue="int?" TItem="Person"
        Value="@_selectedValue1" 
        ValueChanged="@((v) => { Console.WriteLine("ValueChanged:" + v); _selectedValue1 = v; })"
        ValueName="@nameof(Person.Value)"
        LabelName="@nameof(Person.Name)"
[...]

Issue #1587, opened 2021-06-01
SelectOption bug when Value = 0


datnguyen89:
Hi team!
I get a bug when using a select input and value of SelectOption = 0 

```
    <Select @bind-Value="@_selectedValue2"
            Style="width: 120px;"
            TItemValue="int"
            TItem="string">
        <SelectOptions>
            <SelectOption TItemValue="int" TItem="string" Value=@(0) Label="zero" />
            <SelectOption TItemValue="int" TItem="string" Value=@(1) Label="one" />
            <SelectOption TItemValue="int" TItem="string" Value=@(2) Label="two" />
        </SelectOptions>
    </Select>

    @code {
      private int _selectedValue2 = 0;
    }
```

Further technical details
- AntDesign Nuget Package version 0.8.1
![ant-blazor-select-bug](https://user-images.githubusercontent.com/55984933/120268960-10c17780-c2d1-11eb-86df-1ef6f5623242.gif)

anranruye:
Thank you for contacting us @datnguyen89 .

I think this is a bug. But it's not easy to solve this issue. Whether you use `AllowClear` or not, the `Select` component has a state that no select option is selected. Currently we use `TItemValue` default value for the `Value` property of `Select` component if no item is selected. When work with not nullable types, it's hard to judge whether there is a selected item or not if `Value` is '0'.

Use nullable int to workaround:
```
    <Select @bind-Value="@_selectedValue2"
            Style="width: 120px;"
            TItemValue="int?"
            TItem="string">
        <SelectOptions>
            <SelectOption TItemValue="int" TItem="string" Value=@(0) Label="zero" />
            <SelectOption TItemValue="int" TItem="string" Value=@(1) Label="one" />
            <SelectOption TItemValue="int" TItem="string" Value=@(2) Label="two" />
        </SelectOptions>
    </Select>

    @code {
      private int? _selectedValue2 = 0;
    }
```

anddrzejb:
@anranruye Just beat me to the answer. What I was thinking though is to maybe during evaluation of `Value == default` include the `AllowClear`. So if clearing is not allowed and `default` is in the `SelectOptions` then do not clear.

anranruye:
@anddrzejb Sounds good to me. Then what will happen if clearing is allowed? I think it's better to always select the `SelectOption` which matches the `Value` if there is one, even if after click the clear button.

anddrzejb:
[...]

PR #1906, opened 2021-09-01
fix(module:select): value no longer reset on datasource set

### LINKED ISSUE #1536: Immediately Changing Value after item is modified in Select
### Describe the bug

```csharp
RuleGroup ruleGroupToSelect = null;
ruleGroupToSelect = new()
{
    GroupName = this.NewGroupName
};
this.RuleGroups.Add(ruleGroupToSelect);
this.SelectedGroup = ruleGroupToSelect;
```

Where `this.SelectedGroup` has already bound to `Value` of `Select`. `SelectOptions` loop through the list `this.RuleGroups`. After the last assignment, `Select` shows nothing.

Do something as described in the code above.

### Exceptions (if any)

no.

### Further technical details
- AntDesign Nuget Package version - **0.8.0**
- Include the output of `dotnet --info` - **.NET 5.0**
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version - **not relevant**.

### LINKED ISSUE #1719: Select can not set datasource and bind-Value at the same time
### Describe the bug
when I change Select DataSource then I change bind-Value，the bind-Value can't work。

[Index.razor.txt](https://github.com/ant-design-blazor/ant-design-blazor/files/6781124/Index.razor.txt)

### LINKED ISSUE #1905: Select: Delayed DataSource resets Value
### Describe the bug
If `@bind-Value` parameter is set and `DataSource` is set in for example `OnInitializedAsync()` lifecycle event, then the value is reset.

Use case: 2 `Select` components, when first `Select`'s component `DataSource` depends of some value delivered to the host component & second `Select`'s component `DataSource` depends of value of the first `Select`. 
```razor
<Select DataSource="@_list1"
		@bind-Value="@_model.First"
		ValueName="@nameof(Person.Value)"
		LabelName="@nameof(Person.Name)"
		Style="width:120px"
		OnSelectedItemChanged="OnSelectedItemChangedHandler">
</Select>
<br/>
<Select DataSource="@_list2"
		@bind-Value="@_model.Second"
		ValueName="@nameof(Person.Value)"
		LabelName="@nameof(Person.Name)"
		Style="width:120px"
		OnSelectedItemChanged="OnSelectedItemChangedHandler">
</Select>

@using System.Text.Json
@code {
    class Model {
        public string? First { get; set; }
        public string? Second { get; set; }
    };
    record Person(string Value, string Name);

    string _keyValue = "A"; // <= simulated value delivered to the component
    List<Person> _list1,_list2;
    Model _model;

    {
        //simulate model being passed to the host component (retrieved from a DB)
        _model = new() { First = "lucyA", Second = "jacklucyA" };            
    }

    protected override async Task OnInitializedAsync()
    {        
        await SetList1(); // <= simulate Api retrieval that depends on a parameter _keyValue        

        //here, model is reset
        Console.WriteLine($"Model data after loading _list1: {JsonSerializer.Serialize(_model)} "); 
        if (_model.Second is not null) 
        {   //never gets here
            await SetList2(); // <= simulate Api retrieval that depends on the model
            Console.WriteLine($"Model data after loading _list2: {JsonSerializer.Serialize(_model)} ");
        } 
        Console.WriteLine($"Model data at the end: {JsonSerializer.Serialize(_model)} ");
    }

    private async Task SetList1()
    {     
        await Task.Delay(100); // <= simulate Api call
        _list1 = _list2 = GetList(_keyValue);
    }

    private async Task SetList2()
    {
        await Task.Delay(100); // <= simulate Api call
        _list2 = GetList(_model.First);
    }

    private static List<Person> GetList(string key)
    {
        return new List<Person>
        {
            new("jack" + key, "Jack"+ key),
            new("lucy"+ key, "Lucy"+ key),
            new("yaoming"+ key, "YaoMing"+ key)
        };        
    }

    private void OnSelectedItemChangedHandler(Person value)
    {
        Console.WriteLine($"selected: ${value?.Name}");
    }
}
```
#### Expected result:
image

#### Current result:
image

### Further technical details
- AntDesign latest commig
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version: VS 2019 Preview 16.11.2

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Fixes #1719 
Fixes #1905 

### 💡 Background and solution
I found out that issue #1905 was basically caused by PR #1720. I did a deeper dive and I found out that the solution employed by @anranruye was causing additional re-render and resetting of the bound value. The problem was that when `DataSource` changes, then `OnValueChanged` is called which is trying to find current `Value` in the `SelectOptionItems`. If nothing is matching in `SelectOptionItems`, then the `Value` is set to `default`. The thing is, `SelectOptionItems` might not yet be populated with `DataSource`. Thus, I added populating of the `SelectOptionItems` if `DataSource` change is detected (and added a `boolean _alredyRecreatedSelectOptions` to prevent multiple recreation of `SelectOptionItems` in `CreateDeleteSelectOptions()` method. I also rolled back the changes of PR #1720.

I also added the tests to cover the issues. 

### 📝 Changelog
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


components/select/Select.razor.cs
```diff
@@ -112,6 +115,12 @@ public IEnumerable<TItem> DataSource
                         OnDataSourceChanged?.Invoke();
 
                         _datasource = value;
+                        if (_optionsHasInitialized)
+                        {
+
+                            CreateDeleteSelectOptions();
+                            _alredyRecreatedSelectOptions = true;
+                        }
```
anranruye, line 123: `hasChanged = !value.SequenceEqual(_datasource);` is not enough to detect the datasource change. If we always modify the same datasource object(by for example List<T>.Add() and List<T>.Remove()), then `hasChanged` is always `false`.

components/select/Select.razor.cs
```diff
@@ -488,6 +497,14 @@ public override TItemValue Value
...
                 {
                     _selectedValue = value;
+                    if (_isInitialized)
+                    {
+                        OnValueChange(value);
+                        if (Form?.ValidateOnChange == true)
+                        {
+                            EditContext?.NotifyFieldChanged(FieldIdentifier);
+                        }
+                    }
```
anranruye, line 507: I don't understand why you write code in the property setters. As I remember, the setters are called depend on the property handwriting order. If the order of code execution matters, then you should not write code in the property setters. Here we want the code in the `Value` setter to be called after the code in the `DataSource` setter, right?

anranruye:
Notice that #1536 and #1719 are different. One is using DataSource approach, another one is using SelectOption approach. (I asked @ElderJames to reopen #1536, which is not actually solved by #1720) Please make sure you can solve both.

anranruye:
I think there is indeed a bug when DataSource changes, as I mentioned in #1886
图片
The second `Select` doesn't select any option, but the value is not set to `null`.

However, I can't see anything wrong with #1905. You are using both `OnInitialized()` and `OnInitializedAsync()` (why don't you also get the `_model ` object after a delayed task?) so the page is rendered twice, before and after you get the datasource items. The first time, values of `_model` are not null, but the datasources are null, so the model values are set to null.

anranruye:
However, I do get your expected result with 0.8.2 version(#1720 is merged since 0.8.3), I don't know why. Can you tell me what is the difference between putting the code in the property setter and putting the code in OnParametersSet method?

ElderJames:
> @ElderJames can you please remove #1536 from the list of issues this PR should solve? I incorrectly added it there and my privileges do not allow me to remove it by myself.

I can't remove it directly either, but modifying the content of the pull request does. Maybe this is a new mechanism in GitHub.

anranruye:
before you reply, I tried to avoid the first call to `OnValueChange` and got the expected result. I think your code can get the same effect though I haven't download and run your code.

But when I try to solve the issue I mentioned above together(by call On ValueChanged when DataSource changes, I still avoid the first call in this case), things went wrong. Can you try to solve this issue(value not reset to null when DataSource changes) together?

anranruye:
Your use case makes sense and I also have some use case like that(however, I only have two Select in the same time and I get the model after the select DataSource).

In fact, I have another idea. I think we should 
always remain the model value, no matter whether it has a corresponding select option or not. the model value should only be changed when DataSource changes(of course we may need to ignore the first(or also the second?) change when initializing) or the user select an option by ui. Changing the Value property by code only affect the ui, but this value will not be rewritten by the Select component even if it is not valid.

anddrzejb:
@anranruye I have mixed feelings about your proposal. It seems to me a bit inconsistent. Your suggestion is to:
- not change the bound value when for example an element does not exist in the `DataSource` - as far as I can see this can happen only when an item was removed from the `DataSource` 
- change the bound value when the `Datasource` changes - what about during initialization? What about if new `DataSource` actually has the current value?

What I do not like is the idea that a ui library takes a decision that should be left to the developer. Instead, I would actually advocate for additional parameter (an enum) that would allow the consumers to choose:
- reset value if not present in datasource (applies to situation when an item for the `DataSource` is removed or when the `DataSource` is changed)
- always reset value on `DataSource` change
- reset value on `DataSource` change if value not present in the `DataSource`

Or we never change (regardless) and tell the people to actually make their adjustment in `OnDataSourceChange`. Because my change now also triggers when an item is removed from the `DataSource`, that should cover the aforementioned scenarios as well as any new scenarios. We would just shift the responsibility to the consumers & possibly keep the `Select` component a bit simpler.

anranruye:
@anddrzejb I think you are right, and your suggestions are both not bad choices(In fact, I am doing something like that).

But let me explain why my idea seems a bit inconsistent.
Consider two situations:

Select options are hard-coded at the client side. "1","2","3" are available. At the beginning, this works well. But as time goes on, the business develops, "4" and "5" are also valid values. Then some clients update, some clients not. When edit an item on a not updated client, we may get model value "4" from the server, but it is not in the data source("1", "2", "3"). Here we should decide to remain the value "4" or set the model value to null. I think still using "4" as the value is better, and for other situations where the model value(especially when the model is already in db) has a conflict with the select options, we should respect the model first.

Another situation is the typical province - city cooperating I showed above. Consider we have "A" province with its two cities "A1" and A2, and B province with B1 and B2. City field is not required. If we select A-A1 first, then change to 
"B-?", we expect to send "B-null" to the server but the actual result is "B-A1". City is not set to null and even worse, A1 doesn't belong to B.

I like the idea that we never change, just let the user to make their own decision. But there are also some developers who are not against that we implement the most common situations. If we want to do this, then we need to consider the situations I mentioned above.

anranruye:
Maybe we can follow these rules to cover the most common usage:
1. When change Value(by change the variable bound to Value parameter with C# code) only, do not modify the bound variable back even if it is not present in datasource, because this value is explicitly set by the user and we should respect this value.
2. When change Datasource only, if the bound variable value is not present in previous datasource and not present in the new datasource, do not change bound variable; if the bound variable value exists in previous datasource and not present in the new datasource, change bound variable to null/default.
3. When change Value and Datasource both, I'm not sure, I prefer the same behavior with 1 (?) for the same reason, the value is set explicitly by the user/developer.

components/select/Select.razor.cs
```diff
@@ -84,33 +86,29 @@ public IEnumerable<TItem> DataSource
...
                     if (_datasource == null)
                     {
-                        hasChanged = true;
+                        _dataSourceHasChanged = true;
                     }
                     else
                     {
-                        hasChanged = !value.SequenceEqual(_datasource);
+                        _dataSourceHasChanged = !value.SequenceEqual(_datasource)
+                            || (_optionsHasInitialized && SelectOptionItems.Count != _datasource.Count());
```
anranruye, line 106: I don't think we can rely on the item count, and `SelectOptionItems` also contains added tags?

anddrzejb:
I guess here we should as @ElderJames for his opinion. To be honest I think I prefer my last proposal - shift the responsibility to the consumer, do not force the behavior on the consumer. So basically, follow your proposal in 1 & 3. Which in turn would also mean that we apply same logic across the board - we don't change in 2 either.

components/select/Select.razor.cs
anddrzejb: I forgot about the tags. Indeed it does, but it can be simply fixed by subtracting what we have in `AddedTags`.
Count gives a basic check over the basic add/remove. It obviously won't work when a consumer adds & remove the same number of items from the `DataSource`. From what I can see, currently the whole sequence comparison in `DataSource` setter is useless anyway. We could just replace it with reference equality check and I think it would work the same (except probably faster).
But, we should probably keep the `SequenceEqual` check but not to check it against `_dataSource` but against a deep copy of the `_dataSource`. I think then we could safely rely on the change detection.

anddrzejb: Sorry, I think shallow copy should be good enough.

anddrzejb: My tests revealed that shallow copy wont cover all scenarios (given the `DataSource` is an `IEnumerable<Persons>`):
```csharp
        //will NOT detect
        _persons[2].Name = newValue; 
...
        //will detect
        _persons[2] = new Person { Id = 3, Name = "new Person()"  }; 
...
        //will detect
        _persons.Remove(_persons.Last());
        _persons.Add(new Person { Id = 3, Name = "removed and added" });
```
But we cannot really make a deep copy ourselves (possible issues: cycle object graph, references to external objects, events). So, to cover all scenarios, the only other way I can think of is to allow the consumers to pass their own `IEqualityComparer` - I would think these are going to be the advanced scenarios, but still available to the developers.

anranruye: It seems we are breaking this use-case: https://ant-design-blazor.gitee.io/zh-CN/components/select#components-select-demo-TrackItemChanges. Now a custom IEqualityComparer is necessary to track item changes.

anddrzejb: Thanks, I will have a look. 

I also experimented with `IEqualityComparer` but I think it won't really help us here, because in case of changes to an item of the `DataSource` we still are making changes to the same object. So the comparer compares an object to itself. Now I start to think that we should expose our own interface/base class for `DataSource` that will have a `Clone` virtual method, which will do a shallow copy by default but with ability to overload it, so the consumers can create a deep copy implementation. What do you think?

anranruye:
I think you are right. We should apply same logic for all situations and that will make everything simple.

If we decide to do this, then there comes an issue: how should we use `DefaultValue` and `DefaultActiveFirstItem`?

components/select/Select.razor.cs
```diff
@@ -84,33 +86,29 @@ public IEnumerable<TItem> DataSource
                     SelectedOptionItems.Clear();
 
                     Value = default;
```
anranruye, line 88: I don't think we should change the value inside DataSource property. As I mentioned, the order to set `Value` and `DataSource` depends on user hand-writing. And this is actually unnecessary.

anddrzejb:
@ElderJames What do you think about our changes proposal to the value? I do not want to start doing something that you might possibly reject later on.

@anranruye Regarding default behavior - I believe that default value should kick in only during initialization or when clear button is clicked. If we keep the value even if it does not exist in `DataSource`, we should visually reflect it by not showing anything. And possibly describe in detail the `Select`'s behavior in the docs.

components/select/Select.razor.cs
anddrzejb: If our proposal of removing value changes on `DataSource` changes gets accepted, then this will most likely be removed from the setter.

anranruye:
I do know that DefaultValue/DefaultActiveFirstOption is used only once during initialization. That's the problem. I think people who use DefaultActiveFirstOption want to see the first option is selected after DataSource changes(and the previous value is no longer valid). But this break the rule that we don't change the Value.

To be honest, I don't like DefaultValue/DefaultActiveFirstOption at all(and I will never use them). If we want the users to take care the Value which they want, as we discussed, then they should manully set their expected value to `Value` property, rather than DefaultValue property.

anddrzejb:
I wholeheartedly agree with you. `DefaultValue` does seems to me like an overkill, especially that it is anyway loading `Value` with `DefaultValue`. I do however see an appeal in `DefaultActiveFirstOption`. Still, I think we should apply these only during initialization and on clear. Otherwise, keep the original values. Regarding the people who would be negatively impacted by our proposed change - I think we should strive to make the library to behave consistently.

anranruye:
> it is anyway loading Value with DefaultValue
Though our docs says that, it is not true. `DefaultValue` is only applied when the value is not present in the DataSource. Otherwise, Value property is used.

>  I think we should apply these only during initialization and on clear
Basiclly I agree.
image
The doc doesn't mention "Clear" but it mentioned "Reset button within Forms", do you think "Clear" and "Reset" are the same? I think `DefaultValue` can be used when "Clear", but may be not when "Reset".

And I noticed that currently `DefaultValue` is not used on clear(it is used when the value is not present in the DataSource and `AllowClear` is false). 

By the way, @ElderJames agrees with us. We should change the value only when the users select an option or click clear button.

anddrzejb:
Well, form reset is a separate method in `Select` that still just calls `ClearSelectedAsync()`. To be honest, I do not see much of a difference (conceptually) between clearing a select manually or resetting a form. As far as I can say, I would expect them to behave identically. 

What I meant by 
> it is anyway loading Value with DefaultValue  

is that when `Value` is not set but `DefaultValue` is set, then `Value` gets whatever is in `DefaultValue`. So, in essence, `Value = DefaultValue`. Which in turn does not seem to make `DefaultValue` have any added functionality on top of what `Value` has. However I actually think that we still may need `DefaultValue`. The use case is actually the clearing of the `Select`. If we imagine the following scenario:
1. `DefaultValue` set to 10.
2. `Value` set to 3 (either during initialization or later by the user)
3. Clear button is pressed.
4. Because we have `DefaultValue`, we know we should reset it to 10. Without it - we would have only value default.

Of course this could be easily handled in `OnValueChange`. But this does look like a nice functionality.

anranruye:
1. For a form create a new item, "reset" and "clear" are same. For a form which edit an existing item, there is one difference. "clear" will set the values to null/default/DefaultValue. while "reset" will set the values to original values obtained from db(we can implement this in the future and use the same implementation temporary).
2. when clear, I think the `DefaultValue` should be only used for not nullable types. for nullable types, it's reasonable to always set the value to null.
3. maybe we can display a placeholder for these selected values which is untracked in the DataSource.
4. of course the form should validate the value in the model, which will be actually send to the server.

anranruye (approved):

anddrzejb:
@ElderJames Please wait with merging, I have pending changes to push.

@anranruye 
>     1. For a form create a new item, "reset" and "clear" are same. For a form which edit an existing item, there is one difference. "clear" will set the values to null/default/DefaultValue. while "reset"  will set the values to original values obtained from db(we can implement this in the future and use the same implementation temporary).

I believe that `DefaultValue` is actually a good place to put the "obtained from DB" part. In case if the original value needs to be obtained during the reset, the `OnValueChange` action can be used.
 
>     2. when clear, I think the `DefaultValue` should be only used for not nullable types. for nullable types, it's reasonable to always set the value to null.

I do not agree. I feel this makes the lib behave inconsitently. In my opinion we should either always reset to `DefaultValue` or never. 

>     3. maybe we can display a placeholder for these selected values which are untracked in the DataSource.

Then probably we should also show a placeholder in default mode. I am afraid this idea leaks the dev concerns into user space. Also complicates things 😒 .

>     4. of course the form should validate the value in the model, which will be sent to the server when submit the form.

Yes, but we can have in the form `Select` value = 4 that does not exist in the `DataSource`. Let's say we have a `[Required]` attribute on the model. Validation will pass even though visually it will look like the `Select` is empty.

anranruye:
> I believe that DefaultValue is actually a good place to put the "obtained from DB" part. In case if the original value needs to be obtained during the reset, the OnValueChange action can be used.

I agree.

> I do not agree. I feel this makes the lib behave inconsitently. In my opinion we should either always reset to DefaultValue or never.

then let's always reset to DefaultValue

> Then probably we should also show a placeholder in default mode. I am afraid this idea leaks the dev concerns into user space. Also complicates things 😒 .

Yes, I mean that. the placeholder is for both default mode and muiltiple mode

> Yes, but we can have in the form Select value = 4 that does not exist in the DataSource. Let's say we have a [Required] attribute on the model. Validation will pass even though visually it will look like the Select is empty.

so it's better to display a placeholder to tell the user the Select is not empty, though it makes things a bit complicated.

anranruye:
> I do not agree. I feel this makes the lib behave inconsitently. In my opinion we should either always reset to DefaultValue or never.

I just realized an issue, if we always reset to DefaultValue, then how can the user set the value to `null`?

anranruye:
maybe we can display two buttons for "reset" and "clear"

anranruye:
maybe we can display two buttons for "reset" and "clear"

anddrzejb:
> I just realized an issue, if we always reset to DefaultValue, then how can the user set the value to `null`? shall we encourage the users to use `null` as a value, like #1783? I don't think this is a good idea.

I was rather thinking to have a detection that `DefaultValue` was actually set (I think we have something similar in `DatePicker`). If it is not set, then clearing would just set to type default value. Otherwise - to `DefaultValue`. 
 
> In fact, we were talking about "reset", not "clear". I think the meaning of word "clear" is clear - set the value to nothing(in C#, it's represented by `null`). I agree that `DeafultValue` is a good place to hold the value which will be used when "reset". However, when clear button is clicked, what is the user expected, "clear" or "reset"? not certain. I guess they are both required by different people.

Here I believe it is less about what users understand and more about what developers want to achieve. If developer would like clear button to actually clear to null value - then either set `DefaultValue` accordingly or don't use it at all and use `Value` with appropriate first value. As a last resort, there is always `OnValueChange` event. I am not sure about introducing additional reset button. I think it might be more confusing to the users. Or actually be a source of accidental clicks.

> Anyway, we can only use the clear button for one purpose. If we decide to use it as "reset button", then we'd better change the terminology “clear button” and the property name "AllowClear", which are misleading.

What we could do is expose an `EventCallback` that will fire before clearing, that will accept a boolean. Then the dev can attach to it, run the preferred logic and either return true to continue with clearing or false to stop built in clearing logic.

anranruye:
Of course the users can decide whether to use DefaultValue or not. But the issue still exists, when DefaultValue is set, then the final user can not set the value to `null`, unless it is in the DataSource.

anranruye:
And how about if the developer wants both?

A common senorio is, a clear button is displayed in every single form component(for "clear"), and the form also has a reset button which will reset the whole form.

anddrzejb:
I get your point. If developer would like to have both  default value and type default value, then there is no way of doing it without additional button. In this light I do see the use for clear & reset buttons.

anranruye:
I asked @ElderJames about his opinion and he said, the current clear button is only used for "clear", not "reset". If the developers want to reset, they should call the Reset mothod of form or input component. (We can always add a ui button to reset, but now there isn't one).

There is another thing we also feel difficult: is it allowed to clear a Select component whose `TItemValue` is not nullable?

At the beginning we just consider `default(TItemValue)` and `null` the same, and expect `default(TItemValue)` to play the role of `null`. But things are not as what we want. People often use type default value(like 0, false) as option value(#1587). It is still awful when people don't use them as option value, because the value 0/false will be sent to the server if the validation rules don't consider them invalid.

anddrzejb:
As I stated before, `Reset` method in `Select` component does exactly the same as `Clear` button:
```csharp
        /// <summary>
        /// Called by the Form reset method
        /// </summary>
        internal override void ResetValue()
        {
            _ = ClearSelectedAsync();
        }
```
So at this point there is no support for different behavior depending on clear or reset. As far as I understand, we were discussing an independent logic for clear and for reset.

Regarding `TItemValue` being not nullable - again, I would advocate for consistency. When `TItemValue` is nullable, we actually call for type default. Should we break that and offer a different behavior for non-nullable? In case of mentioned #1587, shouldn't we always try to match entered value with existing `TItemValue`? I believe we should. Of course, if there is a select option that is equal to `default(TItemValue)` then the `AllowClear` should probably be ignored. Or otherwise it will work like reset with `DefaultValue` set to `default(TItemValue)`.

anddrzejb:
@anranruye , @ElderJames My latest commit 1a67ca3 this time actually covers #1536. Changes in the commit:
- Better detection of changes made to `DataSource` - using single (for primitive types in the `DataSource`) or double (for reference types) shallow copy. It is achieved using reflection - but to improve performance, the `MemberwiseClone` method is fetched only once. Also, shallow copy is done only if change detection discovers differences. Otherwise, I do not think we should have any performance hit here - the `SequenceEqual()` was there before. 
- This change also includes new `[Parameter] DataSourceEqualityComparer` that can be passed to the component for more advanced scenarios. I believe this change also allows to avoid rebuilding of the `SelectOptionItems` on every `OnParametersSet()` execution. Currently it is a `[Parameter]` but to be honest I would rather this be a public property accessible through instance (same logic as with `Form`) - this is for performance reasons. We should avoid non-primitives in `[Parameter]`. Also, I think this is an advanced scenario, so it should not be a problem for the consumers.
- New tests for change detection:
    - when item in the `DataSource` is replaced (`datasource[1] = new Item()`)
    - when item's label is changed in the `DataSource` (`datasource[1].Label = "Changed label"`)
- Fixes to #1536: the problem here was (and kind-of still is) that when both `Value` & new select option is added at the same time, first what gets set is the `Value`. Because the `SelectOption` is not yet added (first parent renders then its children), the newly added `Value` is reset do default. Until we implement what we discussed here (that is - do not change the `Value` if it does not exist in the `DataSource` or `SelectOption`) the `LastValueBeforeReset` is used to actually revert to value that was existing before reset, but only if new `SelectOption` is added. 
- New tests to support issue #1536 resolution (for `default` & `multiple` scenarios)
- New tests to support checking that value does not reset on `DataSource` changes. 4 out of 6 tests are currently commented out, because they are not passing now. I decided to leave them for now, so hopefully the can be reinstated by anyone who is going to implement what we discussed in this PR.

anranruye:
> As I stated before, Reset method in Select component does exactly the same as Clear button:

> So at this point there is no support for different behavior depending on clear or reset. As far as I understand, we were discussing an independent logic for clear and for reset.

Yes, we should make `ClearSelectedAsync` do the "clear" job and add new method(s) for "reset".

> Regarding TItemValue being not nullable - again, I would advocate for consistency.

I don't like Inconsistency, either.

> In case of mentioned #1587, shouldn't we always try to match entered value with existing TItemValue? I believe we should.

I think so.

> Of course, if there is a select option that is equal to default(TItemValue) then the AllowClear should probably be ignored. Or otherwise it will work like reset with DefaultValue set to default(TItemValue).

That's the problem. Again, I want to emphasize, "clear" is "clear", it's not "reset" or anything else. for Input/InputNumber/TextArea, it's ok to set the value to `0` (for not nullable types). but this is not true for `Select` component. for not nullable types, anything we do will not be "clear". 

The case is, C# has a different type system with Js. Js variables are always nullable. Js ui technologies don't have this issue, but here we have. My suggestion is to forbid "clear" for not nullable types.

> When TItemValue is nullable, we actually call for type default. Should we break that and offer a different behavior for non-nullable?

Yes, I think we should break this for non-nullable types. My answer is to forbid "clear" for not nullable types. Instead, as I mentioned above, we can display a independent reset button.

 @ElderJames doesn't decide yet. If you agree, maybe he will agree too.

anddrzejb:
I think we agree. What I would add is a compiler warning when non-nullable type is used that would contain a link to detailed explanation of the problem to the docs.

components/select/Select.razor.cs
```diff
@@ -84,39 +119,52 @@ public IEnumerable<TItem> DataSource
...
-
                         _datasource = value;
+                        if (IsTItemPrimitive)
+                        {
+                            _dataSourceShallowCopy = _datasource.ToList();
+                        }
+                        else
+                        {
+                            var cloneMethod = GetDataSourceItemCloneMethod();
+                            _dataSourceShallowCopy = _datasource.Select(x => (TItem)cloneMethod.Invoke(x, null)).ToList();
```
anranruye, line 153: When use default DataSourceEqualityComparer and TItem is not primitive, _dataSourceHasChanged is always true and OnDataSourceChanged will be triggered.

ElderJames:
> Yes, I think we should break this for non-nullable types. My answer is to forbid "clear" for not nullable types. Instead, as I mentioned above, we can display a independent reset button.

I agree with you. 
However, if the user still needs Clear, I suggest adding another property, such as `ValueOnClear`, to allow the user to specify which value to change.

components/select/Select.razor.cs
anddrzejb: Please check commit af8d541 - I added internal `DateSourceEqualityComparer` that is comparing only values of label & value properties of the `DataSource`. Also, I included a test to cover the scenario.

anranruye (approved):

anranruye:
@ElderJames Ok. Though this will confuse the concept of "clear" and will bring inconsistency(we should always set the value to null for nullable types), it meets the actual usage habits of users.

ElderJames (approved):

ElderJames:
@anddrzejb @anranruye Thank you for the great works.

anddrzejb:
@ElderJames  @anranruye Thanks for a fruitful discussion. I hope we will manage to implement the proposed changes.

Issue #1283, opened 2021-03-30
Input lags and skips characters


HansenBerlin:
Hi,
first things first: thanks for your outstanding work, I really enjoy working with ant blazor. I am using this template, .net 5, VS19 Enterprise, blazor server and the app gets deployed to an azure app service. Locally (IIS) everything works like a charm, when deployed I face the same issue that became fixed with #579: On form fields, input fields etc. when typing I have a strong lag, sometimes characters are skipped and reappear. In the network waterfall i can see that for every character thats typed a request is sent to the server, which should not be the case IMO.
I tried to find the option to enable optional key debouncing but could not find it.
Would be nice if you could help, as this is the only issue I am facing but its a gamebreaker. Except this everything works like a charm!
Hannes

Hona:
Hi, @HansenBerlin.

First off thank you for using Ant Design Blazor!

Due to the way inputs work, the string that is 2 way-bounded from the input form gets updated from the string to the text box.

In a Blazor Server scenario, with latency being a thing (imagine 200ms for an average request), if you send multiple updates to the server, they will respond back to you with state, but you already typed more.

One way to fix this is to set the debounce time to be 1 or 2+ seconds - such that with slow internet no request will be sent while typing is happening.

If you are using an enter key or button to submit/search etc - you may use an absurdly large debounce time, as the `onblur` event will cause an update (think tabbing out). 

Feel free to play around with several configurations like what I send below, to see what works best in your situation. Remember that Blazor Server by design is different to client-side frameworks like React/Angular etc.

Also note with these examples, the same debounce property applies to the `Input` component.
```html
<!-- Note that this defaults to a debounce time of 250ms, I use this for on intranet instances, which have a typical latency of under 20ms -->
<Search EnterButton="true" @bind-Value="@_search" OnSearch="@System.Console.WriteLine(_search)"/>

<!-- This is a good middle ground allowing for requests to go around the world with a bit of processing time, -->
<Search DebounceMilliseconds="750" EnterButton="true" @bind-Value="@_search" OnSearch="@System.Console.WriteLine(_search)"/>

<!-- If you still have issues, test it with an absurdly long time and let us know if it still works -->
<Search DebounceMilliseconds="10000" EnterButton="true" @bind-Value="@_search" OnSearch="@System.Console.WriteLine(_search)"/>
@code {
    private string _search;
}
```

Hope this helps you figure out the best solution for you!

Hona:
[...]

PR #1917, opened 2021-09-06
Add stop propogation

Hona:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

N/A

### 💡 Background and solution

Events propogating cause many UI refreshes, and general lag - especially in forms/modals

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


Hona:
Test please @ElderJames

ElderJames:
Thanks @Hona ,I don't quite understand what problem this is solving?

Hona:
When propagation is normal it will send the event to all parent components causing quite a bit of lag. See related #1283

ElderJames:
What I don't understand is why only these two events were stopped?

Hona:
OnChange runs alot: (when there is no debounce, it will happen on every key press, or every X ms depending on the value). On Blazor Server applications this is very visible

ElderJames (approved):

PR #1920, opened 2021-09-06
fix(module:textarea): add rows parameter

### LINKED ISSUE #1869: Texarea bug autosize and minrows
When I using AutoSize and MinRows at once => MinRows not worked

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1869

### 💡 Background and solution
1. Added `Rows` parameter. This is to cover the requirement of the issue & to align with antD. Defaults to 3, which is approx of current default of antD (in antD case it looks like the default to about 2.8).
2. Minor typescript optimizations (explicit types)
3. For `AutoSize` added style `resize:none` (align with antD) that turns-off ability to manually resize the `TextArea`. This also could be an additional parameter, but antD does not have it.
4. Debug warning if consumer tries to use either `MinRows` or `MaxRows` and set `AutoSize = false` (with automatic switch to `true`)
5. Included PR #1917 changes also in the `TextArea` component
6. Added comments, rearranged new parameter properties to keep alphabetical order (my ocd...)
7. Remove from `TextArea` unnecessary (IMO) conversion of passed value to `string`
8. Docs fixes
9. Fixed and added tests for `TextArea` to evaluate proper height calculation (only bUnit, typescript tests will have to wait for #1848 to be merged)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Added parameter `Rows`.  |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/input/TextArea.razor
+++ b/components/input/TextArea.razor
@@ -1,6 +1,5 @@
 ﻿@namespace AntDesign
 @inherits Input<string>
-<!--TODO: minheight, maxheight, onResize-->
 
 @{
     Dictionary<string, object> attributes =
@@ -19,7 +18,7 @@
             { "style", Style },
             { "class", ClassMapper.Class },
             { "disabled", Disabled },
-			{ "readonly", ReadOnly },
+			{ "readonly", ReadOnly },            
          };
 
     if (Attributes != null)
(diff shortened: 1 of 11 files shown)


ElderJames (approved):

components/input/TextArea.razor
```diff
@@ -34,13 +33,13 @@
 		@if (Suffix != null)
 		{
 			<span class="@_warpperClassMapper.Class">
-				<textarea @ref="Ref" @attributes="attributes"/>
+				<textarea @ref="Ref" @attributes="attributes" @onclick:stopPropagation="@StopPropagation" @onblur:stopPropagation="@StopPropagation"/>
```
Hona, line 36: Did you mean to do @onchange:stopPropagation=... @anddrzejb?

anddrzejb: Wow...amazing...Can't even copy right.... You are absolutely correct. Already posted #1927 fix.

PR #1932, opened 2021-09-10
fix(module:select): fix data source of type IEnumerable<object>

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

#1906 introduces a new approach to monitor DataSource change, but not handle the situation where ValueName is null. As a result, the following code which uses a DataSource with members of different types no longer works:
```
<Select DataSource="_data" @bind-Value="_value" Style="width:100px;"></Select>

@_value

@code{
    List<object> _data = new List<object> { 1, "2", 3d, true };
    object _value;
}
```

Another issue is, when we add a object of custom class to the `IEnumerable<object>` DataSource, `THelper.ChangeType` will fail. Sample code:
```
<Select DataSource="_data" @bind-Value="_value" Style="width:100px;"></Select>

@_value

@code{
    List<object> _data = new List<object> { 1, "2", 3d, true, new MyClass { Label = "Apple" } };
    object _value;

    class MyClass
    {
        public string Label { get; set; }

        public override string ToString()
        {
            return Label;
        }
    }
}
```

The simple demo code above may not have a real use-case, but obviouly we should make it work well.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix the data source which has members of different types |
| 🇨🇳 Chinese |           |


components/select/Select.razor.cs
```diff
@@ -807,6 +737,89 @@ protected override void OnParametersSet()
...
         }
 
+        private void EvaluateDataSourceChange()
+        {
+            if (DataSource == null && _datasource == null)
+            {
+                return;
+            }
+
+            if (DataSource == null && _datasource != null)
```
anddrzejb, line 747: Moving from setter cause this `if` block to loose at least a part of its usefulness and also is the cause why test `Do_not_react_when_DataSource_not_changed()` is failing. In the setter it was useful when setting `DataSource` from null to a non-null value. When executed in `OnParametersSet()` it is opposite - executes when `DataSource` is set to null when it was not null previously.

anddrzejb (changes requested):

components/select/Select.razor.cs
anranruye: No, I dont think so. I think they should be same before and after the change. `Do_not_react_when_DataSource_not_changed()` passed because you write the `OnDataSourceChanged` attribute after `DataSource` attribute, If we move `OnDataSourceChanged` to before `DataSource`, the test will fail, like this:
image

I add `handlerCalled = false;` before `cut.SetParametersAndRender(parameters => parameters.Add(p => p.DataSource, _persons));` . Now the test will pass.

anddrzejb: Well...it seems you are right on both accounts. But you did clean up the `if` branch. 
There is one thing though which I do not really like in `Select` and I was wondering what is your opinion on this.
I have an issue with `OnDataSourceChange` and with `OnSelectedItemChanged`/`OnSelectedItemsChanged`, namely that they are fired during initialization of the `Select` component. I think they should not fire when the component is constructed, which not is the case now...

anddrzejb (approved):

components/select/Select.razor.cs
anranruye: Sometimes I have the same feeling. But triggering the `Change` events during initialization may be useful when the users want to inject custom logic for the first render. One example is the Table OnChange(now it is used to get remote data). Of course this smells not good, however we should provide an alternative way to inject these logics if we want to change this behavior.

ElderJames (approved):

components/select/Select.razor.cs
anddrzejb: Yeah, I get that, but that custom logic should be placed in the `OnInitialized`/`OnInitializedAsync` of the component using `Select`. For me, I do have logic in `OnSelectedItemsChanged` but I have to put extra `_afterFirstRender` boolean just to avoid running the logic in there during initialization. And this is in multiple places. I just have a feeling that more often you don't have to do any logic on initialization than you do.

anranruye: I agree.

PR #1938, opened 2021-09-12
change log 0.9.4

ElderJames:


anddrzejb:
Shouldn't we mentions breaking changes in `TriggerType` => `AntDesign.Trigger` enum & `PlacementType` => `AntDesign.Placement` enum?

ElderJames:
@anddrzejb Yes, we need it. Maybe we can add it to the changelog.

anddrzejb:
@ElderJames I think you are the only one who can edit the release log.

PR #1939, opened 2021-09-12
change: test cs project to include testProjectWithAdapter

anddrzejb:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [x] Other (about what?)

### 💡 Background and solution
Few days ago I discovered this [Fine Code Coverage](https://marketplace.visualstudio.com/items?itemName=FortuneNgwenya.FineCodeCoverage) extension to VS that allows to generate coverlet reports and actually track the coverage. For `AntBlazor` it looks like this:
image

It required 2 changes to `AntDesign.Test.csproj` file, which I would deem completely harmless, except when you are using the aforementioned extensions - then extremely useful. 
There is a problem here though - after running tests, it takes some time for the coverage report to generate. I consulted with the creator of the extension and we found out that this is not the problem of the extension but either coverlet or blazor. Maybe you remember that we used to have a serious delay when tests were running in CI - like 18 minutes. Only after switching the tests to run in `Release` configuration, the tests execution went down to 3 minutes. Same is here. If in `Debug` configuration, then it takes much longer than when in `Release`. It is still too long, but it seems this problem is related more to coverlet or blazor (or both not cooperating nicely). I know there are going to be very few contributors using this tool (even if we put it in the wiki), but it still is useful. Even if only to me :smirk:.

Also, added mock for disposal of overlay to base class. Because overlay is often hidden in multiple different components, this will prevent occasional bunit exceptions in tests. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/tests/AntDesign.TestKit/AntDesignTestBase.cs
+++ b/tests/AntDesign.TestKit/AntDesignTestBase.cs
@@ -17,6 +17,7 @@ namespace AntDesign.Tests
 
             //Needed for Tests using Overlay
             Services.AddScoped<AntDesign.JsInterop.DomEventService>(sp => new TestDomEventService(Context.JSInterop.JSRuntime));
+            JSInterop.SetupVoid(JSInteropConstants.OverlayComponentHelper.DeleteOverlayFromContainer, _ => true);
 
             CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.GetCultureInfo("en-US");
         }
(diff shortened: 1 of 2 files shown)


ElderJames:
Pretty cool. 

I also wanted to include E2E testing using [Playwright](https://playwright.dev/dotnet/) because there are some behaviors that are triggered by events that we can't cover at the moment, and E2E is the closest thing to user experience.

ElderJames (approved):

anddrzejb:
@ElderJames Good idea with Playwright! I wanted to play with it for some time, just did not get around to do it. I am working on some performance optimizations now, but when I finish, I will try to have a look at it.

PR #1999, opened 2021-10-04
feat(module:table): virtualizing

anranruye:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | table virtualizing |
| 🇨🇳 Chinese | 虚拟化 |

### DIFF
--- a/components/core/Virtualization/IVirtualizeJsCallbacks.cs
+++ b/components/core/Virtualization/IVirtualizeJsCallbacks.cs
@@ -1,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET5_0 || NET_6_0
+namespace AntDesign
+{
+    internal interface IVirtualizeJsCallbacks
+    {
+        void OnBeforeSpacerVisible(float spacerSize, float spacerSeparation, float containerSize);
+        void OnAfterSpacerVisible(float spacerSize, float spacerSeparation, float containerSize);
+    }
+}
+#endif
--- a/components/core/Virtualization/IVirtualizeJsCallbacks.cs
+++ b/components/core/Virtualization/IVirtualizeJsCallbacks.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#if NET5_0 || NET_6_0
+#if NET5_0 || NET6_0
 namespace AntDesign
 {
     internal interface IVirtualizeJsCallbacks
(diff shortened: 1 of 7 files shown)


components/core/Virtualization/IVirtualizeJsCallbacks.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if NET5_0 || NET6_0
```
anddrzejb, line 5: Wouldn't it be more future-proof if it was `NET5_0_OR_GRATER`?

anddrzejb:
Do I see it correctly? You adapted virtualization component to fit `AntBlazor` table? I am to tired to try to understand the details, but if this works, this is amazing! I was thinking to do the same thing for `Select` component for like 6 months now, but somehow I couldn't start working on it. Anyways, I hope this feature gets merged! 🚀 💯

anranruye:
@anddrzejb I changed two places. I add two parameters:
```C#
        [Parameter]
        public string SpacerElement { get; set; } = "div";

        [Parameter]
        public float? ContainerSize { get; set; }
```
The original `SpacerElement` is hard-coded as "div", the new `SpacerElement` parameter can change the generated element type.

```
Line 195:     builder.OpenElement(0, SpacerElement); // original code: builder.OpenElement(0, "div");
```

The `ContainerSize` parameter can specify a explicit size to override the js calculated size.

```
Line 264:     if (ContainerSize != null) containerSize = ContainerSize.Value; // this line is added
```

For `Table` component, the problem is that `<table>` or `<tbody>` element doesn't respect `height` css attribute(unless we change the `display` css attribute). When we apply `height` attribute, it acts as a minimum height but not actual height. So js can not get the expected height. This is why I use an additional parameter to specify the container height. 

For `Select` component, I think there are different difficulties.

anddrzejb:
Thanks for this explanation! I guess a deep-dive into virtualization awaits me once I start working on the virtualization for the `Select` component. But I have some other things I have to/want to finish before I can start any work in that area. Unless you'd like to do it?

components/core/Virtualization/IVirtualizeJsCallbacks.cs
anddrzejb: I played around with what you wrote and at first I could not recreate your findings. What I found out is that you have to have a razor file for this to actually behave as expected.
So, if following this example:
1. Having only `ComponentForNet5.cs` will create an element, as in the screenshot.
2. Having only `ComponentForNet5.razor.cs` will also create an element.
3. Having `ComponentForNet5.razor` and `ComponentForNet5.razor.cs` does not produce the element.
I do not know if you want to make appropriate changes, but at least this is some kind of an explanation.

AndersSahlin:
Hello, is this feature going to end up in Ant Blazor?

anranruye:
@AndersSahlin of course yes.

Issue #1051, opened 2021-01-26
Provide option to build only certain AntDesign components


MutatePat:
### Is your feature request related to a problem?
As Blazor projects get bigger, there becomes a need of using multiple component libraries to effectively build an application quickly. However with each full NuGet package, there comes a wasm start up cost, as well as memory considerations.

What if we could select which components we want to include in the package, so that we only include code for what we use? If I build an application which only uses AntDesign table, I do not need modal, drawer, and other heavy components to be sent to the client. 

### Describe the solution you'd like
Two versions of AntDesign. Full and Lite. Full includes all components we have created, while lite only includes the core folder.

When adding Lite to a project, we should be able to include further components such as Table, Button, Drawer, Affix etc

### Additional context
N/A

ElderJames:
Thanks for contacting us @MutatePat 
You are right, We maybe can publish single component package with a shell script in the github action pipeline.

Or publish all dlls in one package.

Do you have any idea with it?

PR #2016, opened 2021-10-11
new component MultiRangeSlider

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [x] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
This is a completely new component. I actually started investigating `AntBlazor` because I wanted to build that component. I work a lot with periods (both date periods and time periods) and I needed to have component that would suit my needs. Eventually I came up with this PR. I will not go into details of what it is and how it works - please check the preview - component name is `MultiRangeSlider` in `Data Entry` section of the components menu.

In the final stages of development I hit a major performance drop down. Eventually I found out this is due to the events we register for the `Overlay`. The more `Overlay` type of components we have, the bigger performance penalty `AntBlazor` puts on the site. What I mainly refer to is the fact we register `OnMouseUp` for every `Overlay` component (this goes to `document` dom element). That means any click anywhere in the document calls all `OnMouseUp` for every `Overlay` component. I figured we need the `OnMouseUp` only once the overlay is shown and we do not need it any longer, once overlay is gone. The same logic I apply for `mousemove` in the component.

The starting point for `MultiRangeSlider` component was `Slider` component. There are a lot of familiar code sections, but I also improved the overall performance & reduced code duplication. I hope to employ the same optimizations to `Slider` component. I also included a functionality of range dragging, which I will eventually also apply to `Slider`.

I invite all @ant-design-blazor/committer & anyone who is interested to comment on any bugs you are going to find. In the meantime I will continue adding tests. However, @ElderJames , if you find the component ready for merging, please do. 
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  new `MultiRangeSlider` component         |
| 🇨🇳 Chinese |           |


ElderJames:
I have to say it's great work. But I think this component should be used as part of the Slider, or as an extension. 
And the style should also follow ant design, the current one is a bit wide 😂.

ElderJames:
Hey @anddrzejb ,

I've been thinking recently about how to include community-contributed components.

I think our goal is to implement `AntDesign` completely, so it's not appropriate to put community-contributed components in the main component library, as this would increase our maintenance costs.

My suggestion is to put it in a separate library, such as `AntDesign.Contrib`, specifically for community-contributed components.

Feel free to discuss.

ElderJames:
Of course, we can still have these components displayed in our documents and in a separate secondary menu (just like the Charts).

anddrzejb:
I invite again all @ant-design-blazor/committer as well as @theilgaz (as he expressed interest in using this component) to comment here.

So a disclaimer first - obviously I would prefer the component making into the master branch for a number of reasons:
- better exposure,
- in the maintenance loop, so better chance for other contributors to help fixing the bugs & improve the performance
- it is ready, so no more time needed to spend on it.  

Of course I put a lot of effort & time into making this component, so it is a bit uncomfortable to accept it may not get merged. However, this is not "a hill to die on" for me.

So I am ok with making a separate library. Moreover, I suggest to actually start addressing #1051. Anyway, for making a separate nuget for contributors' components, we anyway need to extract the base. My suggestion is to create (on top of already existing `AntDesign`) a component `AntDesign.Core` and probably another nuget package for each other component. I imagine this would mean splitting `AntDesign` project into `AntDesign.Core` and `AntDesign.Components` project? I am not sure here as I have no experience on publishing nuget packages. Can such structure support the `AntDesign` general package & `AntDesign.Core` & other components?

Regarding the documentation - to be honest I don't like the idea. The only reason I don't is due to exposure. I think most people go through the `Components` link and rarely scan other options. If it is possible, I would rather we have a section in the main component list (like we have now "Navigation", "Data Entry", "Data Display"). Again, not a hill to die on, but definitely a preference.

theilgaz:
`MultiRangeSlider` can be used in many different ways to build rich UX for Blazor apps. Multiple range is useful for getting multiple information from user input at once. AntBlazor should have much more components like this.

@anddrzejb maybe we can create `AntDesign.Contrib` project for extending existing components and new ones. As we talked before, there would be nice if we have components like `SelectGrid`, `Scheduler`, `OrganizationTree` etc.

ElderJames:
In my opinion, the independent community library helps to encourage the community to use the infrastructure we provide to create the components they want, without the constraints of the original Ant Design specification.

We can start by separating `AntDesign.Core`.

theilgaz:
`AntDesign.Core` would be great step to start separating AntBlazor infrastructure. 

Reusable components are really important to build solutions/applications in a short time. Our community is growing and all of these line-of-business application requirements should be included to AntBlazor. 

Maybe we can create a sample for AntBlazor.Crud as a out of box template.

Zonciu:
Each component should be able to reference independently and use, I think we can learn from [react-component](https://github.com/react-component), separate repositories for each component to implement component functionality and provide unstyled/simple styled demos. `ant-design-blazor` repository references this components and apply `and design` style.

ElderJames:
@theilgaz You are right, we need an out-of-box CRUD Template for a long time.

anddrzejb:
Yeah, I think if we should have a nuget package for each component. Here is the question though - how will the solution structure look like? My take is that we need to create a new project `AntDesign.Core`. What about the rest? Do we need a separate project per component? Or can the nuget packages for each component be created based on selected files/folders? 

Also, what will go to the `AntDesign.Core`? I mean - should we put the services in there as well? Or they are treated as separate components? I am referring to messaging or dialog services.

anranruye:
I think we should have a separate project for each single component or service. But I'm not sure whether we should have many repositories or only one repository. And we should consider how to release and maintain the version number of the components, do they share the same version or just evolute independently?

anddrzejb:
I guess I will start with extracting the `AntDesign.Core`. Regardless how we proceed further, this I think we agree is the first step. Anyway, I think this has to be done in several PRs, otherwise it will be a nightmare to keep it in sync with master.

Hona:
One thing to mention is that separate projects for each component will make the huge performance issues with this solution become non existent
________________________________

theilgaz:
We should talk about `AntDesign.Core` and separating or not issues in a discussion.

PR #1996, opened 2021-10-03
fix(module:select): fix null value

anranruye:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Fixes #1995
#1783

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix null value option |
| 🇨🇳 Chinese |           |

### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #2023, opened 2021-10-13
fix(module:select): Null option shows in SelectContent

### LINKED ISSUE #1587: SelectOption bug when Value = 0

anddrzejb:
### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fixes #1587 
Addresses issues mentioned in [this comment](https://github.com/ant-design-blazor/ant-design-blazor/pull/1996#pullrequestreview-771855643).
Addresses proposal in [this comment](https://github.com/ant-design-blazor/ant-design-blazor/pull/1906#issuecomment-913105602).

### 💡 Background and solution
To fix the primary issue, in this PR I added a boolean that identifies if there is a `SelectOptionItem` that has its `Value` equal to the default value of `TItemValue`. I use that flag to figure out what to do when the value is selected and when the `Select` is cleared. On top of that, it also adds the parameter `ValueOnClear` which will be used as a primary value when `Select` is cleared. `ValueOnClear` may be set to a value that does not exist in `SelectOptionItem`. In that case, the `SelectContent` is cleared. Otherwise it is set to the existing `SelectOptionItem`. I included test cases to cover for these scenarios (I think I cover all possibilities, at least until proven otherwise :smirk:). 

During my tests I found out we have some weird issues with `Select`. It seems it relies a lot on things happening during re-rendering. I am not a fan of this, as it splits responsibilities and is really hard to follow (and thus debug). In my opinion this makes `Select` more complex & less performant. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  new `[Parameter] ValueOnChange` - stored `Value` that will be used when clear button is pressed      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/LabelTemplateItem.razor.cs
+++ b/components/select/LabelTemplateItem.razor.cs
@@ -1,4 +1,7 @@
-﻿using Microsoft.AspNetCore.Components;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using Microsoft.AspNetCore.Components;
 using AntDesign.Select.Internal;
 
 #pragma warning disable 1591 // Disable missing XML comment
--- a/components/select/Properties.cs
+++ b/components/select/Properties.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
 using Microsoft.AspNetCore.Components.Web;
 
 namespace AntDesign
--- a/components/select/ResponsiveTag.cs
+++ b/components/select/ResponsiveTag.cs
@@ -1,4 +1,8 @@
-﻿#pragma warning disable 1591 // Disable missing XML comment
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#pragma warning disable 1591 // Disable missing XML comment
 
 namespace AntDesign.Select
 {
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
 using System.Collections.Generic;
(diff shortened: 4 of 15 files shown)


anranruye:
Well, I didn't have a look at the codes yet. From your description, I feel this solution makes things more complicated. As we discussed before, the basic logic for `Select` should be simpler. With no fallback strategy, the `Select` component accepts all values of `TItemValue`, if the value is in select-options, then we display that item, otherwise we display nothing or a placeholder identities nothing(null) or a placeholder identities a not-null value which is not in the options. When clear, set the value to default(TItemValue) or ValueOnClear. 

With a fallback strategy, when the value is:
1. value in select-options, display that item, no fallback
2. null, display a placeholder, no fallback
3. fallback value defined by the strategy(null, default(TItemValue) or a user specified value), display nothing or a placeholder
4. other values, fallback

That's all. And for most cases, we don't need a fallback strategy, which makes things as simple as I mentioned in the first paragraph.

jeffraska:
Hi, I think this PR breaks Two-way binding in Default mode while changing binded value to NULL.

- When I dont have any special NULL value option in DataSource:
1. `Select.OnValueChange()` with NULL value is called.
2. TItemValue is `long?` so condition is true and `OnInputClearClickAsync()` is called.
3. `ClearDefaultMode()` is called.
4. Value Equals to default(long?) so this method just returns and do nothing, SelectedOptionItems is not changed and user still see previous value.


- When I have special NULL value option in DataSource:
1. `Select.OnValueChange()` with NULL value is called.
2. `EvaluateValueChangedOutsideComponent()` is called.
3. ActiveOption is not NULL and Value Equals to default(long?) so this method just returns and do nothing, SelectedOptionItems is not changed and user still see previous value.

Or am I doing something wrong and missing something?

anranruye:
@jeffraska Yes, I can reproduce this.

Issue #2067, opened 2021-10-28
Setting Calendar.DateFullCellRender throws an exception


szymski:
### Describe the bug

When one is trying to customize date cell rendering in the Calendar component, an exception is thrown.

### Steps to reproduce
```c#
@{
    Func<DateTime, RenderFragment> RenderCell = time => @<span>Should work</span>;
}
<Calendar DateFullCellRender="RenderCell"/> // Will throw exception on render
```

### Proposed solution

In [calendar component](https://github.com/ant-design-blazor/ant-design-blazor/blob/e698e40293a2046edfcc35e475fc357b5b003d82/components/calendar/Calendar.razor.cs#L55) the DateFullCellRender takes 1 DateTime parameter.
It is later passed to DatePickerDatePanel [here](https://github.com/ant-design-blazor/ant-design-blazor/blob/385dcb4687e0b945b47da1b60f59f6fe32c7b833/components/calendar/internal/CalendarPanelChooser.razor#L26), which is a mistake, because [DatePickerPanelBase.DateRender](https://github.com/ant-design-blazor/ant-design-blazor/blob/e698e40293a2046edfcc35e475fc357b5b003d82/components/date-picker/internal/DatePickerPanelBase.cs#L66) has a different signature. 

Signature of one of these functions has to be changed or conversion to a different delegate type must occur in some place.

### Exceptions (if any)
``
InvalidCastException: Unable to cast object of type 'System.Func`2[System.DateTime,Microsoft.AspNetCore.Components.RenderFragment]' to type 'System.Func`3[System.DateTime,System.DateTime,Microsoft.AspNetCore.Components.RenderFragment]'.
``

### Further technical details
- AntDesign Nuget Package version: 0.10.1

PR #2068, opened 2021-10-28
fix(module: calendar): DateFullCellRender throws an exception (#2067)

szymski:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design-blazor/ant-design-blazor/issues/2067

### 💡 Background and solution

Setting `Calendar.DateFullCellRender` property would result in a cast exception thrown. That was because that function (1 parameter) was passed to `DatePickerBase` which required 2 parameters. I propose a solution where we create a new anonymous function which calls the original `Calendar.DateFullCellRender` with the second parameter discarded.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     Fixed `Calendar.DateFullCellRender` throwing an exception   |
| 🇨🇳 Chinese |           |


ElderJames (approved):
LGTM

PR #2114, opened 2021-11-16
Add Parameter for Calling change on Input for InputNumber

rabberbock:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

I have a need to update as the user types. I added a Parameter to InputNumber that can be used to set whether OnCgange will be called on an input event.

Any thoughts on this behavior and approach?

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


rabberbock:
@ElderJames Just wanted to check in before actually implementing that you would open to this change. Thanks!

ElderJames:
@rabberbock Thanks for contribute this. 
Actually, we have added this mechanism to Input, currently through delayed binding, to solve the problems caused by fast input.

What we need at the moment is to add a switch to disable modification-time binding, as the time delay causes the entry characters to be rolled back when using Server mode.

rabberbock:
@ElderJames I see. Would this be a better implementation for now as it does not update the value right away, but allows the caller to supply an OnInput?

rabberbock:
@ElderJames Would you be open to this change in a different PR?

PR #2129, opened 2021-11-23
feat/ApplyPredefinedQueryModel: Reloading table by setting predefined QueryModel

### LINKED ISSUE #1204: Should use SortModel and FilterModel copies to build table QueryModel
Currently the table OnChange event is triggered twice when click UI pagination buttons(#949). I find this issue when I try to avoid the second one by comparing QueryModels. If the QueryModel has same options with the previous one, then I don't load data from server. This works well. However, when I change column sort direction, the table doesn't change. Then I find that the SortModels of all QueryModels are always same. 
https://github.com/ant-design-blazor/ant-design-blazor/blob/15f4cd134f4d97e6c244708c240233f82eb670c8/components/table/Table.razor.cs#L196-L210
Consider another scenario, if we want to remember the QueryModel options and resue them later, we can't do that because the SortModel and FilterModel are not fixed.

现在在点击UI页面上的分页按钮时OnChange事件会被触发两次, 在这个问题被修复之前我想通过比较两次事件触发的QueryModel参数来避免对服务器发出重复的请求, 由此引出了本帖子所述的问题.

比较方法是每次事件触发时将QueryModel参数与`oldQueryModel`字段(初始值为null)通过自定义扩展方法比较其内容, 如果不同则将QueryModel引用赋值给`oldQueryModel`, 相同则直接return.

这个方法确实避免了对服务器发出重复请求, 但是排序功能不好用了, 原因是排序改变时新的QueryModel和`oldQueryModel`判断相等的结果是true, 也就是说`oldQueryModel`的SortModel没被固定下来, 排序一变化就跟着变了, 源代码表明确实如此. (筛选功能同理)

另一个受影响的场景是有时候需要记住查询参数然后在之后重用.

我从直觉上认为, 作为事件参数的QueryModel其SortModel和FilterModel应该是固定的, 也就是说应该使用列Model当前值的副本来创建QueryModel. 实现 #1201 的时候也应该考虑该问题.

### LINKED ISSUE #2015: Table: applying a predefined QueryModel to table
### Ability to apply a saved table configuration - setting QueryModel (PageIndex, PageSize, SortModel, FilterModel)
I would like to offer users to save their table configuration (QueryModel), and later to apply saved filter and configurations. With that feature users would not have to set filters every time. 

For example, I would do json serializing of Query Model and save that string in database and later I would take and deserialize that string. In OnChange method inside table I would set current QueryModel to deserialized one from database. Unfortunatlly deserializing and making new QueryModel is not possible because it's properties are getter-only, so I didn't find way to apply saved configuration. I am open to any suggestions or ideas how it can be implemented!

AnaNikolasevic:
This feature will give opportunity to serialize/deserialize QueryModel and then apply that predefined QueryModel to Table with ITable.ReloadData(QueryModel queryModel) method. Beside that, there is possibility to reset all filters and sort with ResetData() method. Also one minor bug is fixed, earlier current selected value in 'PageSize / page' selection inside pagination was not updated in case using ReloadData(PageIndex, PageSize) method.

### 🤔 This is a ...

- [ x ] New feature
- [ x ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
fixed #2015 , fixed #1204

### 💡 Background and solution

The main goal of implementing this function was to enable users to save their table configuration (QueryModel) and later apply it. With this feature, users would not have to set Filters, Sort, Page Size every time.

Due to the need to save QueryModel locally, I added some classes to make it easier. The first step is to save the QueryModel and the most common way would be to serialize it and save it as a JSON string. The structure of QueryModel required some serialization settings, so I added a few classes that represent JsonSettings used for serialization. The same settings are also used for the next step - deserialization of the JSON string.
The ITable.ReloadData(QueryModel queryModel) method is used to apply a predefined QueryModel model to a table. With this method, all configurations from QueryModel will be applied to the table - Page Index, Page Size, Filters, Sorting.

The next improvement is the ResetData () method, which allows you to reset all filters and sort applied to the table.

The last minor fix is ​​to update the currently selected value in the 'PageSize / page' selection within the Pagination section. An error was noticed in the ReloadData(PageIndex, PageSize) method where, although a new page size value was set and the number of elements changed, the current value within the 'PageSize / page' selection was not updated.

I am also open to hear any advices or better solutions for this feature.
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2129) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2129) before we can accept your contribution.<br/><br/>:x: AnaNikolasevic [sign now](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2129)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/ant-design-blazor/ant-design-blazor?pullRequest=2129) it.</sub>

ElderJames:
Thanks for contribute this @AnaNikolasevic . But why did you create a `QueryModelJsonSettings` ? I don't see a reference to it anywhere.

ElderJames:
And we should use System.Text.Json .

AnaNikolasevic:
> Thanks for contribute this @AnaNikolasevic . But why did you create a `QueryModelJsonSettings` ? I don't see a reference to it anywhere.

Hello :) I used methods from QueryModelJsonSettings within own code for serializing and deserializing of QueryModel, I wanted to have all necessary setting in one place. Also, I tought it would be useful for other users to access it, so in case they want to serialize/deserialize QueryModel they can just use those settings. I will commit new changes without these settings and use System.Text.Json. If you have any other advices or suggestions, I am open to correcting and implementing them.

ElderJames:
@AnaNikolasevic Sounds good. I think we need to add a new demo to explain this feature.

ElderJames:
You can add a demo like ant.design by using the md file https://github.com/ant-design/ant-design/blob/master/components/table/demo/reset-filter.md

live demo: https://ant.design/components/table-cn/#components-table-demo-reset-filter

AnaNikolasevic:
Hello, I added demo for this feature. Please do review and let me know if you have any other suggestions.

site/AntDesign.Docs/Demos/Components/Table/demo/SaveLoadConfiguration.razor
```diff
@@ -0,0 +1,216 @@
...
+
+    void SaveTable()
+    {
+        SavedQueryModel = Table.GetQueryModel();
+    }
+
+    void LoadTable()
+    {
+        Table.ReloadData(SavedQueryModel);
+    }
```
ElderJames, line 174: I found that Filter does not restore state if you just set an object that has not had a serial number.

components/table/Column.razor.cs
```diff
@@ -443,5 +444,87 @@ private void InitFilters()
...
         }
+
+        void IFieldColumn.ClearFilters() => ResetFilters();
+
+        void IFieldColumn.SetFilterModel(ITableFilterModel filterModel)
+        {
+            foreach (var filter in filterModel.Filters)
+            {
+                if (filter.Value is JsonElement)
+                    filter.Value = GetObjectFromJsonElement((JsonElement)filter.Value);
```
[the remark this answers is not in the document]
AnaNikolasevic: In regular case where QueryModel is **not** deserialized from Json, filter.Value is object and **already has** assigned value, so there is no need for any change of that value

site/AntDesign.Docs/Demos/Components/Table/demo/SaveLoadConfiguration.razor
AnaNikolasevic: Thank you for noticing the bug, I fixed it. Needed `[JsonConstructor]` annotation for JSON deserialization  is only available in net5, but is not available in netcoreapp3.1, so maybe for JSON deserialization should order of TargetFrameworks be `net5;netcoreapp3.1` in AntDesign.Docs.Server project; or to find some other solution. I am open to any advices or suggestions.

oemil: @AnaNikolasevic: Shouldn't the JSON serialization of internal objects be handled outside of antdesign? So, for instance, remove the attribute and add a specific JsonConverter to your own project.

I mean, of course, it is convenient to directly serialize these objects, but this should not impact the dependency framework.

ElderJames (approved):

PR #2197, opened 2021-12-23
Improve sync logic about `Disabled` between `RadioGroup` and `Radio`s

LeaFrock:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Now if a radio is in a group, its `Disabled` will be always the same as `RadioGroup`. However, sometimes people may want to keep some radio disabled, so we should ignore the sync for a disabled radio when it becomes a child of group.

Also, BTW, this PR has fixed some other problems.

* Remove `IsChecked` which is duplicated of `Checked` in `Radio`
* Remove check of `Disabled` in `Select` and `UnSelect`, which should be checked before the method calls
* Use `EqualityComparer` in `EqualsValue`
* Fix extra `ValueChanged` invoking in `OnRadioChange`
* Add some test cases about `Disabled` sync of `RadioGroup`

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Improve sync logic about `Disabled` between `RadioGroup` and `Radio`s    |
| 🇨🇳 Chinese |     完善RadioGroup和Radio之间Disabled状态的同步逻辑   |

### DIFF
--- a/components/radio/Radio.razor
+++ b/components/radio/Radio.razor
@@ -8,7 +8,7 @@
                autofocus="@AutoFocus"
                class="@_inputClassMapper.Class"
                disabled="@Disabled"
-               checked="@IsChecked"
+               checked="@Checked"
                name="@_name"
                value="@Value"
                @ref="_inputRef" />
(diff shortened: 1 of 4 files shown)


ElderJames (approved):
LGTM

PR #2276, opened 2022-02-07
Feat/button aria

bweissronin:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Step closer to helping with ADA compliance.  Since this is my first pull request, I wanted to keep it small and targeted so feedback/guidance was easier.  I did notice that it's asking me to target the feature branch where the documentation was based off master.  If I target feature it includes far more changes.  If needed, I can abandon and recreate this off the feature branch. 

### 💡 Background and solution

1.  Screen readers leverage the aria-label to help direct impaired users.  
2. Nothing visual was changed.  
3. The simplest approach is to add a Parameter on the Button component to allow developers to bind a label value that a Screen Reader could use to help impaired users.  This seemed like the best approach given developers may use a single Button Component in multiple ways. The alternative was the use of attribute splatting which seemed like a clunky implementation for developers to manage.

Example
```<Button Label="Save Form">Save</Button>```

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     A Parameter, Label, has been added to the Button Component to support the aria-label attribute.      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/button/Button.razor
+++ b/components/button/Button.razor
@@ -6,7 +6,7 @@
             @onclick="HandleOnClick" disabled="@Disabled"
             @onclick:stopPropagation="@OnClickStopPropagation"
             @onmouseup="OnMouseUp"
-            ant-click-animating-without-extra-node="@(this._animating ? "true":"false")">
+            ant-click-animating-without-extra-node="@(this._animating ? "true":"false")" aria-label="@Label">
         @if (Loading)
         {
             <span class="ant-btn-loading-icon">
--- a/components/button/Button.razor.cs
+++ b/components/button/Button.razor.cs
@@ -78,6 +78,12 @@ namespace AntDesign
         public string Icon { get; set; }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


ElderJames:
Thanks for the contribution @bweissronin , it makes sense. 

Please sign the CLA first, and I have two suggestions:
- It would be better to use `AriaLabel` for the parameter names, so that more aria-* attributes can be added later, also following the JS API [ariaLabel](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabel).
- Switch the target brunch to `feature`.

bweissronin:
Closing.  Feature recreated off feature branch and PR'd in #2278

PR #2330, opened 2022-03-10
fix(module:Radio) Default name value

bweissronin:
Default the name value so the tab order is preserved.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Tabbing through a form that has multiple RadioGroups loses the tab order unless RadioGroups have a name value set.  This is problematic for users who are disabled and rely purely on the keyboard to complete their form.

### 💡 Background and solution

1.  Multiple RadioGroups on the same form throw the tab order off.
2. N/A
3. Set a default name value on the RadioGroup if one isn't provided.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |      x     |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/Base/AntInputComponentBase.cs
+++ b/components/core/Base/AntInputComponentBase.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq.Expressions;
 using System.Threading.Tasks;
+using AntDesign.Core.Reflection;
 using AntDesign.Forms;
 using AntDesign.Internal;
 using Microsoft.AspNetCore.Components;
@@ -21,6 +22,9 @@ namespace AntDesign
         private bool _previousParsingAttemptFailed;
         private ValidationMessageStore _parsingValidationMessages;
         private Type _nullableUnderlyingType;
+        private PropertyReflector? _propertyReflector;
+
+        protected string PropertyName => _propertyReflector?.PropertyName;
 
         [CascadingParameter(Name = "FormItem")]
         private IFormItem FormItem { get; set; }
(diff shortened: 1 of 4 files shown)


ElderJames:
Thanks for contribute this @bweissronin . I think it's more readable using index.

bweissronin:
> Thanks for contribute this @bweissronin . I think it's more readable using index.

I'm happy to make things more readable.  I'm slightly confused about which index you're referring to.

bweissronin:
> > Thanks for contribute this @bweissronin . I think it's more readable using index.
> 
> I'm happy to make things more readable. You referring to place a tabindex on the input? If so, I believe I tried this before and was getting inconsistent results. Happy to double-check.

@ElderJames I tested using the tabindex only against the examples on the Components documentation page and it doesn't work as intended.  

The below steps were used.  This assumes the demo site is running locally:

1. Set tabindex to 0 on on the `input` in the Radio.razor
2. Go to the Radio demo page.
3. Scroll down to the section for `Generic` examples.  
4. Click the radio for value `1` so it has focus.
5. Press tab.  

The expected result is that it'd tab to the group below starting with value `A` which it doesn't.  If you run through the same scenario without the tabindex attribute and set a default name for the Radio at the RadioGroup level, it'll function as expected.

ElderJames:
Oh, that's my fault. I submitted a commit 07b872493acf52ac192c44814254c96362030438 to use the binding property as the default name. What do you think?

bweissronin:
> Oh, that's my fault. I submitted a commit [07b8724](https://github.com/ant-design-blazor/ant-design-blazor/commit/07b872493acf52ac192c44814254c96362030438) to use the binding property as the default name. What do you think?

Works for me! Thanks for checking it out!

ElderJames (approved):

PR #2400, opened 2022-04-16
Fix FieldIdentifier equality check in Rules Mode OnFieldChanged

GHMonad:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

An InvalidOperationException is thrown when more than one input of a given form is bound to a property with the same name and we want rules based validation, even if they belong to a different instance.
This can happen in a couple of ways, for example we can have a view model that has two properties: *Player1* and *Player2*, both of type *Player*. Currently, we cannot add an input field for *model.Player1.Name* and *model.Player2.Name* both because we would have two fields with *FieldName* *Name*.
Similarly, if our view model had a *Contacts* property of *List<ContactInformation>* type, we could not bind the properties of *ContactInformation* to input fields, because they would clash.

The problem seems to be caused by the way equality is checked for fields in the line below:

https://github.com/ant-design-blazor/ant-design-blazor/blob/bb1030a5a4026c6fac379253db34a1fad5fea33a/components/form/Form.razor.cs#L240

Since FieldIdentifier provides its own implementation to check equality and that takes the underlying model instance into account, we can fix this by simply using that.

### 📝 Changelog

Rules and Complex form validation modes can now work properly with complex composite view models where the same field name is bound to different input elements. Not a breaking change, since the current behavior is more restrictive.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    x      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/Form.razor.cs
+++ b/components/form/Form.razor.cs
@@ -237,7 +237,7 @@ namespace AntDesign
             _rulesValidator.ClearError(args.FieldIdentifier);
 
             var formItem = _formItems
-                .Single(t => t.GetFieldIdentifier().FieldName == args.FieldIdentifier.FieldName);
+                .Single(t => t.GetFieldIdentifier().Equals(args.FieldIdentifier));
 
             var result = formItem.ValidateField();
 
--- a/tests/AntDesign.Tests/Form/EditContext/Form.EditContext.ValidationTests.razor
+++ b/tests/AntDesign.Tests/Form/EditContext/Form.EditContext.ValidationTests.razor
@@ -1,0 +1,56 @@
+﻿@inherits AntDesignTestBase
+@code {
+
+    class Model
+    {
+        public string Name { get; set; } = "";
+    }
+
+    class CompositeModel
+    {
+        public Model First { get; set; } = new Model { };
+        public Model Second { get; set; } = new Model { };
+    }
+
+    FormValidationRule _validateStringValueMaxThreeCharacters = new FormValidationRule {
+        Type = FormFieldType.String, 
+        Max = 3, 
+        Message = "This field cannot be longer than 3 characters." 
+    };
+
+
+#if NET6_0
+	[Fact]
+	public void Form_EditContext_with_RulesModeValidation_and_CompositeModel_should_validate_properly()
+	{
+	    //Arrange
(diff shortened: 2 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2400) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2400) before we can accept your contribution.<br/><br/>:x: GHMonad [sign now](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2400)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/ant-design-blazor/ant-design-blazor?pullRequest=2400) it.</sub>

ElderJames:
Thanks for contribute this @GHMonad . Could you please provide a test case?

GHMonad:
> Thanks for contribute this @GHMonad . Could you please provide a test case?

Hi @ElderJames, sure thing. Can you point me to the existing test cases for that component, I could only find the tests for the form validation logic and not the Form itself.

GHMonad:
> > Thanks for contribute this @GHMonad . Could you please provide a test case?
> 
> Hi @ElderJames, sure thing. Can you point me to the existing test cases for that component, I could only find the tests for the form validation logic and not the Form itself.

Never mind, found some with a NET6_0 precompiler condition so they just didn't show up. I guess I should follow that lead.

ElderJames:
Sorry, please fix the test

GHMonad:
> Sorry, please fix the test

My test passed. The failed test (Button_animation_when_cliked_end_after_some_time) has nothing to do with the changes, it's just flaky since it has a Delay in it...

ElderJames (approved):
LGTM

PR #2404, opened 2022-04-18
fix(module:Tooltip) Tabindex

bweissronin:
Applies a tabindex to the wrapping div to help with focus driven from tabbing.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
The Tooltip does not allow keyboard-based users to access their content because it's not a tab-able element.  Adding a `tabindex` to the `div` that's handling the events fixes this.

1. Users cannot interact with the Tooltip placed on an element via Keyboard.
2. N/A
3. Apply a tabindex to the wrapping `div` that handles the event listeners. 
### 📝 Changelog

Users will be able to tab to elements that leverage the Tooltip component.
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tooltip/Tooltip.razor
+++ b/components/tooltip/Tooltip.razor
@@ -12,7 +12,8 @@
      @oncontextmenu="OnTriggerContextmenu"
      @onfocusin="OnTriggerFocusIn"
      @onfocusout="OnTriggerFocusOut"
-     @oncontextmenu:preventDefault>
+     @oncontextmenu:preventDefault
+     tabindex="0">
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
Fantastic!

PR #2512, opened 2022-06-10
feat(module:datepicker): Scroll to selected time in DatePicker/TimePicker

Alexbits:
### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/internal/DatePickerBase.cs
+++ b/components/date-picker/internal/DatePickerBase.cs
@@ -274,6 +274,27 @@ namespace AntDesign
 
         protected ClassMapper _panelClassMapper = new ClassMapper();
 
+        internal event EventHandler<bool> OverlayVisibleChanged;
+        private readonly object _eventLock = new();
+        event EventHandler<bool> IDatePicker.OverlayVisibleChanged
+        {
+            add
+            {
+                lock (_eventLock)
+                {
+                    OverlayVisibleChanged += value;
+                }
+            }
+
+            remove
+            {
+                lock (_eventLock)
+                {
+                    OverlayVisibleChanged -= value;
+                }
+            }
+        }
+
         protected override void OnInitialized()
         {
             // set default picker type
@@ -783,7 +804,7 @@ namespace AntDesign
             return null;
         }
 
-        public void InvokeStateHasChanged()
+        public new void InvokeStateHasChanged()
         {
             StateHasChanged();
         }
@@ -815,5 +836,7 @@ namespace AntDesign
             }
             return orderedValue;
         }
+
+        protected void InvokeInternalOverlayVisibleChanged(bool visible) => OverlayVisibleChanged?.Invoke(this, visible);
     }
 }
--- a/components/date-picker/internal/DatePickerBase.cs
+++ b/components/date-picker/internal/DatePickerBase.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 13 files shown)


Alexbits:
Hey @ElderJames. I've implemented auto-scroll behavior similar to the one in ant.design. It may need a smoother animation when the selection panel (overlay) is open, but it generally works okay. Please check and let me know what you think.

ElderJames:
Well done! You can refer to BackTop component that it scroll more smoothly.

And I find that it wouldn't scroll when I modify the time value directly:

date-picker-time-scroll-top

Alexbits:
> Well done! You can refer to BackTop component that it scroll more smoothly.
> 
> And I find that it wouldn't scroll when I modify the time value directly:
> 
There are other issues with manual input.

- DatePicker.Razor.cs.OnBlur resets the value to the cached if Enter key is not pressed, the Ok button does nothing.
- The time string with AM/PM is not parsed by FormatAnalyzer.TryPickerStringConvert and the value is not changed as result.

Alexbits:
> * DatePicker.Razor.cs.OnBlur resets the value to the cached if Enter key is not pressed, the Ok button does nothing.
I've tried changing it to accept parsed value immediately but in this case, a unit test fails. So what behavior do we want here? Should it apply the value of manual input with Enter key and Ok button only? But in this case, why is the input from the panel confirmed immediately?  🤔 In the ant.design the inout is applied only after Enter key or Ok button.

components/date-picker/internal/DatePickerBase.cs
```diff
@@ -783,7 +804,7 @@ public virtual void ClearValue(int index = 0, bool closeDropdown = true)
             return null;
         }
 
-        public void InvokeStateHasChanged()
+        public new void InvokeStateHasChanged()
```
ElderJames, line 807: We can change this method to `virtual`

Alexbits: Hey @ElderJames.  This will require changing the implementation in the `AntComponentBase` to `public virtual` from `protected` for `IDatePicker` interface to work. And, in this case, there is probably no need for implementation in the [DatePickerBase.cs](https://github.com/ant-design-blazor/ant-design-blazor/pull/2512/files/d3a808e4c8e106f751d86623e4749b4ee380c51e#diff-dbbc0da12dd4220386a71c6ed7f2afa809932e99e6f44740f3b3e10cd73493f6)

Alexbits: > We can change this in AntComponentBase to `protected virtual`, right?
Yes, we can, but it's `public `in `DatePickerBase`, we can't override it.

Alexbits: It's called by `DatePickerTemplate` via `IDatePicker` interface, `protected `will break this.

ElderJames: It's a little confusing.

Alexbits: > It's a little confusing.

Yes. This was not designed by me. I've added a `new` modifier because of a compiler warning about the method which hides an inherited one.

Alexbits: I've fixed this by removing the `InvokeStateHasChanged` from `IDatePicker` altogether. There is no need for it since it's called a few lines lower in `DatePickerTemplate` when the placeholder value is set. So basically, it was invoked multiple times, causing unnecessary renderings. 

Also, when the Value is selected, there is no need to set the placeholder on mouse hover and call StateHasChanged again? I've fixed this too.

ElderJames: That's great!

Issue #1600, opened 2021-06-04
It seems two tone color icon is not working


ldsenow:
I was trying to set a two tone color for an icon. However, it doesnt seem to apply the twotone color.

`<Icon Type="check-circle" Theme="twotone" TwotoneColor="#3bc45a" />`

ElderJames:
Thanks for contacting us @ldsenow 

We haven't implemented ant design's color scale algorithm yet, so we can't use one color value to generate two colors.

ldsenow:
Anything i could help with this? Or it is already a plan for it?

ElderJames:
No plan yet. Could you please try to implement it with two color parameters? Use `Color` and `SecondColor`.

ldsenow:
Sure. I guess It is a temporary solution for now until u guys get the color scale working.

anddrzejb:
@ldsenow what is your idea of a temporary solution?

rqx110:
this feature is take too long time to implement. and if it not implement yet, why list on the web site? https://ant-design-blazor.gitee.io/zh-CN/components/icon#components-icon-demo-two-tone

👎

ElderJames:
@ldsenow @anddrzejb This issue has been solved !

PR #2513, opened 2022-06-12
feat(icon): two-tone color icon implements

rqx110:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
resolve https://github.com/ant-design-blazor/ant-design-blazor/issues/1600

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     two-tone color icon implements      |
| 🇨🇳 Chinese |     实现双色图标      |

### DIFF
--- a/components/core/JsInterop/modules/components/iconHelper.ts
+++ b/components/core/JsInterop/modules/components/iconHelper.ts
@@ -1,11 +1,43 @@
-﻿export class iconHelper {
+﻿import { generate as generateColor } from "@ant-design/colors";
+
+export class iconHelper {
   static createIconFromfontCN(scriptUrl) {
     if (document.querySelector(`[data-namespace="${scriptUrl}"]`)) {
       return;
     }
-    const script = document.createElement('script');
-    script.setAttribute('src', scriptUrl);
-    script.setAttribute('data-namespace', scriptUrl);
+    const script = document.createElement("script");
+    script.setAttribute("src", scriptUrl);
+    script.setAttribute("data-namespace", scriptUrl);
     document.body.appendChild(script);
+  }
+
+  static generateTwotoneSvgIcon(svgImg: string, primaryColor: string): string {
+    svgImg = svgImg
+      .replace(/['"]#333['"]/g, '"primaryColor"')
+      .replace(/['"]#E6E6E6['"]/g, '"secondaryColor"')
+      .replace(/['"]#D9D9D9['"]/g, '"secondaryColor"')
+      .replace(/['"]#D8D8D8['"]/g, '"secondaryColor"');
+    const secondaryColors = generateColor(primaryColor);
+    const svg: SVGElement = iconHelper._createSVGElementFromString(svgImg);
+    const children = svg.childNodes;
+    const length = children.length;
+    for (let i = 0; i < length; i++) {
+      const child: HTMLElement = children[i] as HTMLElement;
+      if (child && child.nodeName.toLowerCase() === "path") {
+        if (child.getAttribute("fill") === "secondaryColor") {
+          child.setAttribute("fill", secondaryColors[0]);
+        } else {
+          child.setAttribute("fill", primaryColor);
+        }
+      }
+    }
+    return svg.outerHTML;
+  }
+
+  private static _createSVGElementFromString(str: string): SVGElement {
+    const div = document.createElement("div");
+    div.innerHTML = str;
+    const svg: SVGElement = div.querySelector("svg");
+    return svg;
   }
 }
--- a/components/core/JsInterop/modules/components/iconHelper.ts
+++ b/components/core/JsInterop/modules/components/iconHelper.ts
@@ -1,6 +1,4 @@
-﻿import { generate as generateColor } from "@ant-design/colors";
-
-export class iconHelper {
+﻿export class iconHelper {
   static createIconFromfontCN(scriptUrl) {
     if (document.querySelector(`[data-namespace="${scriptUrl}"]`)) {
       return;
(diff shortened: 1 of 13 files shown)


rqx110:
截屏2022-06-12 10 59 14

ElderJames:
@rqx110 Excellent! Thank you so much!

components/core/JsInterop/modules/components/iconHelper.ts
```diff
@@ -1,11 +1,43 @@
...
-    script.setAttribute('data-namespace', scriptUrl);
+    const script = document.createElement("script");
+    script.setAttribute("src", scriptUrl);
+    script.setAttribute("data-namespace", scriptUrl);
     document.body.appendChild(script);
   }
-}
\ No newline at end of file
+
+  static generateTwotoneSvgIcon(svgImg: string, primaryColor: string): string {
```
ElderJames, line 14: I think we just need to get the secondary color in js and replace the value in C#.

components/icon/IconService.cs
```diff
@@ -28,7 +28,7 @@ public static string GetIconImg(string type, string theme)
             return iconImg;
         }
 
-        public static string GetStyledSvg(string svgImg, string svgClass = null, string width = "1em", string height = "1em", string fill = "currentColor", int rotate = 0)
+        public string GetStyledSvg(string svgImg, string svgClass = null, string width = "1em", string height = "1em", string fill = "currentColor", int rotate = 0)
```
ElderJames, line 31: No need to change.

components/core/JsInterop/modules/components/iconHelper.ts
rqx110: I didn't find solution to add attribute to svg string in C#.  maybe regex or xml?

ElderJames: Maybe [Regex.Replace](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.replace?view=net-6.0&WT.mc_id=DT-MVP-5003987).

rqx110:
I can't run the project successfully, after changed the base branch from `master` to `feature`.

image

also when I run it using dotnet6, not woking well.

components/core/JsInterop/modules/components/iconHelper.ts
rqx110: what about:
```cs
public async Task<string> GetTwotoneSvgIcon(string svgImg, string twotoneColor)
        {
            svgImg = svgImg.Replace("fill=\"#333\"", $"fill=\"primaryColor\"")
            .Replace("fill=\"#E6E6E6\"", $"fill=\"secondaryColor\"")
            .Replace("fill=\"#D9D9D9\"", $"fill=\"secondaryColor\"")
            .Replace("fill=\"#D8D8D8\"", $"fill=\"secondaryColor\"");
            
           // get secondaryColors from js
            var secondaryColors = await _js.InvokeAsync<string>(JSInteropConstants.GenerateColor, twotoneColor);

            var document = XDocument.Load(new StringReader(svgImg));
            var svg = document.Root;
            foreach (var path in svg.Nodes().OfType<XElement>())
            {
                if (path.Attribute("fill")?.Value == "secondaryColor")
                {
                    path.SetAttributeValue("fill", secondaryColors[0]);
                }
                else
                {
                    path.SetAttributeValue("fill", twotoneColor);
                }
            }

            return svg.ToString();
        }
```

components/icon/Icon.razor.cs
```diff
@@ -73,8 +73,6 @@ protected override async Task OnInitializedAsync()
                 Spin = true;
             }
 
-            await SetupSvgImg();
```
ElderJames, line 76: We need to render the icons in the prerendering stage.
Therefore, the twotone colors should not be replaced during pre-rendering and should be replaced after the first rendering.

rqx110: done!

components/icon/Icon.razor.cs
```diff
@@ -109,6 +113,10 @@ protected virtual async Task SetupSvgImg()
             {
                 var svg = IconService.GetIconImg(Type.ToLowerInvariant(), Theme.ToLowerInvariant());
                 _svgImg = IconService.GetStyledSvg(svg, svgClass, Width, Height, Fill, Rotate);
+                if (Theme == IconThemeType.Twotone)
+                {
+                    _svgImg = await IconService.GetTwotoneSvgIcon(_svgImg, TwotoneColor);
```
ElderJames, line 118: On the other hand, JS Interop calls need to be made after ensuring that they have been rendered. So it cannot be called during the prerendering phase.

rqx110: so, what do you think? move `if (Theme == IconThemeType.Twotone)` block to after render?

ElderJames: I think we can pass a flag `rendered` into SetupSvgImg method. Call it in OnInitializedAsync with `rendered=false`, and in `OnAfterRenderAsync`  with   `rendered=true` after first render.

```
SetupSvgImg(bool rendered)
```

rqx110: add a flag `_hasRendeered`,   `if (_hasRendered && Theme == IconThemeType.Twotone)`

ElderJames: I found that doing so caused a flash, showing the gray first and then the color would render twice.
icon-two-tone

icon-two-tone

rqx110: that because it render(grey) in prerender stage, and then re-render(twotone) in afterrender stage again.
If do not want this, I think the only choice is not render in prerender stage, it will cause another problem, it will display until after-render stage. 
ff1
ff2

rqx110:
**Note**
1. the flash problem already exist:
ff1


3. if we remove the `OnParametersSetAsync` function, it will not change color in runtime:
ff2

ElderJames:
@rqx110 Thanks for check. How about the last commit?

rqx110:
> @rqx110 Thanks for check. How about the last commit?

last commit is same result of above comment.

I think we can't fix the flash problem(show grey then colored quickly) currently.
but we need allow to change the color on runtime!

ElderJames (approved):

Issue #2496, opened 2022-06-02
DatePicker Clear-Icon Click is invalid


shuaqq2004:
https://antblazor.com/zh-CN/components/datepicker

Examples Basic:
![微信图片_20220602170743](https://user-images.githubusercontent.com/3870105/171596351-e3d95e83-c95c-4a41-bd2e-a38fae803c7f.png)

1. Select the dates of the two DatePicker controls
2. Click the Clear Icon of the first DatePicker control

The object value is changed, but the corresponding Input does not change!!

maybe needs to be called ```StateHasChanged```

Issue #2481, opened 2022-05-28
RangePicker, the input box doesn't switch back and forth


Breave-YY:
RangePicker uses Open to control the display and concealment, and the input box does not switch back and forth
![GIF 2022-5-28 11-51-17](https://user-images.githubusercontent.com/46343752/170808917-4efad068-b361-4d73-9a1a-e9ab90baff7d.gif)

Breave-YY:
Do not click on the input box, the first time you select the date to switch the start date and end date, the second click to show the date pop-up layer, click to select the date will not switch the input box, you can only select the start date

ElderJames:
Hi @Alexbits , could you pleae take a look into that?

Alexbits:
> Hi @Alexbits , could you pleae take a look into that?

Hey @ElderJames, I can confirm the issues with RangePicker behavior in certain scenarios. I will look into it further.

Issue #2441, opened 2022-05-12
datepicker rangepicker timepicker should show dropdown pane when click rightmost icon button


coader:
![image](https://user-images.githubusercontent.com/3700498/167985029-76bb12b9-0d3f-48b4-af24-dda220a1d4d3.png)

when click this button, dropdown pane should shown

Issue #1990, opened 2021-10-01
Not open datepicker on click icon


ardia:
### Describe the bug

![Not open datepicker on click icon](https://user-images.githubusercontent.com/5411424/135572945-c09cf3a3-45f7-40e1-8158-1645da6412f4.gif)

### Further technical details
- AntDesign Nuget Package version
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version

MutatePat:
I'll look into this, don't think this one will be too bad.

anddrzejb:
@MutatePat Any progress on this?

PR #2531, opened 2022-06-19
fix/date picker ok button

Alexbits:
### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
https://github.com/ant-design-blazor/ant-design-blazor/issues/2496
https://github.com/ant-design-blazor/ant-design-blazor/issues/2481
https://github.com/ant-design-blazor/ant-design-blazor/issues/2441
https://github.com/ant-design-blazor/ant-design-blazor/issues/1990

### 💡 Background and solution

This PR Makes Ok button behavior closer to the ant.design

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


Alexbits:
Added fix for https://github.com/ant-design-blazor/ant-design-blazor/issues/2441, https://github.com/ant-design-blazor/ant-design-blazor/issues/1990

Alexbits:
Hey @ElderJames I have a question regarding the unit test below. The code it covers prevents manually entered values from being applied. So when you manually enter a time value for example it always reverted back. Why?
 https://github.com/ant-design-blazor/ant-design-blazor/blob/57269582260049b491c7fa05a237f3110b1397ac/tests/AntDesign.Tests/DatePicker/DatePickerKeyboardInputTests.razor#L30-L49

The code it covers
https://github.com/ant-design-blazor/ant-design-blazor/blob/57269582260049b491c7fa05a237f3110b1397ac/components/date-picker/DatePicker.Razor.cs#L89-L138

I believe it should only be reverted when parsing fails and if this is not the case the  `_cacheDuringInput ` should be set to the parsed value so it will not be reverted during `OnBlur`

ElderJames:
> I believe it should only be reverted when parsing fails and if this is not the case the `_cacheDuringInput ` should be set to the parsed value so it will not be reverted during `OnBlur`

Hi @Alexbits , I agree with you.

PR #2570, opened 2022-07-16
fix:(module:rangepicker): date selection

Alexbits:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design-blazor/ant-design-blazor/issues/1429

### 💡 Background and solution

Fixes quarter/week range selection when the end value is set first.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/RangePicker.razor.cs
+++ b/components/date-picker/RangePicker.razor.cs
@@ -58,34 +58,49 @@ namespace AntDesign
             {
                 var array = Value as Array;
 
+                int? index = null;
+
                 if (_pickerStatus[0].IsValueSelected && _inputEnd.IsOnFocused)
                 {
-                    DateTime? value = null;
-                    GetIfNotNull(Value, 0, notNullValue =>
-                    {
-                        value = notNullValue;
-                    });
-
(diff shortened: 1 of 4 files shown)


ElderJames:
I found another issue that time out of range is not disabled.

image

Alexbits:
> I found another issue that time out of range is not disabled.
> 
> image

In the ant.design the time input is not disabled. Instead, they change their places when the start value is later than the end value or vice versa. I already implemented the behavior when hours, minutes, and seconds in the time input get disabled, and then I checked out the original ant.design and IDK now 🤷‍♂️ 

I found another bug where the keyboard input allows to enter the later value as the start of the range.

ElderJames (approved):
LGTM.

Issue #1551, opened 2021-05-18
Locale DataPicker error


carrisian:
Hello. The DatePicker control is not translated correctly. Also, I can't get the first day of the week to set in Monday and in the example doesn't work correctly. The days of the week always are in English. The word "now" is only translated if I don't show the time in another case doesn't work. This control is necessary for many applications. Is there any chance that You can fix it?

Thank you in advance

anddrzejb:
I believe the first day issue was discussed in PR #1054. But this was never addressed again I think. We could probably discuss this again and see if it is going to solve your problem. 
Could you be more specific when you wrote
> The DatePicker control is not translated correctly.

Which controls? A number of controls are actually translated using json files in [here](https://github.com/ant-design-blazor/ant-design-blazor/tree/master/components/locales). If there is a translation missing in your language, then you could submit a PR. Use either `en-US` or `zh-CN` for reference (not your desired language, as it may not have all properties).

ElderJames:
Hey @Alexbits , would you be interested in taking a look at this question?

PR #2571, opened 2022-07-16
fix:(module:datepicker): wrong day headers when FirstDayOfWeek!=Sunday

Alexbits:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Fixes the day headers order issue from https://github.com/ant-design-blazor/ant-design-blazor/issues/1551

The fix is not ideal maybe. Because if the user changes the order in `DateLocale.ShortWeekDays` this will still cause wrong day headers. So the order of days in `DateLocale.ShortWeekDays` should be preserved.

Anyways, it's better than what we have now.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames:
> The fix is not ideal maybe. Because if the user changes the order in DateLocale.ShortWeekDays this will still cause wrong day headers. So the order of days in DateLocale.ShortWeekDays should be preserved.

@Alexbits I think we need to add some notes to the document.

Alexbits:
> > The fix is not ideal maybe. Because if the user changes the order in DateLocale.ShortWeekDays this will still cause wrong day headers. So the order of days in DateLocale.ShortWeekDays should be preserved.
> 
> @Alexbits I think we need to add some notes to the document.

I'm not sure where @ElderJames?

ElderJames:
Here site\AntDesign.Docs\Demos\Components\DatePicker\doc\index.en-US.md

ElderJames:
@Alexbits Please solve the conflicts

ElderJames (approved):

Issue #2649, opened 2022-08-29
Remove/Refactor unusable code in Breadcrumb


kooliokey:
### Is your feature request related to a problem? Please describe.
I started looking at writing unit tests for Breadcrumb/BreadcrumbItem and noticed there is code that is not currently able to be used by anyone. The `AutoGenerate` functionality in Breadcrumb will never be used. Even if someone sets that parameter to true, they cannot set the `_breadcrumbs` that would be rendered so it is unusable as-is. 

```
<CascadingValue Value="this" IsFixed="@true">
    <div class="@ClassMapper.Class" style="@Style" id="@Id" @ref="Ref">
        @ChildContent
        @if (AutoGenerate)
        {
            foreach (var breadcrumb in _breadcrumbs)
            {
                <BreadcrumbItem>
                    <a href="@breadcrumb.Url" @onclick="(e)=>Navigate(breadcrumb.Url.ToString())">@breadcrumb.Label</a>
                </BreadcrumbItem>
            }
        }
    </div>
</CascadingValue>
```

I assume this is from porting functionality from the react version? Will this code eventually be used and should stay or should we refactor it or remove it until then?

### Describe the solution you'd like
My vote would be to remove this functionality until it can be used. My guess is it came from the router functionality in the react library. I would also vote this route because the documentation doesn't currently mention that property. If we don't want to remove it, perhaps we should make `_breadcrumbs` a parameter that can be passed and the items autogenerated from that.

### Additional context
Related files:
https://github.com/ant-design-blazor/ant-design-blazor/blob/master/components/breadcrumb/Breadcrumb.razor.cs
https://github.com/ant-design-blazor/ant-design-blazor/blob/master/components/breadcrumb/Breadcrumb.razor

ElderJames:
We can remove it for a while, but it does make sense to implement it. But we need to get the routing list of the Blazor, which is more difficult.

Issue #2644, opened 2022-08-27
Test Coverage Improvement


ElderJames:
We plan to release the official 1.0 version when the coverage reaches the desired value > 90%.

The following are all of the components currently in ant-design. Let's add test for them together.

- [ ] affix  
- [ ] alert 
- [ ] anchor 
- [ ] auto-complete  
- [ ] avatar  
- [ ] back-top  
- [ ] badge  
- [ ] breadcrumb
- [ ] button 
- [ ] calendar  
- [ ] card	 
- [ ] carousel 
- [ ] cascader 
- [ ] checkbox 
- [ ] collapse 
- [ ] comment  
- [ ] date-picker  
- [ ] descriptions  
- [ ] divider  
- [ ] drawer 
- [ ] dropdown 
- [ ] empty 
- [ ] form 
- [ ] grid 
- [ ] icon  
- [ ] input-number 
- [ ] input 
- [ ] layout 
- [ ] list  
- [ ] mentions  
- [ ] menu 
- [ ] message 
- [ ] modal  
- [ ] notification 
- [ ] page-header  
- [ ] pagination 
- [ ] popconfirm  
- [ ] popover 
- [ ] progress  
- [ ] radio 
- [ ] rate  
- [ ] result 
- [ ] segmented
- [ ] select 
- [ ] skeleton  
- [ ] slider 
- [ ] space 
- [ ] spin 
- [ ] statistic 
- [ ] steps 
- [ ] switch 
- [ ] table 
- [ ] tabs 
- [ ] tag 
- [ ] time-picker 
- [ ] timeline 
- [ ] tooltip 
- [ ] transfer 
- [ ] tree-select 
- [ ] tree 
- [ ] typography 
- [ ] upload

kooliokey:
@ElderJames That PR was for Alert that just needs moved in the list from affix

ElderJames:
@kooliokey Fixed. Thank you.

ElderJames:
Hello @kooliokey , would you like to continue to help the improvement?

ElderJames:
Hello @kooliokey , we still need to improve coverage, we'd appreciate it if you were available to help! 😊

ElderJames:
We also can use Playwright to setup E2E tests for the complex components, do you have any experience with this?  @kooliokey

kooliokey:
> We also can use Playwright to setup E2E tests for the complex components, do you have any experience with this?  @kooliokey 

[...]

Issue #2540, opened 2022-06-26
AntDesign Breadcrumb component does not have a property matching the name 'OnClick'.


soomon:
According to [https://antblazor.com/en-US/components/breadcrumb#API](https://antblazor.com/en-US/components/breadcrumb#API), a breadcrumb component should allow for an `OnClick` handler property in the breadcrumb item.

Since there is no example on the website, I tried it myself:

```
<Breadcrumb>
    <BreadcrumbItem OnClick="@RedirectToDashboard"><Icon Type="home" /><span>Home</span></BreadcrumbItem>
    <BreadcrumbItem Href="">Application Center</BreadcrumbItem>
    <BreadcrumbItem Href="">Application List</BreadcrumbItem>
    <BreadcrumbItem>An Application</BreadcrumbItem>
</Breadcrumb>

@code {
        
    private void RedirectToDashboard()
    {
        redirector.RedirectToDashboard();
    }
}
```

I get an `InvalidOperationException`

```
InvalidOperationException: Object of type 'AntDesign.BreadcrumbItem' does not have a property matching the name 'OnClick'.
Microsoft.AspNetCore.Components.Reflection.ComponentProperties.ThrowForUnknownIncomingParameterName(Type targetType, string parameterName)
Microsoft.AspNetCore.Components.Reflection.ComponentProperties.SetProperties(ref ParameterView parameters, object target)
Microsoft.AspNetCore.Components.ComponentBase.SetParametersAsync(ParameterView parameters)
Microsoft.AspNetCore.Components.Rendering.ComponentState.SupplyCombinedParameters(ParameterView directAndCascadingParameters)
Microsoft.AspNetCore.Components.Rendering.HtmlRenderer.HandleException(Exception exception)
Microsoft.AspNetCore.Components.RenderTree.Renderer.HandleExceptionViaErrorBoundary(Exception error, ComponentState errorSourceOrNull)
Microsoft.AspNetCore.Components.RenderTree.RenderTreeDiffBuilder.InitializeNewComponentFrame(ref DiffContext diffContext, int frameIndex)
Microsoft.AspNetCore.Components.RenderTree.RenderTreeDiffBuilder.InitializeNewSubtree(ref DiffContext diffContext, int frameIndex)
Microsoft.AspNetCore.Components.RenderTree.RenderTreeDiffBuilder.InsertNewFrame(ref DiffContext diffContext, int newFrameIndex)
[...]

PR #2655, opened 2022-09-01
Add OnClick parameter to BreadcrumbItem, Add Tests, Remove unusable code

kooliokey:
### 🤔 This is a ...

- [X] New feature
- [X] Bug fix
- [ ] Demo update
- [ ] Component style update
- [X] Refactoring
- [ ] Code style optimization
- [X] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#2540 - OnClick
#2649 - Remove unusable code
#2644 - Test coverage
#2658 - Wrong HTML markup causing last separator to display

### 💡 Background and solution
- A user discovered we document an OnClick parameter for BreadcrumbItem, but it was not implemented. Now it is. Also added to the demo.
- Discovered when testing there is unreachable/unusable code for future features. Removed for now. Left public API alone but added comments about currently not being used.
- Test coverage for Breadcrumb/Item
- Update the markup of Breadcrumb/BreadcrumItem to match Ant.Design React

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add OnClick parameter to BreadcrumbItem. Markup of Breadcrumb updated to match Ant.Design React. This could break custom CSS targeting this component's resulting markup. |
| 🇨🇳 Chinese | 将 OnClick 参数添加到 BreadcrumbItem. Breadcrumb 的标记已更新以匹配 Ant.Design React。这可能会破坏针对此组件生成标记的自定义 CSS |

### DIFF
--- a/components/breadcrumb/BreadcrumbItem.razor
+++ b/components/breadcrumb/BreadcrumbItem.razor
@@ -1,8 +1,7 @@
 @namespace AntDesign
 @inherits AntDomComponentBase
 
-<span @ref="Ref" @onclick="@_onClick">
-
+<li @ref="Ref">
 	@if (Overlay != null)
 	{
 		<Dropdown Placement="@Placement.Bottom">
@@ -21,7 +20,7 @@
 	}
 	else
 	{
-		<span class="ant-breadcrumb-link">
+        <span class="ant-breadcrumb-link" @onclick="@_onClick">
 			@FormattedChildContent
 		</span>
 	}
@@ -32,8 +31,7 @@
 			@BreadCrumb.Separator
 		</span>
 	}
-
-</span>
+</li>
 
 @code {
     [Parameter]
--- a/components/breadcrumb/BreadcrumbItem.razor
+++ b/components/breadcrumb/BreadcrumbItem.razor
@@ -52,8 +52,7 @@
     [CascadingParameter]
     public Breadcrumb BreadCrumb { get; set; }
 
-	private RenderFragment FormattedChildContent => Href is null 
+    private RenderFragment FormattedChildContent => Href is null 
         ? ChildContent 
-        : @<a href="@Href">@ChildContent</a>
-	;
+        : @<a href="@Href">@ChildContent</a>;
 }
--- a/components/breadcrumb/BreadcrumbItem.razor
+++ b/components/breadcrumb/BreadcrumbItem.razor
@@ -1,7 +1,7 @@
 @namespace AntDesign
 @inherits AntDomComponentBase
 
-<span @ref="Ref">
+<span @ref="Ref" onclick="@OnClick">
 
 	@if (Overlay != null)
 	{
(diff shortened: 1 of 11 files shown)


components/breadcrumb/BreadcrumbItem.razor
```diff
@@ -37,20 +37,22 @@
...
 
-	[Parameter]
-	public string Href { get; set; }
+    [Parameter]
+    public string Href { get; set; }
 
-	[CascadingParameter]
-	public Breadcrumb BreadCrumb { get; set; }
+    [Parameter]
+    public EventCallback<MouseEventArgs> OnClick { get; set; }
```
ElderJames, line 50: I think it's better to pass in the current `BreadcrumbOption`.

kooliokey: Should we pass back the `BreadcrumbItem` itself? From what I could tell, the `BreadcrumbOption` class is not used anywhere yet. I just didn't remove it because it was public.

ElderJames: Oh, that fine, but it's better to pass some info of current item to user

kooliokey: Yeah, makes sense. I went ahead and did a `Tuple<MouseEventArgs, BreadcrumbItem>` so the user has access to the click arguments and the item since both have useful information. How does that look?

ElderJames: That's fine. Do you plan to fix HTML differences(#2658) here?

kooliokey: I was thinking of doing that in a separate PR after these changes get in. Would that just be PR'd to master or somewhere else since it will completely change the HTML for existing users? Wasn't sure if that would be considered a breaking change or not.

ElderJames: I think this is a breaking change for CSS, not for HTML or Blazor. CSS has been updated in the latest version, but HTML hasn't caught up.

kooliokey: It was a pretty small change to update the HTML markup so I went ahead and did that too. It is now in this PR.

ElderJames (approved):
LGTM

components/breadcrumb/BreadcrumbItem.razor
```diff
@@ -32,25 +31,32 @@
...
-	public string Href { get; set; }
+    [Parameter]
+    public EventCallback<Tuple<MouseEventArgs, BreadcrumbItem>> OnClick { get; set; }
 
-	[CascadingParameter]
-	public Breadcrumb BreadCrumb { get; set; }
+    [CascadingParameter]
+    public Breadcrumb BreadCrumb { get; set; }
 
+    private async void _onClick(MouseEventArgs args)
```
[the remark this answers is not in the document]
kooliokey: Ok. I see you changed the signature of the OnClick to just be MouseEventArgs now though. This method isn't needed anymore then since you can directly pass OnClick to the element.

PR #2686, opened 2022-09-06
fix(module TreeSelect / Tree): search not working on TreeSelect

Magehernan:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Enabling Search on TreeSelect not working.

### 💡 Background and solution

When you try to use search on tree select it not work, the code is like the search in select, so base on that i include the part that was missing and make it work.

At the same time i impove the search on tree because it was case sensitive. I make it work with ignore case and found that the code that apply styles to titles was using the input so i change that to use the original value and insert the span with class and style to make it work.

### 📝 Changelog

Fix search on treeselect 
Change tree search  to ignoreCase and fix display of match to respect the original value.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tree-select/TreeSelect.razor
+++ b/components/tree-select/TreeSelect.razor
@@ -45,6 +45,10 @@
                         ChildrenExpression="ChildrenExpression"
                         DisabledExpression="DisabledExpression"
                         DefaultExpandAll="TreeDefaultExpandAll"
+                        SearchExpression="SearchExpression"
+                        MatchedStyle="@MatchedStyle"
+                        MatchedClass="@MatchedClass"
+                        SearchValue="@_searchValue"
                   >
                     <Nodes>
                       @if (IsTemplatedNodes)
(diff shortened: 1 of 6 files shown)


ElderJames:
@Magehernan Thanks for contribute this. IMO this is a feature for tree-select.

Magehernan:
@ElderJames haha, i wasn't sure and i opted for fix because TreeSelect has EnableSearch but it not work so for that maybe was a fix... i don't know the specification of this feature so maybe it covers the requirements. I hope it helps.

ElderJames:
@Magehernan A searching demo for TreeSelect is need.

Magehernan:
@ElderJames i added one using the multiple example, i use google translate for chinise translation, please check if it is valid.

ElderJames (approved):
LGTM

PR #2761, opened 2022-10-01
Make the "AntDesign.Docs.Wasm" be able to pre-render with the "BlazorWasmPreRendering.Build" NuGet package.

jsakamoto:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [X] Other (about what?)

### 🔗 Related issue link

- https://github.com/jsakamoto/BlazorWasmPreRendering.Build/issues/2#issuecomment-1263303118

### 💡 Background and solution

Currently, the "AntDesign.Docs.Wasm" can not be pre-rendered to static HTML files with the "BlazorWasmPreRendering.Build" NuGet package. Some of the reasons for it come from the "AntDesign.Docs" project not being implemented to run on not only Blazor WebAssembly but also Blazor Server. (Pre-rendering mecanism by the "BlazorWasmPreRendering.Build" NuGet package is based on the standard pre-rendering feature of Blazor Server.)

The changes included in this pull request will fix those problems above, and it will be able to pre-render into static HTML files at publishing time with the "BlazorWasmPreRendering.Build" NuGet package via the dotnet CLI command like below. 

```
dotnet publish .\site\AntDesign.Docs.Wasm -c Release -f net6
```

### 📝 Changelog

* Fix: configure pre-rendering to fetch the URLs that cannot be reached from the root HTML.
    - Pre-renderer will start traversing `<a>` links from root ("/") HTML document. But on this app, the root HTML document doesn't have any `<a>` links such as docs, components, etc. It only works to "redirect" to language-specific URLs such as "en-US" or "zh-CN" by C# code. Therefore developer has to explicitly specify those URLs via MSBuild properties for the pre-renderer.

* Fix: incorrect service scope usage.
    - Using singleton scope means sharing the application state, such as language selection, with ALL USERS when the app runs as a Blazor Server app. This will cause mass confusion among users and will cause application crashes.

* Fix: don't invoke `NavigationManager.NavigateTo()` in the `OnInitializedAsync()` 
    - ... to make it be able to run as not only Blazor WebAssembly but also Blazor Server for pre-rendering.
    - Invoke it in the `OnAfterRenderAsync()` instead.
    - See also: https://github.com/dotnet/aspnetcore/issues/13582#issuecomment-527383902

* Fix: don't invoke JavaScript interop in the `OnInitializedAsync()`.
    - ... to make it be able to run as not only Blazor WebAssembly but also Blazor Server for pre-rendering.
    - Invoke it in the `OnAfterRenderAsync()` instead.
    - See also: https://learn.microsoft.com/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet#prerendering

### ⚠️ NOTICE

The pre-rendering process on the "AntDesign.Docs.Wasm" will **take a few minutes** because it will take over 2 sec to fetch each page. This slow process is caused by **an explicit 2 sec delay at OnInitializedAsync() in the "App.razor" of the "AntDesign.Docs" project**. 

- https://github.com/ant-design-blazor/ant-design-blazor/blob/master/site/AntDesign.Docs/App.razor#L26

I don't know why that explicit delay is needed. When I commented out the explicit delay line, the pre-rendering process was finished in a few seconds.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2761) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2761) before we can accept your contribution.<br/><br/>:x: jsakamoto [sign now](https://cla.dotnetfoundation.org/ant-design-blazor/ant-design-blazor?pullRequest=2761)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/ant-design-blazor/ant-design-blazor?pullRequest=2761) it.</sub>

ElderJames:
Thank you so much @jsakamoto , but I'm not sure why I opened the [preview page above](https://preview-2761-antblazor.surge.sh/en-US/) and it still took a while to show up. I loaded the pre-rendered html file

ElderJames:
Hello @jsakamoto , do you have any ideas?

jsakamoto:
@ElderJames Sorry to late! I'll reply with more detail later. Please give me time for a while.

jsakamoto:
@ElderJames 
>  but I'm not sure why I opened the preview page above and it still took a while to show up

This is by design, particularly with default configurations.

"static pre-rendering" of Blazor WebAssembly apps has a few levels.

The default and easiest level is static pre-rendering for only search engine optimization. That preview page is at this level for now. The pre-rendered HTML files definitely contain pre-rendered content, but that pre-rendered content is covered by "loading..." progress indicator content. So users never see the pre-rendered contents on a web browser until the Blazor WebAssembly engine has been started and renders the application component. However, search engine crawlers can read pre-rendered contents in that static pre-rendered HTML files, so this is meant full for the purpose of search engine optimization.


For example, please imagine the "Counter" page. If the "Loading..." progress contents didn't appear and didn't hide the static pre-rendered contents, users would see the "CLICK ME" button on the "Counter" page immediately. But if users clicked the "CLICK ME" button quickly, there would be nothing to happen because the Blazor WebAssembly engine has not been started. If users run into this behavior, the user must be confused. This is the reason why hide the pre-rendered contents by default from users until the Blazor WebAssembly app has been started.

But of course, we can take the next level of static pre-rendering. I'll explain how to do that later.
(Please see also: [_"Delete the "Loading..." contents"_](https://github.com/jsakamoto/BlazorWasmPreRendering.Build#delete-the-loading-contents))

ElderJames:
Thank you @jsakamoto ! My goal is still to speed up the first load and make the site look faster, although users won't be able to interact until wasm has finished loading

ElderJames:
Another problem was encountered when `/ ` was added to the end of route, causing both NavLink and `/Component/{Name}` to be unrecognized.

```
https://preview-2761-antblazor.surge.sh/zh-CN/components/table/
```

jsakamoto:
@ElderJames 
> My goal is still to speed up the first load and make the site look faster...

Sure. I guess you are already done, setting the `BlazorWasmPrerenderingDeleteLoadingContents` MSBuild property in your project file to `true` makes the pre-rendered static HTML files no longer be covered with the "Loading..." progress contents. In other words, users will see the pre-rendered content immediately after a web browser loads the HTML content. On the other hand, the "Loading..." progress contents are removed from all pre-rendered static HTML files.

But, as you already know, please keep in your mind that the user interactions will not work until the Blazor WebAssembly engine warm up. The responsibility to manage that is the developers of the app.
For example, the "ClickOnce Get" site ( https://clickonceget.azurewebsites.net/ ) that is built on Blazor WebAssembly and is static pre-rendered will show users immediately its pre-rendered contents without "Loading..." contents. But instead, all buttons are lockdowned until the Blazor WebAssembly app has started. (That lockdown animation effect is implemented by 100% pure CSS.) This is one of the examples of how to implement a static pre-rendered Blazor WebAssembly app without covering the entire screen by "Loading..." contents.
It will make sense, isn't it?

jsakamoto:
@ElderJames 

> Another problem was encountered when `/` was added

By default, the "BlazorWasmPreRendering.Build" saves pre-rendered contents to "foo/bar/index.html" for the routed URL path "foo/bar". This is by design intentionally to support general static web contents server. But this behavior causes the problem you said.

Instead, you can set the `BlazorWasmPrerenderingOutputStyle` MSBuild property in your project to "AppendHtmlExtension". After doing it, the pre-rendered contents for the routed URL path "foo/bar" will be saved to the static HTML file "foo/bar.html".

If your static contents web server can respond to the "foo/bar.html" HTML file for the HTTP request path "foo/bar", you can take this strategy, and the problem about NavLink will go away.

Pleas see also: [_"Output style"_](https://github.com/jsakamoto/BlazorWasmPreRendering.Build#output-style)

PR #2803, opened 2022-10-14
perf: Optimize `CssSizeLength` & `CssStyleBuilder`

LeaFrock:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Optimize `CssSizeLength` & `CssStyleBuilder` |
| 🇨🇳 Chinese | 优化`CssSizeLength` 和`CssStyleBuilder` |

### DIFF
--- a/components/core/CssSizeLength.cs
+++ b/components/core/CssSizeLength.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Globalization;
-using System.Linq;
+using System.Runtime.CompilerServices;
 
 namespace AntDesign
 {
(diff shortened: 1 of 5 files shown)


components/core/CssSizeLength.cs
```diff
@@ -145,5 +139,66 @@ public static bool TryParse(string value, out CssSizeLength cssSizeLength)
...
+            {
+                unit = CssSizeLengthUnit.NoUnit;
+                return true;
+            }
+
+            if (!IsNumberOrDot(trimmedValue[0]))
+            {
+                unit = CssSizeLengthUnit.Calc;
+                return true;
+            }
```
ElderJames, line 155: Is the `unit` result backwards?

LeaFrock: What does `backwards` mean? The logic should be the same as before.

```csharp
            if (value.StartsWith("calc", StringComparison.OrdinalIgnoreCase))
            {
                _stringValue = value;
                _value = null;
                _unit = CssSizeLengthUnit.NoUnit;
                return;
            }

            var index = value
                .Select((c, i) => ((char c, int i)?)(c, i))
                .FirstOrDefault(x => !(x.Value.c == '.' || x.Value.c >= '0' && x.Value.c <= '9'))
                ?.i
                ?? value.Length;

            if (index == 0) // The first char is not a number or dot. Or the value is an empty string.
            {
                _stringValue = value;
                _value = null;
                _unit = CssSizeLengthUnit.Calc;
                _parsed = false;
                return;
            }
```

ElderJames: I see, but we may be wrong before. If `trimmedValue` starts with `calc`, the unit should be `Calc`, and shoud be `NoUnit` while it not starts with number or dot.

LeaFrock: > but we may be wrong before

🤣 Really? Please make sure about it and if so, I'll make a commit for this. Also, we need to check if it's a lack of related unit tests which did not expose the issue before.

LeaFrock: The changes are related to [your commit](https://github.com/ant-design-blazor/ant-design-blazor/commit/c8b52156a051bb10e17c4719262bf6524cb03913#diff-0a92621cbff66fd618c8848c2cd86ad991bb9a4b20cddaab1b9ca0fabfa2ee06) @ElderJames  .

ElderJames: Logically, that's true. 😅

LeaFrock: Fixed.

ElderJames (approved):

PR #2816, opened 2022-10-19
Tag Tests & Updates

kooliokey:
- Add more tests for Tag. 
- Deprecate `PresetColor` in favor of using `Color`. 
    - This matches the react library better and prevents confusion. 
    - The `PresetColor` parameter is nullable but throws if you set it to null and it uses the logic in Color behind the scenes anyway.
- Update documentation for Tag to remove `PresetColor` to discourage future usage and indicate a default value for `Color`

### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [X] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [X] Refactoring
- [ ] Code style optimization
- [X] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#2644 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix bug with tag color change after render not always styling properly. |
| 🇨🇳 Chinese |           |


ElderJames:
Thanks for contribute this @kooliokey ! Please leave the PersetColor parameter as the same as Color. This will reduce breaking changes.

kooliokey:
> Thanks for contribute this @kooliokey ! Please leave the PersetColor parameter as the same as Color. This will reduce breaking changes.

Hi! The changes should not introduce any breaking changes and in fact they fix a bug I found with tag - if you have it set to default color, then to a preset, then back to default color it doesn't style properly currently. This change set fixed it. I think the only way it would be a breaking change is if someone had the `Obsolete` tag as an error in their builds. Do you want me to take that out?

Here is what it does now:
image

Click "Change Color"

image

Click "Change Color" again
image

With the code in this PR, the final step results in the correct styling as in step 1.

Sample Code:
image

ElderJames:
`Obsolete ` attribute is ok. 👍

ElderJames (approved):

ElderJames (approved):

PR #2818, opened 2022-10-19
test(module: slider) Add tests, Small refactors

kooliokey:
- Add tests for slider
- Remove commented out code. 
- Refactor checking a double against zero to check against a tight range around zero because it will fail sometimes. 
- Update aria attributes on handles to have proper values.

### 🤔 This is a ...

- [X] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [X] Refactoring
- [ ] Code style optimization
- [X] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#2644 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Slider accessibility updates with aria labels       |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/slider/Slider.razor.cs
+++ b/components/slider/Slider.razor.cs
@@ -420,27 +420,13 @@ namespace AntDesign
                     TValue defaultValue;
                     if (Range)
                     {
-                        //if (typeof(T) == typeof((int, int)))
-                        //{
-                        //    defaultValue = parameters.GetValueOrDefault(nameof(DefaultValue), DataConvertionExtensions.Convert<(int, int), T>((0, 0)));
-                        //}
-                        //else
-                        //{
                         defaultValue = parameters.GetValueOrDefault(nameof(DefaultValue), DataConvertionExtensions.Convert<(double, double), TValue>((0, 0)));
-                        //}
                         LeftValue = DataConvertionExtensions.Convert<TValue, (double, double)>(defaultValue).Item1;
                         RightValue = DataConvertionExtensions.Convert<TValue, (double, double)>(defaultValue).Item2;
                     }
                     else
                     {
-                        //if (typeof(T) == typeof(int))
-                        //{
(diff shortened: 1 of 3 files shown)


components/slider/Slider.razor.cs
```diff
@@ -477,6 +463,7 @@ protected override void OnParametersSet()
             ClassMapper.Clear()
                 .Add(PreFixCls)
                 .If($"{PreFixCls}-disabled", () => Disabled)
+                .If($"{PreFixCls}-horizontal", () => !Vertical)
```
kooliokey, line 466: Added this class because it is in the react version whenever it is horizontal. Didn't appear to affect any visuals for now though. Perhaps a style update will make use of it in the future and prevent a bug.

components/slider/Slider.razor.cs
```diff
@@ -527,23 +514,15 @@ private void ValidateParameter()
                 throw new ArgumentOutOfRangeException(nameof(Step), "Must greater than 0.");
             }
 
-            if (Step != null && (Max - Min) / Step % 1 != 0)
+            var minMaxStepComparison = (Max - Min) / Step % 1;
```
kooliokey, line 517: Is this logic necessary? I don't see it in the react library and the slider library that Ant react uses doesn't bother checking for this either. Did not doing it cause a bug before?

components/slider/Slider.razor.cs
ElderJames: Our current style is only synchronized to 4.20.7 (and many changes to the component html are not synchronized), and the latest one needs to check whether the component style is updated before merging. Could you please take a look at it?

https://github.com/ant-design-blazor/ant-design-blazor/pull/2814

components/slider/Slider.razor.cs
ElderJames: Maybe not.

ElderJames (approved):
LGTM

PR #2821, opened 2022-10-19
test(module: core): Add unit tests

LeaFrock:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Improve unit tests cover for `core` module     |
| 🇨🇳 Chinese |    增加核心模块的单元测试覆盖       |


LeaFrock:
>Failed AntDesign.Tests.button.ButtonTests.Button_animation_when_cliked_end_after_some_time

I don't know why the CI  fails. All tests work well in my local environment. I need your check for it, thanks! @ElderJames

components/core/Helpers/JsonElementHelper.cs
```diff
@@ -3,15 +3,13 @@
...
 
 using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Text.Json;
 
 namespace AntDesign.Core.Helpers
 {
-    public static class JsonElementHelper<TValue>
+    internal static class JsonElementHelper<TValue>
```
kooliokey, line 10: There are a couple classes that were changed to internal or private that I'd recommend keeping public. If someone is using them in their project it will break. I think making them internal is fine but in a breaking change release

LeaFrock: Thanks for your review!  `JsonElementHelper<TValue>` and `THelper` (what a name 🤣 ) are absolutely not suitable for public exposure as they're highly encapsulated helpers which aim to only serve for this lib and external users should not rely on them.

> I think making them internal is fine but in a breaking change release

You're right. The release note should notice this if it's approved. And for users who already use them, the workaround is easy: make themself sure they do want to use the codes , and copy them then.

Anyway, I need to discuss with @ElderJames . If he is more inclined to your opinion, I'll roll the metioned changes back.

ElderJames: This pr is better to focus on the test project first and not add changes.

LeaFrock: Well, I'll handle it tonight.

ElderJames (approved):
LGTM. It help a lot!

Issue #2823, opened 2022-10-20
Initialize Modal with maximization


coolmint123:
Dear Developers,

### Modal maximization specification

I have a button that maximizes Modal,
Press the maximize button and then maximize.
I want to initialize Modal with maximization.
Please consider it.

zxyao145:
That sounds very good, and I will try to add this feature.

PR #2834, opened 2022-10-23
Feat/modal maximization in initialize

zxyao145:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#2823

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    support the maximization of Modal initialization   |
| 🇨🇳 Chinese |     支持 Modal 初始化的时候最大化      |


components/modal/config/ModalOptions.cs
```diff
@@ -134,6 +134,12 @@ public bool ConfirmLoading
         /// </summary>
         public RenderFragment RestoreBtnIcon { get; set; } = DialogOptions.DefaultRestoreIcon;
 
+        /// <summary>
+        /// Maximize the Modal during component initialization, and it will ignore the Maximizable value.
+        /// </summary>
+        [Parameter]
+        public bool MaximizationOnInit { get; set; } = false;
```
ElderJames, line 141: It would be better to rename it `Maximized` instead, and user can use this parameter to toggle the maximization. Like visible.

zxyao145: Sorry, this may cause confusion of component status. Because one of them belongs to the property of the component and the other belongs to the state of the component. 

Like Visible, it cannot be modified within a component, but can only be passed from as a property.

ElderJames (approved):

Issue #2841, opened 2022-10-27
Found two bugs with Datepick when typing


PCichigo:
### Describe the bug

Hi, there : )

I just found two bugs with Datepick from recent version, I will separate these two bugs as case A and case B below.

=== CASE A

This bug only occurred typing date in Datepick input. 
For example, start by typing "2022-10-27" in a simple Datepick input box, and then click some where else or using Tab button to make it lose focus (so user can fill the other input next), the value will reverse back to the value before changing. 
This seem to occurred when typing, I found no problem when clicking date on calendar, or using Enter button after typing.

![datebug20221027a](https://user-images.githubusercontent.com/102783349/198283107-7191616c-b03f-4f32-8986-ae3f2caefa57.jpg)

This also happen no matter what the start value of the Datepick is. As you can see from the image below.

![datebug20221027b](https://user-images.githubusercontent.com/102783349/198283880-f5194f82-f006-4919-b0ae-3cb85dca6592.jpg)

=== CASE B

Ok, this one is kinda weird to find. 
First, we start with any value in a simple Datepick input box (or null), then changing the value by typing in the input box, so that the graphic of calendar is changing (this can be done changing year or month, bug will not occurred if changing day only). 
After finish the typing, click on any date in the calendar (including the same day you type), the result value will be wrong (it take the day from the location where before the calendar graphic change).

This seem to only occurred if the calendar graphic change by typing.

![datebug20221027c](https://user-images.githubusercontent.com/102783349/198283285-7ded2a19-a539-49cd-8a96-0c92e3095f0f.jpg)

### Steps to reproduce (please include code)

Both case are using the same basic code from DatePicker tutorial as follow:

`<DatePicker TValue="DateTime?" Picker="@DatePickerType.Date" />`

Step to reproduce, can be found in description and images above.

### Exceptions (if any)

No error exception.

### Further technical details
- AntDesign version 0.12.6
- Blazor Server, C# .net 6
- VS 2022 Community, version 17.3.3

ElderJames:
Thanks for contacting us @PCichigo , If @Alexbits has time, maybe he can help to look into the issue.

Alexbits:
Hey @PCichigo. Thank you for submitting the issue.

Case A is by design. Please check the original [ant.design](https://ant.design/components/date-picker/) behavior as a reference. You should press Enter key to confirm your input.

Case B looks like a bug. I will check it.

PCichigo:
[...]

PR #2847, opened 2022-10-30
fix:(module: datepicker): tab key does not confirm the value

Alexbits:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#2841

### 💡 Background and solution

This fixes two issues from #2841. 

Also, it brings RangePicker's behavior closer to the original antdesign. It gives the ability to select the range values more conveniently. Now it's possible to select the start date later than the end date without clearing the value.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/DatePicker.Razor.cs
+++ b/components/date-picker/DatePicker.Razor.cs
@@ -80,8 +80,6 @@ namespace AntDesign
             }
         }
 
-        private DateTime? _cacheDuringInput;
-
         protected void OnInput(ChangeEventArgs args, int index = 0)
         {
             if (index != 0)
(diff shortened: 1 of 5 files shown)


ElderJames:
Could we also support the short cut for date/month/year navigate?

Alexbits:
> Could we also support the short cut for date/month/year navigate?

@ElderJames What shortcut do you mean?

ElderJames:
@Alexbits Just use keyboard to choose date

ElderJames (approved):

Alexbits:
> @Alexbits Just use keyboard to choose date

@ElderJames I will look into this later. The tab key should move the input focus to the datetimepanel, and this should allow selecting days with the keyboard.  I will look into this later.

ElderJames:
@Alexbits That's great! Thank you.

YingXiaoMing:
> this

how to use focus, it press tab can skip,the datepicker be empty

PR #2854, opened 2022-11-03
fix:(module:datepicker): smoothScrollTo causes an infinite loop

Alexbits:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

The smoothScrollTo function continued to loop when it shouldn't causing various issues with the selected time animation in the DatePicker/RangePicker with time input. I replaced it with the ts function from the picker used in the ant.design. This issue should be hot-fixed ASAP.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/JsInterop/modules/dom/manipulationHelper.ts
+++ b/components/core/JsInterop/modules/dom/manipulationHelper.ts
@@ -5,7 +5,7 @@ import * as enums from '../enums';
 
 
 let cachedScrollBarSize: number | undefined = undefined;
-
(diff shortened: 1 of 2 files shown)


ElderJames (approved):
Thanks.

PR #2916, opened 2022-11-29
Table: Add sorting for RemoteDataSource

sebastianmattar:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

We are extensively using tables with remote data sources for our application. It was not possible to use the sorting function for this scenario as it was tightly coupled to the in-memory model.
To enable sorting for remote datasources without breaking current code, I added the RemoteFieldName paramer for the Column component. This causes sorting changes to be provided to the Table.OnChange event that allows sorting on the remote side.

### 📝 Changelog

As this is an addition, existing behaviour should not change.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -15,6 +15,13 @@ timeline: true
 
 ---
 
+### 0.13.1
+
+`2022-11-29`
+
+- 🐞 Fixed Input that when its value is changed in code, it would rollback on click. [#2906](https://github.com/ant-design-blazor/ant-design-blazor/pull/2906) [@ElderJames](https://github.com/ElderJames)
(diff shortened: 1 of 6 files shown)


ElderJames:
Thanks for contribute this @sebastianmattar . 
But I don't quite understand what's being solved here. Does the current remote data source not work?

https://antblazor.com/en-US/components/table#components-table-demo-ajax

Issue #2915, opened 2022-11-29
sentry.io a datepick error collected


GlodenBoy:
version: 0.13.0

I use sentry in the client to collect error logs and have recently received frequent errors for a date fetch component.

Steps to reproduce:
Click on the date display drop-down box and it will appear, strangely, I did not find an error in the console, but it was intercepted by sentry.

<img width="534" alt="6bdf72495ed9cc9be3b41a25c148423" src="https://user-images.githubusercontent.com/2329528/204555439-cdc6ee88-f757-40d6-a812-fa6ee20973be.png">

The messagebox below is the message I intercepted sending an error report

<img width="629" alt="c67ec766724c84d525bb7bff1321a95" src="https://user-images.githubusercontent.com/2329528/204555482-4f838a05-eba0-4962-b86d-548c11c2cd90.png">

at System.Nullable`1.get_Value()
   at AntDesign.Internal.DatePickerDatetimePanel`1.ScrollToSelectedHourAsync(Nullable`1 duration)

Alexbits:
@GlodenBoy Thanks. Fixed.

PR #2920, opened 2022-11-30
fix:(module:datepicker): Exception on input with time when Value is null

Alexbits:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#2915

### 💡 Background and solution

Fixes InvalidOperationException in the DatePicker/RangePicker (with time input enabled) that happens after clicking the input field when Value is null.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/internal/DatePickerDatetimePanel.razor.cs
+++ b/components/date-picker/internal/DatePickerDatetimePanel.razor.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
@@ -96,7 +96,6 @@ namespace AntDesign.Internal
             List<int> disabledHours = new List<int>();
             List<int> disabledMinutes = new List<int>();
             List<int> disabledSeconds = new List<int>();
-            DatePickerDisabledTime userDisabledTime = null;
 
             if (DisabledHours is not null)
             {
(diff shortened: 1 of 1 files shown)


components/date-picker/internal/DatePickerDatetimePanel.razor.cs
```diff
@@ -195,26 +194,30 @@ private async Task ScrollToSelectedTimeIfChangedAsync()
...
-                _ = ScrollToSelectedMinuteAsync(0);
-                _ = ScrollToSelectedSecondAsync(0);
+                return;
             }
+
+            var currentValue = Value.Value;
+
+            _ = ScrollToSelectedHourAsync(currentValue, 0);
+            _ = ScrollToSelectedMinuteAsync(currentValue, 0);
+            _ = ScrollToSelectedSecondAsync(currentValue, 0);
```
ElderJames, line 220: I think this can be combined into one interop

Alexbits: > I think this can be combined into one interop

Do you mean javascript interop?

ElderJames: That's right.

Alexbits: I'm unsure how to combine it in one javascript call and if it is worth it. This will require adding more javascript. From my point of view, the input panel re-renders itself way too often, which should be looked into. Also, maybe there is a way to integrate scroll in the time-panel-column rendering process 🤔

ElderJames (approved):

components/date-picker/internal/DatePickerDatetimePanel.razor.cs
ElderJames: Yes, we need to minimize the number of renderings.

PR #2947, opened 2022-12-14
fix(module: RangePicker): Always run default disabled logic, even with custom function provided

wss-kroche:
We ran into the need to use custom disabled logic, but also wanted to keep the default logic that kept the start/end in the proper order. 

~This PR adds a property to allow that setting which "mode" you want to use and then wraps the DisabledDate function provided in a method to check the type of check to run. It then either runs one or both.~

Through discussion below, this PR updates the disabled logic to always apply the default logic even when custom is provided.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Update RangePicker disabled date logic to always apply default logic even when custom is provided. This keeps ranges in the proper order even with custom disabled logic. |
| 🇨🇳 Chinese | 更新 RangePicker 禁用日期逻辑以始终应用默认逻辑，即使在提供自定义时也是如此。即使使用自定义禁用逻辑，这也会使范围保持正确的顺序。 |

### DIFF
--- a/components/date-picker/RangePicker.razor.cs
+++ b/components/date-picker/RangePicker.razor.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
 using AntDesign.Core.Extensions;
@@ -65,11 +65,28 @@ namespace AntDesign
 
         private bool ShowRanges => Ranges?.Count > 0;
 
+        private readonly Func<DateTime, bool> _defaultDisabledDateCheck;
+
+        private Func<DateTime, bool> _disabledDate;
+
+        [Parameter]
+        public override Func<DateTime, bool> DisabledDate
+        {
+            get
+            {
+                return _disabledDate;
+            }
+            set
+            {
+                _disabledDate = (date) => (value is not null && value(date)) || _defaultDisabledDateCheck(date);
+            }
+        }
+
         public RangePicker()
         {
             IsRange = true;
 
-            DisabledDate = (date) =>
+            _defaultDisabledDateCheck = (date) =>
             {
                 int? index = null;
 
@@ -116,6 +133,7 @@ namespace AntDesign
                     return index == 0 ? formattedDate1 < formattedDate2 : formattedDate1 > formattedDate2;
                 }
             };
+            DisabledDate = null;
         }
 
         private async Task OnInputClick(int index)
--- a/components/date-picker/internal/DatePickerBase.cs
+++ b/components/date-picker/internal/DatePickerBase.cs
@@ -236,7 +236,7 @@ namespace AntDesign
         public EventCallback<DateTimeChangedEventArgs> OnPanelChange { get; set; }
 
         [Parameter]
-        public Func<DateTime, bool> DisabledDate { get; set; } = null;
+        public virtual Func<DateTime, bool> DisabledDate { get; set; } = null;
 
         [Parameter]
         public Func<DateTime, int[]> DisabledHours { get; set; } = null;


ElderJames:
Thanks for contribution @wss-kroche , I think we can implement '[Disabled Date & Time](https://4x.ant.design/components/date-picker/#components-date-picker-demo-disabled-date)' as antd directly.

wss-kroche:
> Thanks for contribution @wss-kroche , I think we can implement '[Disabled Date & Time](https://4x.ant.design/components/date-picker/#components-date-picker-demo-disabled-date)' as antd directly.

I'm not sure I understand what you mean. Disabled date/time is implemented in ant and in ant blazor, but there is default logic for that disabled date for range picker that keeps the dates in the correct order. We need custom logic for disabled date, but we need the default as well because if you pick the dates in the wrong order the component completely breaks too. This is adding the option to override the default logic or use both.

site/AntDesign.Docs/Demos/Components/DatePicker/demo/RangePickerDemo.razor
```diff
@@ -11,13 +11,28 @@
 <RangePicker TValue="DateTime?[]" Picker="@DatePickerType.Quarter" />
 <br />
 <RangePicker TValue="DateTime?[]" Picker="@DatePickerType.Year" />
-
+<br />
 <RangePicker TValue="DateTime?[]" ShowTime='@("HH:mm")' OnChange="OnTimeRangeChange" />
+<br />
+<br />
+Custom Disabled Only:
+<br />
```
ElderJames, line 19: Please put the showcase to `DisabledDate.razor`.

ElderJames:
> > Thanks for contribution @wss-kroche , I think we can implement '[Disabled Date & Time](https://4x.ant.design/components/date-picker/#components-date-picker-demo-disabled-date)' as antd directly.
> 
> I'm not sure I understand what you mean. Disabled date/time is implemented in ant and in ant blazor, but there is default logic for that disabled date for range picker that keeps the dates in the correct order. We need custom logic for disabled date, but we need the default as well because if you pick the dates in the wrong order the component completely breaks too. This is adding the option to override the default logic or use both.

Ok, I see it. But I don't get the idea why do you need "Custom Disabled Only".

wss-kroche:
> > > Thanks for contribution @wss-kroche , I think we can implement '[Disabled Date & Time](https://4x.ant.design/components/date-picker/#components-date-picker-demo-disabled-date)' as antd directly.
> > 
> > 
> > I'm not sure I understand what you mean. Disabled date/time is implemented in ant and in ant blazor, but there is default logic for that disabled date for range picker that keeps the dates in the correct order. We need custom logic for disabled date, but we need the default as well because if you pick the dates in the wrong order the component completely breaks too. This is adding the option to override the default logic or use both.
> 
> Ok, I see it. But I don't get the idea why do you need "Custom Disabled Only".

Only to give flexibility to allow running custom, default or both. Do you think the default should always run? I wasn't sure if it should always run or not but if it should we can just make it always run and use the DisabledDate as an additional one to run. That would be a breaking change though and need to go to feature as well I think.

ElderJames:
> Do you think the default should always run? 

Yeah, I think that's more close to antd's behavior.

site/AntDesign.Docs/Demos/Components/DatePicker/demo/RangePickerDemo.razor
wss-kroche: Removed demo since it now uses default always

ElderJames (approved):
LGTM. Thanks!

Issue #1229, opened 2021-03-09
Transfer component does not pre-populate selected keys if async web service is called 


osbornj6:
### Describe the bug

I've tried to use the Transfer component after consuming an async web service, so that I can populate the selected key values from the data inside the web service response.  However, the target element of the transfer component is not populated in such cases.  If I simply comment the line out with the call to the async web service, it will behave as I expected it. 

### Steps to reproduce (please include code)

Using one of your examples and removing some elements of it, I came up with this sample .razor page code...

@page "/"
@inject HttpClient Http

Transfer example.

<div>
    <Transfer DataSource="dataSource"
              TargetKeys="targetKeys.ToArray()"
              ShowSearch="true"
              SelectedKeys="selectedKeys"
              Titles="titles"></Transfer>
</div>

@code {

    public MyClass myclass = new MyClass();

    private List<TransferItem> dataSource = new List<TransferItem>();
    private List<string> targetKeys = new List<string>();
    private string[] selectedKeys = Array.Empty<string>();
    private string[] titles = { "Source", "Target" };
    private bool Disabled = false;

    protected override async Task OnInitializedAsync()
    {
        // If the line below is commented out, it will work OK.  If it's not commented out, the target element is empty...
        myclass = await Http.GetFromJsonAsync<MyClass>("https://localhost:44341/api/MyClassService/123");

        var random = new Random();
        for (int i = 0; i < 20; i++)
        {
            var data = new TransferItem
            {
                Key = i.ToString(),
                Title = $"Content{i + 1}",
                Description = $"description of content{i + 1}",
                ["Chosen"] = random.Next(0, 100) * 2 > 100
            };

            if (data["Chosen"] is bool _bool && _bool)
            {
                targetKeys.Add(data.Key);
            }

            dataSource.Add(data);
        }
        base.OnInitialized();
    }

    private OneOf.OneOf<string, RenderFragment> Render(TransferItem item)
    {
        //return new RenderFragment(buildTree);
        return $"{item.Title} - {item.Description}";
    }

}

We will close this issue if:
- the repro project you share with us is complex. We can't investigate custom projects, so don't point us to such, please.
- if we will not be able to repro the behavior you're reporting
-->

### Exceptions (if any)
None

### Further technical details
- AntDesign Nuget Package version: v0.7.1
- Include the output of `dotnet --info`: 
- .NET SDK (reflecting any global.json):
 Version:   5.0.200
 Commit:    70b3e65d53

[...]

PR #2977, opened 2023-01-04
feet-bug(module: Transfer): refresh data on change SelectedKeys or TargetKeys parameters

Magehernan:
I added reinitializacion of values when TargetKeys or SelectedKeys change from outside so it properly display the correct values.
I am using this in production and works fine, and is more easy to use this way. I hope it helps.

### 🤔 This is a ...

- [X] New feature
- [X] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/transfer/Transfer.razor.cs
+++ b/components/transfer/Transfer.razor.cs
@@ -93,6 +93,31 @@ namespace AntDesign
                 .Add(PrefixName)
                 .If($"{PrefixName}-rtl", () => RTL);
 
+        }
+
+        public override async Task SetParametersAsync(ParameterView parameters)
+        {
+            var needRefresh = false;
+            if (parameters.TryGetValue(nameof(TargetKeys), out string[] newTargetKeys))
+            {
+                needRefresh |= TargetKeys != newTargetKeys;
+            }
+
+            if (parameters.TryGetValue(nameof(SelectedKeys), out string[] newSelectedKeys))
+            {
+                needRefresh |= SelectedKeys != newSelectedKeys;
+            }
+
+            await base.SetParametersAsync(parameters);
+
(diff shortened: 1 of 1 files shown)


ElderJames:
Thanks for contribution @Magehernan ! 

I think it would be better to refrsh data in the setter of these parameters.

ElderJames:
We also need to implement two-way binding, which is to update the external `SelectedKeys` and `TargetKeys` from within the component. Are you interested in helping?

Magehernan:
> I think it would be better to refrsh data in the setter of these parameters.
I try but the problem is it need all the other parameter set for it to work, so for it to work it needed a flag for set initialized and then OnInitialized call Refresh... When i do this, it call Refresh 2 times and with this implementation only is called 1 so more efficient. Let me know what you think.

> We also need to implement two-way binding, which is to update the external `SelectedKeys` and `TargetKeys` from within the component. Are you interested in helping?

i can try, let me know if you prefer all on this pull request or the two-binding in another.

ElderJames:
We can fix the two-way binding in another PR. In this PR we need to check if it fix https://github.com/ant-design-blazor/ant-design-blazor/issues/1229

Magehernan:
I almost sure that it will be fix because i am doing async request to the server for fill the targetKeys and i was having that problem that the targetkeys don't appear and with this fix, it works fine, but maybe there is another case.

ElderJames (approved):

PR #2986, opened 2023-01-05
fix(module: select): add tags when datasource is empty

### LINKED ISSUE #2965: Select component removes selected Tag options when Datasource is empty
### Describe the bug

If you have a `Select` component with `Mode` set to `tags`, and the `Datasource` is an empty collection, the user selected tag options will be hidden from the select option list. But they will still be registered as selected options.

For the below example, it starts with an empty datasource. It would be expected since the select component is tags mode, that it would still accept user selected partial tag options and display them. If you add a tag option (`12` in this example), it will still be set as the value from the 2-way binding. But the UI will show a thin empty box with no selected tag (image below).

image

Once the datasource is not empty, if you enter a different tag option (`14` in this example), it will then show all the ones that were previously hidden.

image

```
<Select Mode="tags"
        DataSource="@_items"
		@bind-Values="@_selectedValues"
		TItemValue="string"
		TItem="string"
		EnableSearch>
</Select>

<Button OnClick="@OnButtonClick">Click me to update datasource</Button>
@code
{
    List<string> _items = new List<string>();
    IEnumerable<string> _selectedValues;

    protected void OnButtonClick()
    {
        const int min = 10;
        const int max = 36;

        for (var i = min; max > i; i++)
        {
            var value = Convert.ToString(i, 16).PadLeft(2, '0') + i.ToString();
            _items.Add(value);
        }
    }
}
```

wss-javeney:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

fixes #2965 

### 💡 Background and solution

When there was a select component set to `Tags` mode and the datasource was empty, any entered tag options were being hidden from user even though they went through the two way binding. 

There was an issue where whenever the datasource was empty it would use `Clear()` on the option list instead of letting the change trickle down to where it updates the option list based on the datasource.  To ensure the user entered tags do not go away I add them after the option list is created.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Select mode Tags will retain tag options when datasource is empty   |
| 🇨🇳 Chinese |  選擇模式標籤將在數據源為空時保留標籤選項  |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -476,24 +476,6 @@ namespace AntDesign
                 return;
             }
 
-            // clear DataSource
-            if (DataSource?.Any() == false && SelectOptionItems.Any())
-            {
-                SelectOptionItems.Clear();
-                SelectedOptionItems.Clear();
-
-                Value = default;
-
-                _datasource = DataSource;
-                _dataSourceShallowCopy = new List<TItem>();
-                _dataSourceCopy = new List<TItem>();
-                TypeDefaultExistsAsSelectOption = false;
-
-                OnDataSourceChanged?.Invoke();
-
-                return;
-            }
-
             // DataSource maybe changed
             if (DataSource != null)
             {
@@ -647,12 +629,11 @@ namespace AntDesign
             if (SelectOptionItems.Any())
             {
                 // Delete items from SelectOptions if it is no longer in the datastore
-                for (var i = SelectOptionItems.Count - 1; i >= 0; i--)
+                foreach (var selectOption in SelectOptionItems)
                 {
-                    var selectOption = SelectOptionItems.ElementAt(i);
                     if (!selectOption.IsAddedTag)
                     {
-                        var exists = _datasource.Where(x => x.Equals(selectOption.Item)).FirstOrDefault();
+                        var exists = _datasource.FirstOrDefault(x => x.Equals(selectOption.Item));
 
                         if (exists is null)
                         {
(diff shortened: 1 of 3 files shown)


components/select/Select.razor.cs
```diff
@@ -476,24 +476,6 @@ private void EvaluateDataSourceChange()
...
-                _datasource = DataSource;
-                _dataSourceShallowCopy = new List<TItem>();
-                _dataSourceCopy = new List<TItem>();
-                TypeDefaultExistsAsSelectOption = false;
-
-                OnDataSourceChanged?.Invoke();
-
-                return;
-            }
-
```
ElderJames, line 496: Why delete this part of the code?

wss-javeney: Sorry for the delayed response, I will still respond just for documentation purposes...

 I removed this because when a Datasource is used to build the select option list...the "tags" get added to the SelectOptionItems hash set. So when the datasource is just empty, those were getting wiped out but the 2-way binding was not updating so they were still "selected" values. It is a weird edge case but I felt that it was valid for a user to have inputted "tag" options even if the datasource was empty.

ElderJames: Thank you for explaining. It makes sense.

Issue #2725, opened 2022-09-18
ReuseTabs close other tab rendering error


berkerdong:
![image](https://user-images.githubusercontent.com/36178221/190912590-f9f19155-e7b2-4985-97e7-1c325526da27.png)

_ReuseTabs component When there are many open Tabs, there will be a sliding effect, which is a very good implementation in itself, and it does a great job._

**ReuseTabs组件 当打开的Tabs比较多的时候，会出现滑动效果，这本身是很好的实现，做的很棒。**

---------------------------------------------------------------------------------------------------------------------------------------

_But there is a small problem. When one of the Tabs is selected and all other tabs are closed, the following figure will appear. It seems that the UI rendering is not recalculated when the action is performed._

**但是有一个小问题，当选中其中一个 Tabs 右击鼠标，点击除此之外全部关闭，就会出现如下图的情况，似乎在执行该动作的时候没有对UI呈现重新计算。**
![image](https://user-images.githubusercontent.com/36178221/190912613-f79b503f-c89c-4eea-a718-231e71f9e16d.png)

PR #3002, opened 2023-01-09
fix(module: tabs): ReuseTabs close other tab rendering error

berkerdong:
fix 2725 ReuseTabs close other tab rendering error

The original judgment logic of this module will result in a style rendering exception if the closed tab is not the currently selected tab

https://github.com/ant-design-blazor/ant-design-blazor/issues/2725


ElderJames (approved):
LGTM.

PR #3010, opened 2023-01-12
fix(module:select): Selected options will display even with HideSelected set to true when searching or clearing search

wss-kroche:
This can be seen in the "hide selected" demo on the documentation. If you select an option in the list, then type in a search that would include that selected option it displays - it should not. If you then remove the search it will still display. This PR fixes those issues.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix(module:select): Fix bug where selected options will display even with HideSelected set to true when searching or clearing search |
| 🇨🇳 Chinese | fix(module:select)：修复在搜索或清除搜索时即使将 HideSelected 设置为 true 也会显示所选选项的错误 |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -1209,7 +1209,7 @@ namespace AntDesign
             }
             else
             {
-                SelectOptionItems.Where(x => x.IsHidden).ForEach(i => i.IsHidden = false);
+                UnhideSelectOptions();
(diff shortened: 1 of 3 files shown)


ElderJames (approved):
LGTM

PR #3037, opened 2023-01-24
Protected selected items' selection sequences;

CuteLeon:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Issue: Select component can not present the sequence of each selected item.
For example: Select 2, 1 items in Select component, Select'd Values is 1, 2 (reversed sequence) instead of 2,1 (correct sequences)
Solution: Initlaize a private variable to fix the condition.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/SelectBase.cs
+++ b/components/select/SelectBase.cs
@@ -810,6 +810,7 @@ namespace AntDesign
 
             if (ValuesChanged.HasDelegate)
             {
+                _selectedValues = newSelectedValues;
                 await ValuesChanged.InvokeAsync(newSelectedValues);
             }
             else
--- a/components/select/SelectBase.cs
+++ b/components/select/SelectBase.cs
@@ -808,9 +808,9 @@ namespace AntDesign
                 }
             }
 
+            _selectedValues = newSelectedValues;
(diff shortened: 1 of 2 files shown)


components/select/SelectBase.cs
```diff
@@ -810,6 +810,7 @@ protected async Task InvokeValuesChanged(SelectOptionItem<TItemValue, TItem> new
 
             if (ValuesChanged.HasDelegate)
             {
+                _selectedValues = newSelectedValues;
```
ElderJames, line 813: It's better to put the assigment out of the 'if' block.

CuteLeon: sure, updated;

ElderJames (approved):
LGTM.

PR #3129, opened 2023-02-23
fix(module: select): OnSelectedItemsChanged is not triggered in form

### LINKED ISSUE #3126: 0.14.2 -> 0.14.3 Multiselect not working, Values is not set
### Describe the bug

When upgrading from 0.14.2 to 0.14.3 the multi select does not work any more. I think it was introduced by this [commit](https://github.com/ant-design-blazor/ant-design-blazor/commit/dac5055c48ca389bf3d5b65828218497597bb9ae). The `OnSelectedItemsChanged` is also triggered with an empty enumerable.

Using the code snippet from the demo page:
```
<Select Mode="multiple"
                Placeholder="Please select"
		@bind-Values="@_selectedValues1"
		TItemValue="string"
		TItem="string"
		OnSelectedItemsChanged="OnSelectedItemsChangedHandler"
		EnableSearch
		AllowClear>
		<SelectOptions>
			@foreach(var item in _items)
			{
				<SelectOption TItemValue="string" TItem="string" Value=@item Label=@item />
			}
		</SelectOptions>
</Select>
@code
{
    List<string> _items;
    IEnumerable<string> _selectedValues1;

    {
        const int min = 10;
        const int max = 36;
        _items = new List<string>();

        for (var i = min; max > i; i++)
        {
            var value = Convert.ToString(i, 16).PadLeft(2, '0') + i.ToString();
            _items.Add(value);
        }

		_selectedValues1 = new List<string> { "0a10" , "0c12"};
		_selectedValues2 = new List<string> { "0a10" , "0c12"};
    }

    private void OnSelectedItemsChangedHandler(IEnumerable<string> values)
    {
		if (values != null)
			Console.WriteLine($"selected: ${string.Join(",", values)}");
    }
}
```
### Further technical details
- AntDesign Nuget Package version 0.14.3
- dotnet 7.0.102
- Visual Studio 17.4.4

ElderJames:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

fixed #3126

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/SelectBase.cs
+++ b/components/select/SelectBase.cs
@@ -811,7 +811,15 @@ namespace AntDesign
                 }
             }
 
-            Values = newSelectedValues;
+            if (ValuesChanged.HasDelegate)
+            {
+                await ValuesChanged.InvokeAsync(newSelectedValues);
+            }
+            else
+            {
+                Values = newSelectedValues;
+                StateHasChanged();
+            }
         }
 
         protected void ClearSearch()
--- a/components/select/SelectBase.cs
+++ b/components/select/SelectBase.cs
@@ -451,6 +451,12 @@ namespace AntDesign
 
             EvaluateValuesChangedOutsideComponent(values);
 
+            if (ValuesChanged.HasDelegate)
+                await ValuesChanged.InvokeAsync(Values);
+
+            if (OnSelectedItemsChanged.HasDelegate)
+                await OnSelectedItemsChanged.InvokeAsync(SelectedOptionItems.Select(s => s.Item));
+
             if (_dropDown.IsOverlayShow())
             {
                 //A delay forces a refresh better than StateHasChanged().
@@ -460,12 +466,6 @@ namespace AntDesign
                 await Task.Delay(1);
                 await UpdateOverlayPositionAsync();
             }
-
-            if (ValuesChanged.HasDelegate)
-                await ValuesChanged.InvokeAsync(Values);
-
-            if (OnSelectedItemsChanged.HasDelegate)
-                await OnSelectedItemsChanged.InvokeAsync(SelectedOptionItems.Select(s => s.Item));
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


LuukGlorie:
@ElderJames 
I did some quick debugging and found out that the problem might be somewhere else in the code
```
protected async Task OnValuesChangeAsync(IEnumerable<TItemValue> values)
{
          if (!_isInitialized) // This is important because otherwise the initial value is overwritten by the EventCallback of ValueChanged and would be NULL.
          {
              return;
          }
          if (!SelectOptionItems.Any())
          {
              return;
          }
          if (values == null)
          {
              await ValuesChanged.InvokeAsync(default);
              await OnSelectedItemsChanged.InvokeAsync(default);
              return;
          }
          EvaluateValuesChangedOutsideComponent(values);
          if (_dropDown.IsOverlayShow())
          {
              //A delay forces a refresh better than StateHasChanged().
              //For example when a tag is added that is causing SelectContent to grow,
              //this Task.Delay will actually allow to reposition the Overlay to match
              //new size of SelectContent.
              await Task.Delay(1);
              await UpdateOverlayPositionAsync();
          }
          if (ValuesChanged.HasDelegate)
              await ValuesChanged.InvokeAsync(Values);
          if (OnSelectedItemsChanged.HasDelegate)
              await OnSelectedItemsChanged.InvokeAsync(SelectedOptionItems.Select(s => s.Item));
}
```
The `await Task.Delay(1);` causes the SetParameters to run again with the old value for `Values`, because the `ValuesChanged` has not yet been called. So maybe the problem is also solved by first calling the changed event handlers and then resizing the dropdown?

ElderJames:
@LuukGlorie Thank you, I have tried your way and it works. Please check the commit [cb944d6](https://github.com/ant-design-blazor/ant-design-blazor/pull/3129/commits/cb944d6b657163f3f2971712cb06b4d83e808d88)

Issue #3163, opened 2023-03-08
Add character format in Statistic


Alerinos:
The character format is missing in Statistic, different culture has different. For example, in Europe, 123 456,23 is popular
In the US, a comma is the same as a space in Europe. In Europe, a comma is a fractional number.

EU:
Number: 123 456 789
Price: 100,23

USA:
Number: 123,456,789
Price: 100.23

Alternatively, if the component depends on the built-in culture, it is worth mentioning it in the documentation:
https://antblazor.com/en-US/components/statistic

Alerinos:
I'm sorry but I didn't pay attention to ``GroupSeparator``, I think it's worth adding one example with this parameter to the documentation.

ElderJames:
Thanks for contacting us @Alerinos . Could you please to help adding the demo?

Alerinos:
@ElderJames Take a look if I have completed the documentation correctly, this is my first PR in this project.

PR #3165, opened 2023-03-09
Number separator usage

Alerinos:
No example discussed here: https://github.com/ant-design-blazor/ant-design-blazor/issues/3163

@dotnet-policy-service agree


Alerinos:
@dotnet-policy-service agree

PR #3166, opened 2023-03-09
Separator usage in Statistic Doc

Alerinos:
No example discussed here: https://github.com/ant-design-blazor/ant-design-blazor/issues/3163

@dotnet-policy-service agree

### DIFF
--- a/site/AntDesign.Docs/Demos/Components/Statistic/demo/Separator.razor
+++ b/site/AntDesign.Docs/Demos/Components/Statistic/demo/Separator.razor
@@ -1,0 +1,5 @@
+<Row Gutter="16">
+    <Col Span="24">
+        <Statistic Title="Active Users" Value="123456" DecimalSeparator=" " />
+    </Col>
+</Row>
--- a/site/AntDesign.Docs/Demos/Components/Statistic/demo/Separator.razor
+++ b/site/AntDesign.Docs/Demos/Components/Statistic/demo/Separator.razor
@@ -1,5 +1,4 @@
-<Row Gutter="16">
-    <Col Span="24">
-        <Statistic Title="Active Users" Value="123456" DecimalSeparator=" " />
-    </Col>
+<Row Type="flex">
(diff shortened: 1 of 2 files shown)


site/AntDesign.Docs/Demos/Components/Statistic/demo/Separator.razor
```diff
@@ -0,0 +1,4 @@
+<Row Type="flex">
+    <Statistic Title="Number" Value="123456" GroupSeparator=" " Style="text-align: center;"></Statistic>
+    <Statistic Title="Decimal" Value="50.5" DecimalSeparator=" " Style="margin: 0 32px; text-align: center;"></Statistic>
```
ElderJames, line 3: Why did you set the separators as a space?

Alerinos: https://learn.microsoft.com/pl-pl/dotnet/api/system.globalization.numberformatinfo.numberdecimalseparator?view=net-7.0
Depending on the culture, we have different separators. For me it is a space in Poland/Europe. You can think about embedding the default culture value in the component if there is no separator.

ElderJames: Oh, I see. Therefor we maybe can use ` LocaleProvider.CurrentLocale.CurrentCulture.NumberFormat.NumberDecimalSeparator` and `LocaleProvider.CurrentLocale.CurrentCulture.NumberFormat.NumberGroupSeparator` as default value.

Alerinos: @ElderJames 
This can be done in the second PR. What with that? I think the documentation should be updated with an additional example.

ElderJames: That's fine.

ElderJames (approved):

PR #3216, opened 2023-04-04
fix(module:table): fix not logging exception and ignore JSDisconnectedException

### LINKED ISSUE #3215: Exception is not logged in Table DisposeAsync exception handler
### Describe the bug
In the exception handler for Table DisposeAsync the wrong overload for LogError is used. Instead of `LogError(ILogger, Exception, String, Object[])` [Link](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logerror?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-loggerextensions-logerror(microsoft-extensions-logging-ilogger-system-exception-system-string-system-object())) the method `LogError(ILogger, String, Object[])` is used, so the exception is not logged in this way.

`Logger.LogError("AntDesign: an exception was thrown at Table 'DisposeAsync' method.", ex);`

### Further technical details
- AntDesign Nuget Package version 14.4

LuukGlorie:
Thrown exception was not logged because LogError overload was used incorrectly. When calling JS while disposing a component a JSDisconnectedException could be thrown, which should be ignored.

Closes #3215

### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#3215 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/table/Table.razor.cs
+++ b/components/table/Table.razor.cs
@@ -9,6 +9,7 @@ using AntDesign.JsInterop;
 using AntDesign.TableModels;
 using Microsoft.AspNetCore.Components;
 using Microsoft.Extensions.Logging;
+using Microsoft.JSInterop;
 
 namespace AntDesign
 {
@@ -666,7 +667,7 @@ namespace AntDesign
             base.Dispose(disposing);
         }
 
-        public async ValueTask DisposeAsync()
+        async ValueTask IAsyncDisposable.DisposeAsync()
         {
             try
             {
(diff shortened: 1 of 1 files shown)


LuukGlorie:
@dotnet-policy-service agree

components/table/Table.razor.cs
```diff
@@ -666,7 +667,7 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
         }
 
-        public async ValueTask DisposeAsync()
+        async ValueTask IAsyncDisposable.DisposeAsync()
```
ElderJames, line 670: Dose we need this change?

ElderJames (approved):

PR #3221, opened 2023-04-07
add `SimplifySearchResults ` to hide the nodes without any matched parent or child from the search results

huangjia2107:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  add `SimplifySearchResults` to hide the nodes without any matched parent or child from the search results  |
| 🇨🇳 Chinese |  添加 `SimplifySearchResults` 来简化搜索结果的显示，隐藏父节点或子节点均无匹配的节点  |

### DIFF
--- a/components/tree/Tree.razor.cs
+++ b/components/tree/Tree.razor.cs
@@ -429,6 +429,12 @@ namespace AntDesign
         [Parameter]
         public string MatchedClass { get; set; }
 
+        /// <summary>
+        /// Determine whether to simplify the search results
+        /// </summary>
+        [Parameter]
+        public bool SimplifySearchResults { get; set; }
+
         private void SearchNodes()
         {
             var allList = _allNodes.ToList();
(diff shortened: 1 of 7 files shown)


components/tree/style/mixin.less
```diff
@@ -1,4 +1,4 @@
-@import '../../style/mixins/index';
+﻿@import '../../style/mixins/index';
```
ElderJames, line 1: Please fix the encoding.

components/tree/style/mixin.less
ElderJames, line 1: Please rename the file with 'patch.less'.

components/tree/style/mixin.less
huangjia2107: The file encoding is changed from utf-8 to utf-8 with bom，so the changing chars is �ｿ

components/tree/style/mixin.less
huangjia2107: fixed in the latest commit

ElderJames (approved):

Issue #3235, opened 2023-04-21
Refactor visibility (Change internal to public/protected internal)


rhodon-jargon:
### Is your feature request related to a problem? Please describe.
Multiple times while using this library I wanted to customize some components in ways that the existing component's attributes didn't allow for. When I then try to either inherit from that component, or create a new component reusing some of the original components code/internal components, I always find that it is impossible because a lot of the internally used methods (including virtual methods) are marked internal.

While I understand that most of these properties shouldn't be part of the consuming API, this makes it basically impossible to make custom versions of these components while reusing the code from this library.

### Describe the solution you'd like
I understand that this can't be done very quickly, but I'd like all properties and fields that need to be used for custom implementations (especially for base classes of components) to be made `protected internal`, so that they are still not visible for consumers, but can be used in inheriting classes.

Additionally, a lot of the properties of interfaces like `IForm` would need to be made public. For example, when I made a completely custom input component deriving from `AntInputComponentBase`, I had to create a work-around to access the `IForm.ValidateOnChange` property.

### Additional context
As a concrete example, I tried to create a custom version of the TreeSelect component that inherits from `SelectBase<TItem, TItem>` instead of `SelectBase<string, TItem>`, since the `Value` I needed to supply to it was already a `TItem`. I started by copying the code from `TreeSelect.razor`. However, that already didn't work because a lot of the properties and fields of `SelectBase` are internal.
(Like `SelectOptionItems` and `SelectMode`)

PR #3239, opened 2023-04-25
fix(module: form): Make AntInputComponentBase.OnValidated public

rhodon-jargon:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Partially related to #3235 

### 💡 Background and solution

In 95c4f13 (v0.14.0), The ValidationMessages setter became private. I was using that for an input control wrapper (that transformed the Value parameter) around a built-in input component. I needed to use the FieldIdentifier of the outer component for retrieving the ValidationMessages, but I wanted the original component to display the ValidationMessages. 

### 📝 Changelog

`AntInputComponentBase.OnValidated` is now public.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Allow to manually set ValidationMessages again using `AntInputComponentBase.OnValidated`  |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/input/Input.cs
+++ b/components/input/Input.cs
@@ -366,7 +366,7 @@ namespace AntDesign
             }
         }
 
-        internal override void OnValidated(string[] validationMessages)
+        public override void OnValidated(string[] validationMessages)
         {
             base.OnValidated(validationMessages);
             SetClasses();


ElderJames:
Thank you @rhodon-jargon , I still have no idea why `OnValidated` needs to be called from other components in your case. 
I see that this method is not currently used by all components, only Input is used to update styles.

Do you want to update the style of the input? We can find another way to deal with it.

Issue #3238, opened 2023-04-25
Tree component DefaultCheckedKeys change, the default options do not change


Jiang2020426:
### Describe the bug

Tree组件的DefaultCheckedKeys，当DefaultCheckedKeys的值切换时，默认勾选项不会变化；
bind-CheckedKeys属性并不是双向绑定，建议bind-CheckedKeys属性做成双向绑定，这样用户只需要动态变化bind-CheckedKeys的值，就可以达到属性DefaultCheckedKeys的效果

### Steps to reproduce (please include code)
<Tree TItem="SysMenu"
      Checkable
      DataSource="@menus"
      DefaultCheckedKeys="@menuIds"
      DefaultExpandAll
      TitleExpression="x => x.DataItem.Name"
      ChildrenExpression="x => x.DataItem.Child"
      KeyExpression="x=>x.DataItem.Id.ToString()">
</Tree>
@code {
    private string[] menuIds = new string[] { "4" };
    private List<SysMenu> menus = new List<SysMenu>
    {
    new SysMenu{ Id=1,Name="系统维护",Child=new List<SysMenu>
    {
     new SysMenu{Id=4,Name="订单管理"}  , 
     new SysMenu{Id=6,Name="订单管理2"}  ,
    } },
    new SysMenu{ Id=2,Name="系统维护1"},
    new SysMenu{ Id=3,Name="系统维护2"},

    };

    private void click()
    {
        menuIds = new string[] { "2","3" };
        StateHasChanged();
    }
    public class SysMenu
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public List<SysMenu> Child { get; set; }
    }
}

### Further technical details
- AntDesign Nuget Package version：0.14.4
![1](https://user-images.githubusercontent.com/68992508/234225127-ccd9f05d-5af7-4b64-8967-526038916f09.png)

- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version

Jiang2020426:
<Tree TItem="SysMenu"
      Checkable
      DataSource="@menus"
      DefaultCheckedKeys="@menuIds"
      DefaultExpandAll
      TitleExpression="x => x.DataItem.Name"
      ChildrenExpression="x => x.DataItem.Child"
      KeyExpression="x=>x.DataItem.Id.ToString()">
</Tree>

AndrewKaninchen:
[...]

PR #3243, opened 2023-04-26
feat(module: tree): Added public method GetNode(string key)

AndrewKaninchen:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

There is currently no way to set the selection of a Tree component from outside. #3238

### 💡 Background and solution

As the code currently stands, it's difficult to implement the ideal solution, which would be two-way binding of the SelectedNode, SelectedNodes SelectedKey and SelectedKeys properties, or DefaultCheckedKeys as described in #3238. 

As such, I've come up with this simple solution of exposing a method to get TreeNode objects directly from the Tree through their keys. Since the TreeNode class has public method SetSelection and the Tree class has DeselectAll, this covers all use-cases.

This is not mutually exclusive with a better solution, as the GetNode method is still potentially useful for other purposes (it does, for instance, have information on whether a node is collapsed or not, which the DataSource nodes wouldn't necessarily have).

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Tree: Added public method GetNode(string key).|
| 🇨🇳 Chinese |           |

- [x ] Changelog is provided or not needed

### DIFF
--- a/components/tree/Tree.razor.cs
+++ b/components/tree/Tree.razor.cs
@@ -652,6 +652,12 @@ namespace AntDesign
             }
             return base.OnFirstAfterRenderAsync();
         }
+        
+        /// <summary>
+        /// Get TreeNode from Key
+        /// </summary>
+        /// <param name="key">Key</param>
+        public TreeNode<TItem> GetNode(string key) => _allNodes.FirstOrDefault(x=>x.Key == key);
 
         /// <summary>
         /// Find Node


AndrewKaninchen:
@dotnet-policy-service agree

ElderJames (approved):

Issue #2243, opened 2022-01-10
Custom condition on table's filters 


DrTekno:
### Problem
![image](https://user-images.githubusercontent.com/33718346/148773101-21f756a1-15e4-4916-8fad-bb8a8c07c887.png)
Sometimes user need only a few conditions.

### Solution
Let the possibility to select wich filter condition available in the dropdown

anranruye:
tracked by #2057

Issue #2949, opened 2022-12-15
Table PropertyColumn Filterable with TypeConverter throws NotImplementedException


michaelcsikos:
I have a simple class which wraps a string to sort them more naturally, e.g.
`File1.txt`
`File2.txt`
`File10.txt`
(The default string sort would display `File10.txt` before `File2.txt`)

I have implemented a `TypeConverter` to convert to and from `string`.

Adding `Filterable` to the `PropertyColumn` throws the following exception when I try to _apply_ a filter:
```
System.NotImplementedException: The method or operation is not implemented.
   at AntDesign.FilterExpression.FilterExpressionResolver`1.GetFilterExpression()
   at AntDesign.TableModels.FilterModel`1.FilterList[TItem](IQueryable`1 source)
   at AntDesign.TableModels.QueryModel`1.ExecuteQuery(IQueryable`1 query)
   at AntDesign.Table`1.InternalReload()
   at AntDesign.Table`1.ReloadAndInvokeChange()
   at AntDesign.Table`1.AntDesign.ITable.ReloadAndInvokeChange()
   at AntDesign.Column`1.FilterConfirm(Boolean isReset)
   at AntDesign.Column`1.<get_FilterToolTipSorter>b__145_10()
   at Microsoft.AspNetCore.Components.EventCallbackWorkItem.InvokeAsync[T](MulticastDelegate delegate, T arg)
   at Microsoft.AspNetCore.Components.ComponentBase.Microsoft.AspNetCore.Components.IHandleEvent.HandleEventAsync(EventCallbackWorkItem callback, Object arg)
   at AntDesign.Button.HandleOnClick(MouseEventArgs args)
   at Microsoft.AspNetCore.Components.ComponentBase.CallStateHasChangedOnAsyncCompletion(Task task)
   at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task taskToHandle, ComponentState owningComponentState)
Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost: Error: Unhandled exception in circuit 'Nrudtywd1lOhLn59zOqn10MwbnuPkNKXPxh3cwZ_W-0'.

System.NotImplementedException: The method or operation is not implemented.
   at AntDesign.FilterExpression.FilterExpressionResolver`1.GetFilterExpression()
   at AntDesign.TableModels.FilterModel`1.FilterList[TItem](IQueryable`1 source)
[...]

Issue #2305, opened 2022-02-22
How to customize TableFilterCompareOperator


smallprogram:
When the user only needs certain filter conditions, it is not appropriate to use the built-in filter.
By default, the built-in filter contains a lot of filter conditions.
![image](https://user-images.githubusercontent.com/6050360/155180291-5e9fcca3-00f7-458f-b26e-22dd86237a80.png)
 I read the source code and found that this is controlled by the TableFilterCompareOperator enumeration class, but I don't know how to customize it, such as only showing "contain", "equal to" these two contents filter method.
Is there any way to customize the filter conditions?

Issue #2057, opened 2021-10-22
Table Column Filters - Set the default operator and the list of operators available


tale80:
Some ideas to improve filters on table columns :
- It will be good to be able to set the default filter operator (for a string, it's by default on "contains" operator and no option to set it to "Equal" by default).
- Another good addition will be to be able to define the list of operators availables by column.

PR #3279, opened 2023-05-31
feat(module: table): Allow custom Field filters

rhodon-jargon:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#2949, #2057, #2305 

### 💡 Background and solution

The previous `FilterExpression` code was very tightly coupled (because some of the code to support a type as filter was in `Column.razor`), and therefor didn't allow Field-type filters for custom types. I decoupled both the `FilterExpressions` and the type-specific `Column` code into implementations of `IFieldFilterType`. They are resolved by default from an `IFieldFilterTypeResolver`, but can be customized in three ways by the user.

With the `IFieldFilterType`, you can customize the default selected CompareOperator (as requested in #2057) and all available CompareOperators (requested in both #2057 and #2305). It also supports filtering additional types by customizing the generated expression and/or customizing input rendering.
(For the `DateTimeFieldFilterType`, I added an abstract `DateFieldFilterType` base class which allows developers (like me :) ) to more easily implement date filters for other date types like `DateOnly` or `DateTimeOffset`)

Also added support for default filters using the `Column`'s `DefaultFilters` property. This works with `TableFilterType.FieldType`, and these filters will be selected at the start but allow complete modification by the user.

### 📝 Changelog

Custom Field filter support in Table, currently only for the default CompareOperators. Select a custom `IFieldFilterType` using one of the following methods:
1. Globally, by changing the `IFilterTypeResolver` service registered in the service provider. (You may call `DefaultFieldFilterTypeResolver.Resolve` for unoverridden types to retain the default functionality)
2. In a table, by setting the `Table`'s `FieldFilterTypeResolver` property.
3. For a single column, using the `FieldFilterType` property.

Use `Column`'s `DefaultFilters` property to set the filters selected at startup.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Custom and default Field filter support in Table. See #3279 for more info. |
| 🇨🇳 Chinese |           |

I unfortunately don't have time to update the doc/demo or tests at the moment. If you feel it necessary, I hope you are alright with doing that. If something is unclear, feel free to ask.


ElderJames:
Very cool! Thank you @rhodon-jargon .

ElderJames:
Hello @rhodon-jargon , I'm going to merge this feature in next release. Could you please add a demo to show how this work?

ElderJames:
@rhodon-jargon Thank you ! I will add some additional info.

components/table/Filters/IFieldFilterType.cs
```diff
@@ -2,16 +2,16 @@
...
-namespace AntDesign.FilterExpression
+namespace AntDesign.Filters
 {
-    public interface IFilterExpression
+    public interface IFieldFilterType
     {
-        TableFilterCompareOperator GetDefaultCompareOperator();
+        TableFilterCompareOperator DefaultCompareOperator { get; }
+        RenderFragment<TableFilterInputRenderOptions> FilterInput { get; }
+        bool SupportsCompareOperator(TableFilterCompareOperator compareOperator);
```
ElderJames, line 14: I reviewed the IFieldFilterType code and found that the SupportsCompareOperator method only determines which operators to display, but not in which order. I think we can use a list instead, so that the user can easily specify the type and order. As long as we loop the render in the order of the list.

```
    StringFieldFilterType stringFieldFilterType=new();

    {
        stringFieldFilterType.SupportsCompareOperators.AddRange(new[]
        {
            TableFilterCompareOperator.LessThan,
        });

        base.OnInitialized();
    }

```

ElderJames:
also fixed #2243

components/table/TableLocale.cs
```diff
@@ -45,30 +45,23 @@ public class FilterOptions
...
-
-        public string GreaterThanOrEquals { get; set; } = "Greater Than Or Equals";
-
-        public string LessThanOrEquals { get; set; } = "Less Than Or Equals";
-
-        public string IsNull { get; set; } = "Is Null";
-
-        public string IsNotNull { get; set; } = "Is Not Null";
-
-        public string TheSameDateWith { get; set; } = "The Same Date With";
```
ElderJames, line 72: These are used for the localization, not just in English.

ElderJames (approved):

PR #3282, opened 2023-06-01
fix(module: table): Incorrect deselection in virtualized table

rhodon-jargon:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

If you selected a row in a virtualized table with the 'radio' selection type, scrolled down enough that the original row was unloaded, and selected a new row, the old row would not be deselected.

It happened because only the loaded rows were changed. That is now fixed by always deselecting all rowdata, even if unloaded, when selecting a row in 'radio' type.

Also cleaned up the selection code a bit (though it's still a bit cluttered), e.g. removing `_allRowDataCache` as it was not possible for it to contain multiple rowdatas per item, meaning it was the same as `_dataSourceCache`.

Added some extra overloads of `SetSelection`, I think it would be most ideal to remove the `Selection.Key` property (and thus the string overload of `Table.SetSelection`) in the future, maybe somehow giving it access to the TItem of the row so it can call the `SetSelection(TItem)` overload (in addition to some other things it would make easier).

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | A 'radio' `Selection` column will now correctly deselect all other items in a table with `EnableVirtualization`. |
| 🇨🇳 Chinese |           |


ElderJames:
Thank you for contribution @rhodon-jargon . I think #3284 is also related the selection, I wonder if we can repair it together?

ElderJames:
Hello @rhodon-jargon , I found the `select all` doesn't work now.
image

ElderJames:
Can't we merge rowdata with tabledataitem? I think they're almost one to one.

ElderJames:
Hello @rhodon-jargon , I'm reviewing this PR again and find another issue that when I select some rows, and remove then by changing SelectedRows outside, it work incorrectly and make the memory explode.
image
image

ElderJames:
Thank you @rhodon-jargon! I also find the issue that selection in virtualization mode would be clear while the table fetching additional data. I think we need to keep the state inside the table and retore after reload the same data.

table-selection

ElderJames (approved):
LGTM. Thanks!

Issue #3280, opened 2023-06-01
RadioGroup with blazor wasm breaks when value not in Radio Group Options


rabberbock:
### Describe the bug

When using a RadioGroup and setting the value to something not in the RadioGroup options (in our case clearing it to null), it results in an infinite loop. 
*Note*: This occurs only in blazor wasm and not when running using blazor server.

Same issue from https://github.com/ant-design-blazor/ant-design-blazor/issues/3121

### Steps to reproduce (please include code)

Example https://github.com/rabberbock/ant-design-blazor/commit/0015550d0e6b2dfa86ee3fdc8723c0e9314583f3
I ran the demo using `npm run start:wasm`
### Exceptions (if any)

![image](https://github.com/ant-design-blazor/ant-design-blazor/assets/3247622/44110cc4-8937-418d-a92f-fc5c0a6f8bc0)

### Further technical details
- AntDesign Nuget 0.15.0

ElderJames:
Fixed in #3287 .

PR #3286, opened 2023-06-04
Fix infinite loop in Radio Group for blazor wasm

rabberbock:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design-blazor/ant-design-blazor/issues/3280

### 💡 Background and solution
When a value was not one of the radio options, the `_selectedRadio` was not set to null to clear it out. Therefore, there was an infinite rendering issue. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/radio/RadioGroup.razor.cs
+++ b/components/radio/RadioGroup.razor.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


ElderJames:
Thank you @rabberbock , but I have another way to fixed this issue.
The main reason for this problem is that we process the selection logic in AfterRender and then render it again.
I'm going to remove this processing from it to reduce the render times and also simplify the judgment here.
Please help me reviewing the fixes https://github.com/ant-design-blazor/ant-design-blazor/pull/3287

rabberbock:
> Thank you @rabberbock , but I have another way to fix this issue. The main reason for this problem is that we process the selection logic in AfterRender and then render it again. I'm going to remove this processing from it to reduce the render times and also simplify the judgment here. Please help me reviewing the fixes #3287

Thanks! Will check it out.

rabberbock:
Closing in favor of #3287

Issue #3315, opened 2023-06-20
Drawer: Class ant-scrolling-effect Problem with overflow: hidden;


true-perfect-code:
If you place a button on the drawer that calls up any page (e.g. via **NavigationManager.NavigateTo("MyPage");** )
then the body of the calling HTML page loses the scroll bar. This happens by using the class **ant-scrolling-effect** in Drawer and property **overflow: hidden,** which is then no longer reset after closing Drawer.

The workaround is to reload the page, for example, with the call:
`NavigationManager.NavigateTo("MyPage", true); `

The problem is, that all session variables are lost by using this workaround (e.g. if you have defined them via **builder.Services.AddScoped** Service.

Regards
perfect-code

zxyao145:
@true-perfect-code hi, I'm sorry for this issue, but a temporary solution is as follows, call `JSInteropConstants.enableBodyScroll` yourself

```c#
[Inject]
private NavigationManager NavigationManager { get; set; }

protected override void OnInitialized()
{
	NavigationManager.LocationChanged += NavigationManager_LocationChanged;
}


private async void NavigationManager_LocationChanged(object sender, Microsoft.AspNetCore.Components.Routing.LocationChangedEventArgs e)
{
	Console.WriteLine("NavigationManager_LocationChanged");
	await JsInvokeAsync(JSInteropConstants.EnableBodyScroll);
}
```

true-perfect-code:
@zxyao145 Thanks for temporary solution.

true-perfect-code:
@ElderJames 
Has this update been released? I have installed the latest version 16.1, but I still have the problem.

I use Drawer as a component and the link calls CallBack function

Thanks
pc

ElderJames:
Maybe the NavigationManager still not be disposed in time.

true-perfect-code:
In the case that someone has a component on the drawer with the button that navigates to another page, for example:

Component
```
<Button OnClick="_OnClick_Page">Go to page</Button>

    private void _OnClick_Page()
    {
        OnClick_Page.InvokeAsync();
    }
```

Then it may be that the scroll bar "disappears" (as already described in the article).

**My workaround is the following:**
I eliminated component and moved it to the drawer. This eliminated the button an my problem with scrollbar. 
I still had links to other pages in the component, but these are solved via standard HTML links and seem to work without any problems.

PR #3316, opened 2023-06-21
fix(module: drawer): scroll not enable when page url changed

zxyao145:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#3315

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   fix(module: Drawer):  scroll not enable when page url changed   |
| 🇨🇳 Chinese |      修复（ Drawer 组件）：滚动条没有重置在页面 url 发生变化时候   |


ElderJames:
Could we EnableBodyScroll when the drawer is disposing?

zxyao145:
> Could we EnableBodyScroll when the drawer is disposing?

This is not friendly for service usage, we need to make modifications to the DrawerService too. Which solution do you prefer?

ElderJames:
@zxyao145 I prefer the disposing way.

zxyao145:
> @zxyao145 I prefer the disposing way.

get it

components/drawer/Drawer.razor.cs
```diff
@@ -517,6 +511,12 @@ private void CalcDrawerStyle()
         protected override void Dispose(bool disposing)
         {
             _timer?.Dispose();
+
+            if (_isOpen)
```
ElderJames, line 515: @zxyao145 I changed here. Because if there are more than one drawer on one page, each their Dispose() would be called, and EnableBodyScroll would be invoke more then one times.

However, one problem is that if several drawers are opened at the same page, you need to recognize that all drawers are currently closed before calling EnableBodyScroll.

ElderJames (approved):

PR #3391, opened 2023-08-15
feat(Module:Drawer) auto focus input on drawer open

eizzn:
when a drawer opens and it contains a form, the first visible input gains focus

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/JsInterop/modules/dom/manipulationHelper.ts
+++ b/components/core/JsInterop/modules/dom/manipulationHelper.ts
@@ -251,7 +251,7 @@ export class manipulationHelper {
               let form = forms[0];
               let inputs = form.querySelectorAll('input, textarea, select');
               try {
-                  [].forEach.call(inputs, function (input) {
+                  inputs.forEach(function (input:HTMLElement) {
                       var style = window.getComputedStyle(input);
                       if (style.display !== 'none' && style.visibility !== 'hidden') {
                           input.focus();
--- a/components/core/JsInterop/modules/dom/manipulationHelper.ts
+++ b/components/core/JsInterop/modules/dom/manipulationHelper.ts
@@ -242,6 +242,31 @@ export class manipulationHelper {
     styleHelper.removeCls(document.body, "ant-scrolling-effect");
   }
 
+  static focusOnChildFormIfExists(parentId) {
+      let BreakException = {};
+      let parent = document.getElementById(parentId);
+      if (parent) {
+          let forms = parent.getElementsByTagName('form');
+          if (forms.length > 0) {
+              let form = forms[0];
+              let inputs = form.querySelectorAll('input, textarea, select');
+              try {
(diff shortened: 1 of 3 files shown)


components/core/JsInterop/modules/dom/manipulationHelper.ts
```diff
@@ -242,6 +242,31 @@ export class manipulationHelper {
...
+          let forms = parent.getElementsByTagName('form');
+          if (forms.length > 0) {
+              let form = forms[0];
+              let inputs = form.querySelectorAll('input, textarea, select');
+              try {
+                  [].forEach.call(inputs, function (input) {
+                      var style = window.getComputedStyle(input);
+                      if (style.display !== 'none' && style.visibility !== 'hidden') {
+                          input.focus();
+                          throw BreakException;
```
ElderJames, line 258: Why not just use `foreach ... of` loop?

components/drawer/Drawer.razor.cs
```diff
@@ -441,6 +441,7 @@ protected override async Task OnAfterRenderAsync(bool isFirst)
                     }
             }
             await base.OnAfterRenderAsync(isFirst);
+            await JsInvokeAsync(JSInteropConstants.DomMainpulationHelper.FocusOnChildFormIfExists, Id);
```
ElderJames, line 444: Maybe this should be call every time the drawer was opened.

ElderJames (commented):
Thank you for the contribution @eizzn . Please add a new parameter to perform the feature because there are not all drawer are including form.

eizzn:
did not go with parameter because the js checks if there is a form before attempting focus.

components/core/JsInterop/modules/dom/manipulationHelper.ts
```diff
@@ -242,6 +242,31 @@ export class manipulationHelper {
...
+          let forms = parent.getElementsByTagName('form');
+          if (forms.length > 0) {
+              let form = forms[0];
+              let inputs = form.querySelectorAll('input, textarea, select');
+              try {
+                  inputs.forEach(function (input:HTMLElement) {
+                      var style = window.getComputedStyle(input);
+                      if (style.display !== 'none' && style.visibility !== 'hidden') {
+                          input.focus();
+                          throw BreakException;
```
ElderJames, line 258: I don't think throwing an exception is a good way to exit the loop. You can use `inputs.find(input=> ... )` to find the input

eizzn:
closing
use HTML attribute autofocus as alternative

PR #3393, opened 2023-08-18
Llp/antdesign blazor

llp1520:
This is a new feature that adds drag and scroll zoom when previewing an image, increasing the zoom level.


llp1520:
已经关了吧

Issue #3454, opened 2023-10-11
InputNumber doesn't have attribute for maxlength


chazikaifa:
The project I'm working on has a requirement that all the input components should have max length limits. However, `InputNumber` doesn't have attribute for `maxlength`. 

Although I can control the value using `Min` and `Max` , say from 0 to 10, users can still input very long string like '5.555555...' .  My temporary solution is using JS to add `maxlength` to the `<input>` or check in the `OnInput` EventCallback, which will cause unnecessarily overhead.

So, I believe it should be an enhancement if `InputNumber` can support `maxlength` attribute.

ElderJames:
Thanks for contacting us @chazikaifa , if you are interested, please feel free to submit a PR.

PR #3455, opened 2023-10-13
Add MaxLength parameter for InputNumber

chazikaifa:
Add MaxLength parameter for InputNumber.

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
[InputNumber doesn't have attribute for maxlength #3454](https://github.com/ant-design-blazor/ant-design-blazor/issues/3454)

### 💡 Background and solution

As described in the issue, the parameter Max and Min can NOT completely replace the function of maxlength for the component `<InputNumber>`. Parameter Maxlength is added and apply to the `<input>`. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |      Add MaxLength parameter for InputNumber     |
| 🇨🇳 Chinese |      数字输入框添加MaxLength属性     |

### DIFF
--- a/components/input-number/InputNumber.razor.cs
+++ b/components/input-number/InputNumber.razor.cs
@@ -60,6 +60,9 @@ namespace AntDesign
         public TValue Min { get; set; }
 
         [Parameter]
+        public int MaxLength { get; set; } = -1;
+
+        [Parameter]
         public bool Disabled { get; set; }
 
         [Parameter]
--- a/components/input-number/InputNumber.razor.cs
+++ b/components/input-number/InputNumber.razor.cs
@@ -60,7 +60,7 @@ namespace AntDesign
         public TValue Min { get; set; }
 
         [Parameter]
-        public int MaxLength { get; set; }
+        public int? MaxLength { get; set; }
 
         [Parameter]
         public bool Disabled { get; set; }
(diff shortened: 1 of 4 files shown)


components/input-number/InputNumber.razor.cs
```diff
@@ -59,6 +59,9 @@ public TValue DefaultValue
         [Parameter]
         public TValue Min { get; set; }
 
+        [Parameter]
+        public int MaxLength { get; set; } = -1;
```
ElderJames, line 63: Please set the default value to `null` so that the `maxlength` attribute would not be generated by default.

ElderJames:
Thanks for contribution @chazikaifa , plase updaet the API doc for the component.

components/input-number/InputNumber.razor.cs
chazikaifa: Thanks for your reply! Updated in the latest commit.

ElderJames: The property type also ned to be nullable `int?`

chazikaifa:
> Hello @chazikaifa ,please change the type to `int?` for the parameter.

Sorry for taking so long. I cannot push code in my working place and my private VPN is down these days.

ElderJames (approved):
LGTM

PR #3524, opened 2023-11-23
Update GroupedColumn_.razor

SuperQuestions:
Fix error example code for incorrect examples of grouping bar charts

### This is a Bug fix 

When I used the grouped column chart, I found that GroupedColumn no longer existed. The new example should be to use the IsGroup property of Column to display the grouped column chart.


SuperQuestions:
@dotnet-policy-service agree

ElderJames (approved):

ElderJames:
Thank you for contribution @SuperQuestions , could you please help us update the documentation of charts? It need to update to 0.4.0.

ElderJames (approved):

PR #3538, opened 2023-11-30
 fix(module: textarea) affix-wrapper-disabled doesn't append

zuevus:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

The TextArea component older then 14-th version of antdesign-blazor has the "affix-wrapper-disabled" class on span element of textarea wrap. It looked like this
image
But at the moment we don't have the class
image
We used it in our project for deactivate the textarea hover.
Solution is to add "affix-wrapper-disabled" class to wrapper span of textarea, as it has already done for inputs(base class) and as it was adding before in yonger then 14-th versions.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    SetClass method of textarea component adds {PrefixCls}-affix-wrapper-disabled to _warpperClassMapper |
| 🇨🇳 Chinese |           |

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -15,6 +15,22 @@ timeline: true
 
 ---
 
+### 0.16.3
+
+`2023-12-04`
+
+- Table
+  - 🛠 Refactor some internal components to render fragments, reducing allocation and avoid side effects cause by life cycle. [#3545](https://github.com/ant-design/ant-design/pull/3545) [@ElderJames](https://github.com/ElderJames)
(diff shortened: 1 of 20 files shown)


zuevus:
> @dotnet-policy-service agree

@dotnet-policy-service agree

ElderJames:
Thank you @zuevus , would you please add a unit test to cover this?

zuevus:
> Thank you @zuevus , would you please add a unit test to cover this?

@ElderJames thanks for comments. I guess I've done everything as you said and debug it. I've checked a test locally for it and pass it. Should I do something more for the issue?

Issue #3453, opened 2023-10-06
.NET 8 using Form of AntDesign with error FormName parameter if using <EditForm>


nguyendkn:
The POST request does not specify which form is being submitted. To fix this, ensure <form> elements have a @formname attribute with any unique value, or pass a FormName parameter if using <EditForm>.
<img width="1065" alt="Screenshot 2023-10-07 at 00 36 13" src="https://github.com/ant-design-blazor/ant-design-blazor/assets/100742244/cdf9a4f6-e29d-4be9-8192-5e4091091e42">
<img width="840" alt="Screenshot 2023-10-07 at 00 36 32" src="https://github.com/ant-design-blazor/ant-design-blazor/assets/100742244/50f81b50-0d80-4fb5-b843-ffffb8f13e33">
<img width="503" alt="Screenshot 2023-10-07 at 00 36 50" src="https://github.com/ant-design-blazor/ant-design-blazor/assets/100742244/6bb053d7-1ac8-4606-81f5-68f89585d80f">

ElderJames:
Thanks for contacting us @nguyendkn . We need to add the `Name` property for the components. Would you please help us sending a PR for that?

stfei:
我也一样的问题，而且当关闭预呈现时就不会报错
<Routes @rendermode="new InteractiveServerRenderMode(prerender: false)" />

CrosRoad95:
I have simillar issue. tried setting render mode as @stfei proposed, didn't work, can not add method="post" because:
```
[2024-01-03T07:30:42.129Z] Error: System.InvalidOperationException: Object of type 'AntDesign.Form`1[[ProjectName.LoginFormModel, ProjectName, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]' does not have a property matching the name 'method'.
   at Microsoft.AspNetCore.Components.Reflection.ComponentProperties.ThrowForUnknownIncomingParameterName(Type targetType, String parameterName)
   at Microsoft.AspNetCore.Components.Reflection.ComponentProperties.SetProperties(ParameterView& parameters, Object target)
   at Microsoft.AspNetCore.Components.ComponentBase.SetParametersAsync(ParameterView parameters)
   at Microsoft.AspNetCore.Components.Rendering.ComponentState.SupplyCombinedParameters(ParameterView directAndCascadingParameters)
```

full example:
```cs
<Form Model="@Model" Name="login" method="post" // method attribute throws error
[...]

PR #3608, opened 2024-01-03
Add Method and FormName parameter for Form

CrosRoad95:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [x] Other - Blazor .net8 compatbility

### 🔗 Related issue link

https://github.com/ant-design-blazor/ant-design-blazor/issues/3453

### 💡 Background and solution

You can not submit forms using post request what means you can not login using some methods such a cookie

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Adds `Method` parameter for `Form` component |
| 🇨🇳 Chinese |     |

### DIFF
--- a/components/form/Form.razor
+++ b/components/form/Form.razor
@@ -8,6 +8,7 @@
     <EditForm class="@ClassMapper.Class"
               style="@Style"
               id="@Id"
+              method="@Method"
               EditContext="_editContext"
               OnValidSubmit="OnValidSubmit"
               OnInvalidSubmit="OnInvalidSubmit">
--- a/components/form/Form.razor
+++ b/components/form/Form.razor
@@ -9,6 +9,7 @@
               style="@Style"
               id="@Id"
               method="@Method"
+              FormName="@FormName"
               EditContext="_editContext"
               OnValidSubmit="OnValidSubmit"
               OnInvalidSubmit="OnInvalidSubmit">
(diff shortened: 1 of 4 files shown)


CrosRoad95:
@dotnet-policy-service agree

ElderJames:
Thanks for contribution @CrosRoad95 , please update the documentation in https://github.com/ant-design-blazor/ant-design-blazor/blob/master/site/AntDesign.Docs/Demos/Components/Form/doc/index.en-US.md
Add is better to add a demo for that.

CrosRoad95:
Added documentation, adding example may be hard for me as i have issues with properly running the project

ElderJames (approved):

ElderJames:
@CrosRoad95 I noticed that Form originally has a Name attribute. Can I use it instead of FormName?

CrosRoad95:
I left "Name" property as is because of backwards compatbility and potentially it has different use case

ElderJames:
It was used for processing multiple forms in interactive rendering mode with [FormProvider](https://antblazor.com/en-US/components/form#components-form-demo-form-context). I think they are the same.

Issue #3410, opened 2023-08-31
Select - Virtualization of 2000 items


PrincessLunaOfficial:
Virtualization of huge amount of items in Select element does not work correctly.
You may test it in your own example here
https://antblazor.com/en-US/components/select

If you enter "22" in search box under "2000 Items example", you will see it bugged out, search results will start changing every second.

![Screenshot 2023-08-31 101055](https://github.com/ant-design-blazor/ant-design-blazor/assets/88183054/9a1cb12b-e2a1-4c07-9aee-4e31e36d9f58)

I see the same behaviour in my own project. Tested in different browsers (Chrome/Edge/Fire Fox)

ElderJames:
track in #3460

RVDH-work:
> track in [#3460](https://github.com/ant-design-blazor/ant-design-blazor/issues/3460)

Unfortunately this got broken again somewhere down the line. The issue can be reproduced on the example page mentioned. So #3460 needs to be reopened probably.

Magehernan:
the problem is different but is like this, we where talking about it on here #4771

Issue #3460, opened 2023-10-20
Select Virtualization Bug


HuaFangYun:
![70469c1903a8051fbb5142ebb1e5eb7](https://github.com/ant-design-blazor/ant-design-blazor/assets/30250585/52514ba7-1dba-44d5-ba68-2fa9f1010ee0)
![3bc01a60dc0e9339335388f8847be0c](https://github.com/ant-design-blazor/ant-design-blazor/assets/30250585/cdc26967-893b-4912-8e82-cc5ae656a5fe)
When I add the drop-down list to allow virtualization, the search appears in blank space, and when the scroll bar is dragged to the bottom, the content is automatically displayed. When virtualization is removed, the search is correct.

HuaFangYun:
![8f762dbee9347fc3c2dee46ddf73256](https://github.com/ant-design-blazor/ant-design-blazor/assets/30250585/99699eb6-d0cd-42b6-82c1-6d26b944e2c8)
When searching for content, select half of the text to display

Magehernan:
I see the an issue with this, in my case when you have a select with EnableVirtualization and EnableSearch the search dosn't work.

This is the code that use virtualization or not

https://github.com/ant-design-blazor/ant-design-blazor/blob/5d02b0b53e4aefae328812a589a8044f2d8048bc/components/select/Select.razor#L137-L155

In here it receive the collection to display but receive always all the items.
When you search it not filters the items, it mark the item IsHidden to true that makes the SelectOption to set a css to display:none.

So de problem is that the virtualize don't know of this internal beheaviour. It want the items to be displayed and it will base on the amount, show what is needed.

I try doing this in the template and it seems to work.

```csharp
var data = ctx.selectOptionItems.Where(t=>!t.IsHidden && !t.IsSelected).ToList();
```

in wasm work fine using 1300 items... but i think is needed to make a more deep change using cache for not recaculate this every time it draws.

I will submit a pull request with some changes

ElderJames:
Thank you @Magehernan , looking forward your PR.

Magehernan:
hi @ElderJames  i send the pull request trying to change the minimum code to make it work kepping an eye on performace but select has a lot of options so maybe i miss some case.

Issue #3021, opened 2023-01-17
Select component has wrong order when changing DataSource


tht245806497:
### Describe the bug

When changing the DataSource of a Select, the display order of options is inconsistent with the data source.

### Steps to reproduce (please include code)

Please see the demo code. 

### Screenshots

The first display is ok:
![image](https://user-images.githubusercontent.com/20590093/212896974-ed11e618-341c-4bb4-a521-3f0613ae00bb.png)

Switch data source, the first two options are in wrong order：
![image](https://user-images.githubusercontent.com/20590093/212897179-7e8911a6-21cc-46df-ae03-71b175a988db.png)

Switch the data source again, the options are displayed in reverse：
![image](https://user-images.githubusercontent.com/20590093/212897406-1e9155e9-7137-4bd3-a663-b39789c56be1.png)
![image](https://user-images.githubusercontent.com/20590093/212899959-4883ac12-8568-4de9-99b6-ea707c08c02f.png)

### Further technical details
- AntDesign Nuget Package version: 0.13.3
- TestDemo: 
[BlazorApp1.zip](https://github.com/ant-design-blazor/ant-design-blazor/files/10434915/BlazorApp1.zip)

ElderJames:
Thanks for contacting us @tht245806497 . We already know about this issue, so please set DataSource to null before switching for workaround.

```
@page "/"

<PageTitle>Index</PageTitle>

<RadioGroup TValue="bool" @bind-Value=isPersons1 ButtonStyle="@RadioButtonStyle.Solid" OnChange="OnChange">
    <Radio RadioButton Value=true>DataSource1</Radio>
    <Radio RadioButton Value=false>DataSource2</Radio>
</RadioGroup>

<Select TItem=Person
        TItemValue=int
        DataSource=dataSource
        LabelName=@nameof(Person.Label)
        ValueName=@nameof(Person.Id)
        Placeholder="Select a person"
        Style="width: 400px">
</Select>

@code
{
    record Person
    {
        public int Id { get; set; }
        public string Name { get; set; } = null!;
        public bool NotAvailable { get; set; }
        public string Label => $"{Id}-{Name}";
    }

    List<Person> _persons1 = new List<Person> {
        new Person {Id = 1, Name = "Jack"},
        new Person {Id = 2, Name = "Lucy"},
        new Person {Id = 3, Name = "Yaoming"},
        new Person {Id = 4, Name = "Frieda"},
[...]

PR #3625, opened 2024-01-11
(module: select) Fix scroll in select not working correctly when EnableVirtualization is true.

Magehernan:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#3460 Select Virtualization Bug
#3410 Select - Virtualization of 2000 items

### 💡 Background and solution

When Select filters options it not remove the option from the DOM it only hide it and leave it on the collection SelectOptionItems.

When passing all items include the hidden the virtualization component can't calculate the height of the scroll.

I create a member _filterOptionsVirtualization to store only the items that aren't hidden and refactor the template to use it, This way is a better more performant than recreating every time is renders.

I create another method to fill list with all the options when it start and when the search is clear to maintain the list that is needed because the virtualization component don't support IEnumerable and it was using .ToList every time it was rendered.

Please any suggestion or change to this solution is welcome... i only test it with the default type.

### 📝 Changelog

Fix search when virtualization is enabled.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor
+++ b/components/select/Select.razor
@@ -66,7 +66,7 @@
                                         @{
                                             @if (!IsGroupingEnabled)
                                             {
-                                                @selectOptionsRender((this, SortedSelectOptionItems, ItemTemplate))                                                                                               
+                                                @selectOptionsRender((this, ItemTemplate))                                                                                               
                                             }
                                             else
                                             {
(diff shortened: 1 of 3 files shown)


ElderJames:
Thanks for contribution @Magehernan , did this also fixed #3410 and #3021?

Magehernan:
@ElderJames i look at #3021 but it is not related to virtualization so i don't think this pull request solve that.
I run some test for #3410 with search with tags.

imagen

it looks like it is working fine. But only test with the demo example that is already in component page.

Magehernan:
i try the actual documentation web so it work bad in there buy i think i need to fix when you click on a option... it not reseting the search so let me fix that first.

Magehernan:
@ElderJames Hi, i review my changes and see that for the default mode is works but when you add multiple with it differents settings like not closing the overlay when you select a item and so other things.
I think in this case the optimization by caching the list of virtual elements is more or less innecesary and to make it work i think is necesary to change a lot of things.

I added another select on the demo example but with default type.

So the i think this is good to merge is a simple change that only apply a Where on the options and remove some innecesary checks on ICollection that accept IEnumerable.

ElderJames (approved):

PR #3636, opened 2024-01-16
add input onchanged demo

JIIIIL:


JIIIIL:
@dotnet-policy-service agree

ElderJames:
Thanks for contribution @JIIIIL , would you please add the `TValue` parameter on the API section of the Table documentation?

PR #3656, opened 2024-01-31
feat(module: select) Add FilterExpression on select for customize how to filter when searching

Magehernan:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

In spanish and other language you have characters like áéíóúñ and more, that for searching hide if you use the simple character.
By default it continue working like before but now you can pass a function to personalize how to compare the values and have more liberty to customize it by your requirements.

### 📝 Changelog

Added parameter FilterExpression for customize how to filter the labels in a select, is optional and by default work like before, no need to change anything.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -241,7 +241,7 @@ namespace AntDesign
         /// </summary>
         [Parameter] public Action<string> OnSearch { get; set; }
 
-        [Parameter] public Func<string, string, CultureInfo, bool> FilterExpression { get; set; } = (label, searchValue, cultureInfo) => label.Contains(searchValue, StringComparison.InvariantCultureIgnoreCase);
+        [Parameter] public Func<string, string, bool> FilterExpression { get; set; } = (label, searchValue) => label.Contains(searchValue, StringComparison.InvariantCultureIgnoreCase);
 
         [Parameter] public string PopupContainerMaxHeight { get; set; } = "256px";
 
@@ -1307,7 +1307,7 @@ namespace AntDesign
                 bool firstDone = false;
                 foreach (var item in SelectOptionItems)
                 {
-                    if (FilterExpression(item.Label, searchValue, CultureInfo))
+                    if (FilterExpression(item.Label, searchValue))
                     {
                         if (!firstDone)
                         {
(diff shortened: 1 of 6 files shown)


ElderJames:
Thanks for contribution @Magehernan , I think we no need to pass the ClutureInfo to the expression, because we can get this outside with `CulutreInfo.CurrentClutureInfo`.

Magehernan:
@ElderJames Ok, i was passing that because in the case of using the comparator for ignore non spaces characters in needed but like you say i can change that and use it externally.

I will make the change. 
Thanks

Magehernan:
I change it, let me know if you see anything else. Thanks!

ElderJames:
Thank you. I have another idea that we can pass the SelectOptionItem to the expression that use can use the data to filter.

Magehernan:
You are totally right, i sent the changes.

ElderJames:
Awesome! Would you please add a test for this feature?

Magehernan:
i add 2 test, one with the default expression and one with the custom expression i am using.
Is my first time making a test for blazor so i copy another to reuse... i think is working fine.
I will go to sleep i am on the other side of the world, have a good day!

ElderJames:
Thank you @Magehernan ! I would merge this PR. Good night.

ElderJames (approved):
LGTM.

Magehernan:
I don't know if i do something wrong, but i see a lot of changes that i do not make in this pull request, and i don't see the merge.
Let me know if i can help to solve this.

ElderJames:
Hi @Magehernan , we need merge this new feature to the feature branch. So could you please rebase the commit to the feature branch?

Magehernan:
Hi, i am not sure but i do the rebase on the branch feature, hope this is what was needed, let me know, thanks!

ElderJames:
Yes, thank you.

ElderJames (approved):

PR #3659, opened 2024-02-01
Fix label in SegmentedOption

CrosRoad95:
It self explain, but: new `SegmentedOption<int>(1, "one")` will render "1" instead of "one" which is a label i set


ElderJames (approved):
LGTM

PR #3668, opened 2024-02-05
fix default scroll style

thirking:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Table scroll style is not consistent with the browser default style.
3e440311cd35ff6b50e292b45cff799c

### 💡 Background and solution

In order to solve the issue of misplacement of table fixed columns, a property called `ScrollBarWidth `was added. 
By reading the source codes of `antd` and `ant design vue`, I found that we can get the scrollbar size in the browser by calling js method.

### 📝 Changelog

I kept the ScrollBarWidth attribute and made it null by default, in which case the scroll bar style is consistent with the browser default style.
image

and we can still set this property manually.
image

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   fix default ScrollBar style in Table   |
| 🇨🇳 Chinese |  修复Table中默认滚动条样式  |

### DIFF
--- a/components/table/Table.razor.cs
+++ b/components/table/Table.razor.cs
@@ -212,8 +212,8 @@ namespace AntDesign
         private bool _waitingDataSourceReload;
         private bool _waitingReloadAndInvokeChange;
         private bool _treeMode;
-        private string _scrollBarWidth = "17px";
-
+        private string _scrollBarWidth;
+        private string _realScrollBarSize = "15px";
         private bool _hasFixLeft;
         private bool _hasFixRight;
         private int _treeExpandIconColumnIndex;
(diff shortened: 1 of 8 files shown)


components/table/Table.razor.cs
```diff
@@ -635,6 +636,23 @@ protected override void OnInitialized()
             FieldFilterTypeResolver ??= InjectedFieldFilterTypeResolver;
         }
 
+        public override async Task SetParametersAsync(ParameterView parameters)
+        {
+            if (parameters.TryGetValue<string>(nameof(ScrollBarWidth), out var value))
+            {
+                if (value is null)
+                {
+                    var scrollBarSize = await JsInvokeAsync<double>(JSInteropConstants.DomMainpulationHelper.GetScrollBarSize, false);
```
ElderJames, line 645: Please get the scrollbar size once and cache the value globally.

thirking: added `ClientDimensionService.cs`, now we can get the scrollbar size once and cache the value globally in C#.

ElderJames (approved):

PR #3671, opened 2024-02-06
Update ReuseTabsService.cs  add  AddReuseTabsPageItem method

jxcproject:
issues: 3660

### 🤔 This is a ...

- [x ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tabs/Reuse/ReuseTabsService.cs
+++ b/components/tabs/Reuse/ReuseTabsService.cs
@@ -181,7 +181,7 @@ namespace AntDesign
         /// </summary>
         /// <param name="pageUrl">The url of target page</param>
         /// <param name="title">The title show on the tab</param>
-        public void CreateTab(string pageUrl, RenderFragment? title)
+        public void CreateTab(string pageUrl, RenderFragment? title = null)
         {
             if (_pageMap.ContainsKey(pageUrl))
             {
--- a/components/tabs/Reuse/ReuseTabsService.cs
+++ b/components/tabs/Reuse/ReuseTabsService.cs
@@ -178,40 +178,46 @@ namespace AntDesign
 
 
           
-        public void AddReuseTabsPageItem(params string[] keys)
+     
+        public void AddReuseTabsPageItem(string pageUrl, string? routeUrl, RenderFragment? title)
         {
-            var list = GetAllAssembly();
-            foreach (var key in keys)
+            /*
+              Use it like this
+                TabService.AddReuseTabsPageItem("/module/1000/1001", "/module/{group}/{menu}", (builder) => { builder.AddContent(0, "Title"); });
+                TabService.AddReuseTabsPageItem("/module/1000/1002", "/module/{group}/{menu}", null);
+                TabService.AddReuseTabsPageItem("/user/setting", null, null);
+             */
+        
+            if (!_pageMap.ContainsKey(pageUrl))
             {
-                if (!_pageMap.ContainsKey(key))
+                var list = GetAllAssembly();
+                foreach (var item in list)
                 {
-                    foreach (var item in list)
+                    var pageType = item.ExportedTypes
+                    .Where(w =>
                     {
-                        var pageType = item.ExportedTypes
-                        .Where(w =>
+                        var routeAttribute = w.GetCustomAttribute<RouteAttribute>();
+                        var reuseTabsAttribute = w.GetCustomAttribute<ReuseTabsPageAttribute>();
+                        var url = reuseTabsAttribute?.PinUrl ?? routeAttribute?.Template;
+                        return url == (string.IsNullOrWhiteSpace(routeUrl) ? pageUrl : routeUrl);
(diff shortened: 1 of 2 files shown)


components/tabs/Reuse/ReuseTabsService.cs
```diff
@@ -176,6 +176,50 @@ private void AddReuseTabsPageItem(Type pageType)
...
+        {
+            /*
+              Use it like this
+                TabService.AddReuseTabsPageItem("/module/1000/1001", "/module/{group}/{menu}", (builder) => { builder.AddContent(0, "Title"); });
+                TabService.AddReuseTabsPageItem("/module/1000/1002", "/module/{group}/{menu}", null);
+                TabService.AddReuseTabsPageItem("/user/setting", null, null);
+             */
+        
+            if (!_pageMap.ContainsKey(pageUrl))
+            {
```
ElderJames, line 192: Please return immediately while the _pageMap cantains the pageUrl.

components/tabs/Reuse/ReuseTabsService.cs
```diff
@@ -176,6 +176,50 @@ private void AddReuseTabsPageItem(Type pageType)
             _pageMap[url] = reuseTabsPageItem;
         }
 
+
+          
+     
+        public void AddReuseTabsPageItem(string pageUrl, string? routeUrl, RenderFragment? title)
```
ElderJames, line 182: Can we merge these two urls into one? I don't think there are any scenarios that need to be used together.

ElderJames:
Thank you for contribution @jxcproject , please sign the CLA first and resolve the review comments.

components/tabs/Reuse/ReuseTabsService.cs
jxcproject: It is used when a page needs to input parameters and loads different components according to different parameters, that is, the page is public. I'm making a low code platform here, and basically all forms are drag-and-drop configurations, so they all share a page and load different form configurations according to different menus.
For example, in my example above, the route setting of my public page is /module/{group}/{menu}, if you search by the actual URL /module/1000/1001,

var routeAttribute = w.GetCustomAttribute<RouteAttribute>();
var reuseTabsAttribute = w.GetCustomAttribute<ReuseTabsPageAttribute>();
var url = reuseTabsAttribute?. PinUrl ?? routeAttribute?. Template;

In this case, the url is /module/{group}/{menu}, and the corresponding pageType cannot be found.

components/tabs/Reuse/ReuseTabsService.cs
jxcproject: The above is to judge that it does not exist before entering the loop code, why do you need to judge that there exists and return it directly? Are you trying to add else {return;} } ?

ElderJames: I just want to reduce cod block nesting

components/tabs/Reuse/ReuseTabsService.cs
ElderJames: I konw the route template. But I think this method just adds tabs rather than actually accessing the page. Until the user hits the tab, blazor finds the corresponding page type through PageUrl(with route parameters) and renders it.

jxcproject:
> Thank you for contribution @jxcproject , please sign the CLA first and resolve the review comments.

II've signed the CLA, see if it works?

ElderJames:
> > Thank you for contribution @jxcproject , please sign the CLA first and resolve the review comments.
> 
> II've signed the CLA, see if it works?

No, please reply @dotnet-policy-service agree

jxcproject:
@dotnet-policy-service agree

jxcproject:
@microsoft-github-policy-service agree

components/tabs/Reuse/ReuseTabsService.cs
ElderJames: > But the question is, how to find the corresponding pageType according to /module/1000/1001? I don't have a good plan.

In theory, when you click the tab, it would navigate to the target url and RouteView would find the PageType and route parameters.

ElderJames: Then it should call the ReuseTabsService.TrySetRouteData method and build the ReuseTabsPageItem.
https://github.com/ant-design-blazor/ant-design-blazor/pull/3671/files/e2b47e83ccb4cd63bb6f91978fae7b4e1a9cf7c3#diff-902b04cffed819dd7d81d3579281070c11cef725b98e487d5add8f571010cf2cR56-R81

ElderJames (approved):

Issue #3651, opened 2024-01-26
Message Info report: Key values must be unique


Forfires:
I use it in Blazor Hybird, and when the program pops up messages several times, the program reports an error: 

> InvalidOperationException: More than one sibling of component 'AntDesign.MessageItem' has the same key value, 'AntDesign.MessageConfig'. Key values must be unique.

I called it as follows

`
 _ = _message.Info(notification.Message);
`

zxyao145:
@Forfires Hi, What is the type of notification.Message? Is it `MessageConfig`? If so, please ensure that different references are passed, such as using `_ = _message.Info(new MessageConfig(){ .... });`

Forfires:
I've tried calling it this way,but the problem persists
`   _ = _message.Info(new MessageConfig()
       {
           Content = "保存成功",
           Key = Guid.NewGuid().ToString()
       });`

Forfires:
It should be added that I triggered this method about 80 times in 10 seconds

zxyao145:
@Forfires Hi, it requires more context information, which I did not reproduce locally, or you can provide a demo. 

This is my demo, I click the button once per second, it works well.

```razor

@inject MessageService _messageService

<Button Type="primary" OnClick="@ShowMsg">
    ShowMsg
</Button>

@code {

    private void ShowMsg()
    {
        for (int i = 0; i < 10; i++)
        {
            _messageService.Info(
                new MessageConfig()
                {
                    Content = "保存成功",
                    Key = Guid.NewGuid().ToString()
                });
        }

        _messageService.Info(
            new MessageConfig()
            {
                Content = "END",
                Key = Guid.NewGuid().ToString(),
                Duration = 5
            });
    }
}
```

ElderJames:
Is it possible that multiple AntContainters are set up?

Forfires:
> @Forfires Hi, it requires more context information, which I did not reproduce locally, or you can provide a demo.
> 
> This is my demo, I click the button once per second, it works well.
> 
> ```
> @inject MessageService _messageService
> 
> <Button Type="primary" OnClick="@ShowMsg">
>     ShowMsg
> </Button>
> 
> @code {
> 
>     private void ShowMsg()
>     {
>         for (int i = 0; i < 10; i++)
>         {
>             _messageService.Info(
>                 new MessageConfig()
>                 {
>                     Content = "保存成功",
>                     Key = Guid.NewGuid().ToString()
>                 });
>         }
[...]

PR #3698, opened 2024-02-24
fix: message non thread safe on Webview

zxyao145:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#3651

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     fix: message non thread safe on Webview      |
| 🇨🇳 Chinese |    修复 message 在  Webview 上非线程安全的问题      |


components/message/Message.razor.cs
```diff
@@ -64,11 +66,10 @@ private MessageConfig Extend(MessageConfig config)
 
         #endregion global config
 
-        private readonly List<MessageConfig> _configs
-            = new List<MessageConfig>();
+        private readonly ConcurrentDictionary<string, MessageConfig> _configDict
```
ElderJames, line 69: I think it's better to use `ConcurrentQueue` because you don't have to handle the `Order`.

zxyao145: ConcurrentQueue cannot be satisfied:

1. It is necessary to follow the `MessageConfig` based on the `Message.Key`, this must be done using Dictionary like object.
2. It needs to be able to delete the specified MessageConfig, But the operation of ConcurrentQueue can only be done at the head or tail of the queue, and cannot be done from the middle.

ElderJames (approved):

Issue #3637, opened 2024-01-16
Can't get Select component to refresh


michaelcsikos:
Inside a `Drawer` I have several `Select` components which are bound to a `ViewModel` object. When the values are changed in the bound `ViewModel` the `Select` components are not refreshing to show the current value. If I output the value on the page as text, the values are correct. I have tried calling `StateHasChanged` but it's not helping. What can I do to force the `Select` components to refresh?

michaelcsikos:
To be clear, the items in the `DataSource` for the `Select` are not being changed in the code, it's just the `@bind-Value` which is not refreshing.

ElderJames:
Thanks for contacting us @michaelcsikos . It's it only have issue in Drawer? How about put the select components on the page?

michaelcsikos:
Hi @ElderJames, if I place a second instance of the same components on the page bound to the same `ViewModel` it refreshes correctly, so it appears to only a problem with the drawer.

If I select a different value in the drawer `Select`, the second instance on the page immediately refreshes to show the correct value. But editing the page's `Select` does not refresh in the drawer.

zxyao145:
@michaelcsikos Hi, could you provide a demo? I haven't reproduced it on my local (may be different from your usage)

https://github.com/ant-design-blazor/ant-design-blazor/assets/24973084/4ca56ffd-4345-4a86-95d1-0f1863d0ee09

edwardbarford:
@michaelcsikos Is this select that is not working contained in a form with ValidateOnChange = true?

If so the above pull request will fix. Otherwise this is a separate bug but the pull request is still valid.

My first pull request so go easy on me :).

PR #3703, opened 2024-02-26
Changes to Value bound to select are not reflected in selected option when contained in a Form with  ValidateOnChange = true

edwardbarford:
fix(module:select): run OnValueChange from OnParametersSet when form ValidateOnChange = true

When the parent Form has ValidateOnChange = true the OnValueChange method needs to be executed when OnParametersSet executes in order to set the newly selected option.

### 🤔 This is a ...

- [ ] New feature
- [ x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Found a bug where selects don't have their selected options updated when they are contained in a form with ValidateOnChange = true.

I suspect this issue is related: https://github.com/ant-design-blazor/ant-design-blazor/issues/3637

### 💡 Background and solution

Found a bug where selects don't have their selected options updated when they are contained in a form with ValidateOnChange = true.

The OnValueChange func only ran when ValidateOnChange = false. I've changed it to run irrespective.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fixed bug where selects contained in forms with ValidateOnChange = true don't appear to update when bound values change .  |
| 🇨🇳 Chinese |           |

- [ x] Doc is updated/provided or not needed
- [ x] Demo is updated/provided or not needed
- [ x] Changelog is provided or not needed

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -464,10 +464,8 @@ namespace AntDesign
                 {
                     EditContext?.NotifyFieldChanged(FieldIdentifier);
                 }
-                else
-                {
-                    OnValueChange(_selectedValue);
-                }
+
+                OnValueChange(_selectedValue);
             }
             base.OnParametersSet();
         }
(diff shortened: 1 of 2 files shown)


edwardbarford:
@dotnet-policy-service agree

Issue #3750, opened 2024-03-23
Order of 'Select' data changed after replacing data source


Velociraptor45:
### Describe the bug

When I change the data source of a Select component, the order of its items is altered. The funny part is, the initially set data source does not have this problem. It also looks like this bug has been here before and came back (see #2439).

### Steps to reproduce (please include code)
We have two Selects here. The first one changes the data source of the second one.
```
<Select
    DataSource="_types"
    TItem="BaseType"
    TItemValue="int"
    LabelName="@nameof(BaseType.Name)"
    ValueName="@nameof(BaseType.Id)"
    DefaultValue="1"
    OnSelectedItemChanged="OnChange">
</Select>

<Select
    DataSource="GetDataSource(_type)"
    TItem="SubType"
    TItemValue="Guid"
    LabelName="@nameof(SubType.Name)"
    ValueName="@nameof(SubType.Id)"
    IgnoreItemChanges="false">
</Select>

@code {
    private int _type = 1;

    private readonly List<BaseType> _types = new()
    {
        new BaseType(1, "Type 1"),
        new BaseType(2, "Type 2")
    };

    private readonly List<SubType> _subTypeDs1 = new()
    {
        new(Guid.NewGuid(), "Item 1", 1),
        new(Guid.NewGuid(), "Item 2", 2),
        new(Guid.NewGuid(), "Item 3", 3),
        new(Guid.NewGuid(), "Item 4", 4),
        new(Guid.NewGuid(), "Item 5", 5),
        new(Guid.NewGuid(), "Item 6", 6),
        new(Guid.NewGuid(), "Item 7", 7),
        new(Guid.NewGuid(), "Item 8", 8),
        new(Guid.NewGuid(), "Item 9", 9)
    };

    private readonly List<SubType> _subTypeDs2 = new()
    {
        new(Guid.NewGuid(), "Item2 1", 1),
        new(Guid.NewGuid(), "Item2 2", 2),
        new(Guid.NewGuid(), "Item2 3", 3),
        new(Guid.NewGuid(), "Item2 4", 4),
        new(Guid.NewGuid(), "Item2 5", 5)
    };

    private void OnChange(BaseType newType)
    {
        _type = newType.Id;
        StateHasChanged();
    }

    private IEnumerable<SubType> GetDataSource(int type)
    {
        if (type == 1)
            return _subTypeDs1;

        return _subTypeDs2;
    }

[...]

PR #3806, opened 2024-04-16
Fix List order after datasource change

miguelkmarques:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#3750 and #3021

### 💡 Background and solution
In my project we have faced this issue mentioned above, so i have investigated the root cause.
Due to how HashSet works internally on .NET, when an item is removed from a HashSet (using the Remove method), the space allocation maybe still be there, so when a new item is added (using the Add method), it is not guaranteded to be inserted at the end of the HashSet, so it will not follow the order from the Datasource sent to the Select component, causing the bug.

The solution is to call the Clear method on the HashSet SelectOptionItems after storing all the existing options in the Dictionary, and add everything from the new Datasource, and keep the select state of existing items.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -657,7 +657,6 @@ namespace AntDesign
 
                         if (exists is null)
                         {
-                            SelectOptionItems.Remove(selectOption);
                             RemoveEqualityToNoValue(selectOption);
 
                             if (selectOption.IsSelected)
@@ -668,6 +667,7 @@ namespace AntDesign
                     }
                 }
             }
+            SelectOptionItems.Clear();
 
             //A simple approach to avoid unnecessary scanning through _selectedValues once
             //all of SelectOptionItem where already marked as selected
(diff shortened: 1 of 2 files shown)


miguelkmarques:
@dotnet-policy-service agree

ElderJames:
Thanks for contribution @miguelkmarques , could you please add tests for the changes?

miguelkmarques:
> Thanks for contribution @miguelkmarques , could you please add tests for the changes?

Sure, I will add some tests

miguelkmarques:
I figured out that the component Parameter IgnoreItemChanges must be always false for my solution to work when the datasource is changed. So i have two options.
1. Update my solution to only call the Clear method when IgnoreItemChanges equals false. So this issue will stil occur when IgnoreItemChanges equals true (which is the default value).
2. Remove the Parameter IgnoreItemChanges, but this will be a breaking change, since it's used for performance reasons.

I'll follow the first option to not submit any breaking changes.

ElderJames:
I prefer the first option too. Thank you for figureing out that.

miguelkmarques:
Hi @ElderJames, solution updated and tests added, please review.
Thank you!

ElderJames (approved):

PR #3863, opened 2024-05-14
Add HideColumnsByName parameter to handle GenerateColumns hide some column

dessli:
…olumn

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Many times, automatically generated columns need to hide certain class fields. By passing an array of field names through HideColumnByName, these fields can be easily hidden.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames:
Thanks for contribution @dessli , have you tested dynamic hiding/display columns? Because column itself has a Hidden parameter, you can use Definitions to specify which column is hidden.

dessli:
> Thanks for contribution @dessli , have you tested dynamic hiding/display columns? Because column itself has a Hidden parameter, you can use Definitions to specify which column is hidden.

Hello @ElderJames , yes,  have tried Definitions, something like this:
```
<GenerateColumns Range="1.." Definitions="HideColumn"/>

public void HideColumn(string columnName, object columnObject)
{
	var hideColumnList = new List<string> { "IdCard", "Phone" };
	if (!hideColumnList.Contains(columnName)) return;
	var type = columnObject.GetType();
	var baseType = type.BaseType;
	var property = baseType?.GetProperty("Hidden");
	if (property != null)
	{
		property.SetValue(columnObject, true);
	}
}
```

But I think having a parameter like HideColumnsByName would be more convenient.🤔

ElderJames (approved):

ElderJames:
All right, please add the new parameter to the API doc.

ElderJames (approved):

PR #3877, opened 2024-05-22
Add GenerateFormItem component for automatic generation basic FormItem

dessli:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Simplify the construction of basic forms or multi-field forms.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


components/form/GenerateFormItem.razor.cs
```diff
@@ -0,0 +1,215 @@
...
+    [CascadingParameter(Name = "Form")]
+    private IForm Form { get; set; }
+
+
+    /// <summary>
+    /// If true then the rules will be required by CheckTypeRequire.
+    /// Recommended only for forms without complex validation, where only simple validation for empty values is required.
+    /// </summary>
+    [Parameter]
+    public bool RulesRequireByType { get; set; }
```
ElderJames, line 31: I think we don't need this parameter and use `ValidateMode` instead to set the validation.

dessli: I agree with your perspective. We can remove this parameter and use ValidateMode instead to set the validation. I’ll make the necessary adjustments to the code, or have you already started making the changes?

ElderJames: Not yet, please.

components/form/GenerateFormItem.razor.cs
```diff
@@ -0,0 +1,215 @@
...
+        if (underlyingType == typeof(DateTime))
+        {
+            return MakeDatePicker(property, isRequired);
+        }
+
+        if (underlyingType.IsEnum)
+        {
+            return MakeEnumSelect(property);
+        }
+
```
ElderJames, line 118: I think we also need to consider nested objects.

components/form/GenerateFormItem.razor.cs
```diff
@@ -22,13 +22,8 @@ public partial class GenerateFormItem<TModel> : ComponentBase
...
     private IForm Form { get; set; }
 
-
-    /// <summary>
-    /// If true then the rules will be required by CheckTypeRequire.
-    /// Recommended only for forms without complex validation, where only simple validation for empty values is required.
-    /// </summary>
     [Parameter]
-    public bool RulesRequireByType { get; set; }
+    public Func<PropertyInfo, FormValidationRule[]?> ValidateRules { get; set; }
```
dessli, line 26: @ElderJames I have removed RulesRequireByType and added ValidateRules.

components/form/GenerateFormItem.razor.cs
dessli: I’ll think about how to handle this.

components/form/GenerateFormItem.razor.cs
```diff
@@ -52,23 +47,29 @@ protected override void BuildRenderTree(RenderTreeBuilder builder)
             }
 
             var displayName = property.GetCustomAttribute<DisplayNameAttribute>()?.DisplayName ?? property.Name;
```
ElderJames, line 49: We can use `PropertyReflector` to get the display name from localization.

components/form/GenerateFormItem.razor.cs
```diff
@@ -52,23 +47,29 @@ protected override void BuildRenderTree(RenderTreeBuilder builder)
...
-            childContent ??= GenerateByType(underlyingType, property, displayName, isRequired);
+            childContent ??= GenerateByType(underlyingType, property, displayName);
 
             // Make FormItem 
-            if (isRequired)
+            if (validateRule != null)
             {
-                var formValidation = GetFormValidationRuleByType(underlyingType);
                 builder.OpenComponent(0, typeof(FormItem));
                 builder.AddAttribute(1, "Label", displayName);
```
ElderJames, line 70: The Label would auto generated inside FormItem.

dessli:
@ElderJames  Hi, I’ve refactored the code to support the automatic generation of nested types. For the automatic generation of array classes, I’m considering using Dynamic Field in the future.

ElderJames:
That's very cool @dessli ! And yes, using Dynamic Fields can reduce some of the code, and users can customize the components they want.

ElderJames:
And I have an idea that we can use [Collapse](https://antblazor.com/en-US/components/collapse) to wrap the sub form generated by nested types. So that the style will be better.

dessli:
> And I have an idea that we can use [Collapse](https://antblazor.com/en-US/components/collapse) to wrap the sub form generated by nested types. So that the style will be better.

ok, I’m considering a feature for multi-style selection as well.

dessli:
@ElderJames Hi, I’ve added the multi-style feature, which currently supports Collapse and Block styles. The default style is set to Collapse.

ElderJames:
Looks like validation is not working in nest block.
image

dessli:
> Looks like validation is not working in nest block. image

It seems to be caused by the subclass reflection not being properly bound to the component. I’ll make corrections later. I’ve been really busy with work lately.

dessli:
@ElderJames Hi, The validation is not working problem, I’ve found the cause. It was an issue with the demo code where I forgot to add ValidateMode and ValidateComplexType.

ElderJames:
All right. But I think the required attribute is by default.

ElderJames (approved):

ElderJames:
Thank you for contribution @dessli .

Issue #3204, opened 2023-03-28
[BUG] `Layout` with `Menu` responsive will selected two `MenuItem`


GF-Huang:
### Describe the bug

- Expected: it will only allow to select single `MenuItem` in a time.

- Unexpected: But in collapsed state, it allow to selecting 2.

https://user-images.githubusercontent.com/4510984/228338696-459d38e2-66bb-4f29-bf7a-cedce09c754e.mp4

### Further technical details
- Browser: `Microsoft Edge 版本 111.0.1661.54 (正式版本) (64 位)`

PR #3894, opened 2024-05-31
fix(module: menu): MenuItem IsSelected visual state for non-default use cases

iits-timon-holzhaeuser:
### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
There is no issue for this bug

### 💡 Background and solution
I was wrapping my menu items into their own components for reuse. After this, the menu items selection was not updated properly.
I recorded a minimal set of examples (the no content case was discovered during testing for this fix):
menu_without_fix

The StateHasChanged, which is called on the hosting menu, not the MenuItem itself, does not seem to trigger the css classes to be refetched by Blazor. Issue #3204 maybe related to this.

After some testing, I decided to just call the StateHasChanged on the MenuItem when Select/Deselect is called. After all, these methods only get called when the active state was actually changed. Which fixed the issue:
menu_with_fix

Minimal example code:
image

### 📝 Changelog
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | MenuItem's `IsSelected` visual state is now correctly updated, when it's wrapped into a component or has no child content |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/menu/MenuItem.razor.cs
+++ b/components/menu/MenuItem.razor.cs
@@ -152,10 +152,8 @@ namespace AntDesign
             if (!skipParentSelection)
                 ParentMenu?.Select(isInitializing);
 
-            // fixed https://github.com/ant-design-blazor/ant-design-blazor/issues/3204
-            // It seems that the `StateHasChanged()` call in parent menu doesn't work correctly when the menuitem was warpped by a tooltip.
-            if (ShowTooltip)
-                StateHasChanged();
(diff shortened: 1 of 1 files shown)


iits-timon-holzhaeuser:
@dotnet-policy-service agree

ElderJames (approved):
Thank you!

PR #3949, opened 2024-07-01
add tree property includeindeterminate to tree component

MarvelTiter:
add new property named `IncludeIndeterminate` to Tree Component
this will make binding result will contains the indeterminate state node
such as bind-CheckedKeys, bind-SelectedKeys, bind-SelectedNodes etc..

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tree/Tree.razor.cs
+++ b/components/tree/Tree.razor.cs
@@ -23,6 +23,12 @@ namespace AntDesign
         [CascadingParameter(Name = "TreeSelect")]
         public ITreeSelect TreeSelect { get; set; }
 
+        /// <summary>
+        /// bind-[values] will include indeterminate state node
+        /// </summary>
+        [Parameter]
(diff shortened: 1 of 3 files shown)


pankey888:
"Indeterminate" is only related to "checked" and has no association with "selected.", so there's no need to modify the ”SelectedKeys“. And if you add the node's key to "CheckedKeys" whose checked state is indeterminate, the next time all the children will become checked, because we don't know it is indeterminate, unless you add more details.

PR #4072, opened 2024-08-07
Globally fix string number formatting incorrect in different cultures

jbatonnet:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

I didn't find any existing issue, but I'm working on a project and I have the issue locally.
Description below.

### 💡 Background and solution

In many locations in the code, string interpolation is used to format floating points numbers to string without specifying the culture info. It will use the current culture and we'll end up with numbers being formatted like "1234, 5678px" for example.
The browser will reject css in this format and we'll have a broken style.

In my specific scenario, I used an <Affix> element with my culture "English (Sweden)" / "en-SE".
While inspecting in my browser, I can see this style being applied:
image

The structure `CssSizeLength` exists and already formats the number correctly using InvariantCulture.
This uses this structure instead of direct string interoplation whenever a floating point number is used in a string interpolation.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Changed all floating point direct string interpolation to use `CssSizeLength`. |
| 🇨🇳 Chinese | ? |

There shouldn't be any risk nor visible changes. This should fix many unspotted existing issues.


jbatonnet:
@dotnet-policy-service agree

ElderJames:
Thank you for contribution @jbatonnet ! This is a major discovery!

jbatonnet:
Okay, I finally made it wok to test in the demo.
FYI, the demo forces either en-US or zh-CN locales, hence hiding all other locales potentiel issues.

Locally, I added "en-SE" in `Routes.razor` and copied *.en-US,.json to *.en-SE.json
Then, I was able to navigate to /en-SE/components/affix, see the issue, apply this fix and see it fixed.

I'll see if I can add/modify tests in this regard.

jbatonnet:
I agree, the project would benefit from such a modification. Now in many places those numbers are computed in place with +, -, * and / operators with other variables or consts.
It could be a big refactoring and we would need to fully implement a proper CssSizeLength with all needed math operations.

ElderJames:
Hello @jbatonnet , will you make the latest change in this PR?

PR #4116, opened 2024-08-22
i18n: remove comments from locale json files

ogix:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

After updating to v20.0.2 we started getting exception:
`JsonReaderException: '/' is invalid after a value. Expected either ',', '}', or ']'. LineNumber: 3 | BytePositionInLine: 4.`

After some investigation I have noticed that there is a problem reading `lt_LT` locale. Then I noticed that there are comments in lt locale json file and JSON deserialized does not expect them.

So this PR just removes comments that causes exception.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | remove comments from lt locale json file  |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/locales/bn-BD.json
+++ b/components/locales/bn-BD.json
@@ -1,13 +1,10 @@
 {
   "locale": "bn-BD",
   "Pagination": {
-    // Options
     "items_per_page": "/ পৃষ্ঠা",
     "jump_to": "যাও",
     "jump_to_confirm": "নিশ্চিত",
     "page": "পৃষ্ঠা",
-
-    // Pagination
     "prev_page": "আগের পৃষ্ঠা",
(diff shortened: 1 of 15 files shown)


ElderJames:
Thanks for contribution @ogix , we make a mistake. Would you please help fix other locale files?

ogix:
ok, I can check other files as well

ogix:
updated

ElderJames (approved):
Thank you!

Issue #4231, opened 2024-09-23
Set search input in Select manually in code


Velociraptor45:
### Is your feature request related to a problem? Please describe.
I need to manage the search input of a Select (Search enabled) in a variable outside the Select component (just like e.g. "Value" is able to set the value from outside).

### Describe the solution you'd like
I'd like to have a property (SearchInput) on the Select that sets the internal search input & executes the search if the new value is different from the old value (in case search is enabled).

If you're ok with that proposal, I'd try to implement it myself and create a PR?

ElderJames:
Thanks for contacting us @Velociraptor45 , sure you can create a PR for this 😁.

PR #4241, opened 2024-09-27
feat(module: select): Introduce SearchInput parameter for setting seach input manually

Velociraptor45:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

I need to manage the search input that's displayed in the Select outside the component, see #4231.

### 💡 Background and solution

This PR introduces the `SearchInput` parameter that lets the user set the search input from outside, which brings one big important note: If this parameter is set, we cannot manage this value on the inside anymore (e.g. clear it on selecting a value or on unfocus), but the user has to do it on their own. Otherwise, we'd get conflicting states inside the component and outside of it (you'll see this especially in the `SelectBase` when the `_searchValue` is only cleared if `_manageSearchInputExternally` is `false`).
However, I tried to keep the changes a non-invasive as possible.

I'm conflicted as to naming the parameter `SearchInput` or `SearchValue`. The latter would conform to the existing backing field, whereas the former is more user-friendly in my opinion. Let me know what you think.

I tried to write a test for ever test case there is, but maybe double-check if I forgot some.
I also did intensive manual testing with all three Select types (default, multiple, tags) to make sure this works.
There are no breaking changes.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Add SearchInput property that lets the user set & manage the search input outside the component |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -437,9 +437,12 @@ namespace AntDesign
                 await OnValueChangeAsync(_selectedValue);
             }
             
-            if(_searchInputHasChanged && !string.IsNullOrEmpty(_searchValue))
+            if(_searchInputHasChanged)
             {
-                FilterOptionItems(_searchValue);
+                if(string.IsNullOrEmpty(_searchValue))
+                    ResetOptionItems();
+                else
+                    FilterOptionItems(_searchValue);
             }
             
             await base.OnParametersSetAsync();
@@ -1202,14 +1205,19 @@ namespace AntDesign
             }
             else
             {
-                UnhideSelectOptions();
-                if (SelectMode == SelectMode.Tags && CustomTagSelectOptionItem is not null)
-                {
-                    SelectOptionItems.Remove(CustomTagSelectOptionItem);
-                    CustomTagSelectOptionItem = null;
-                }
+                ResetOptionItems();
             }
             OnSearch?.Invoke(_searchValue);
+        }
+
+        private void ResetOptionItems()
+        {
+            UnhideSelectOptions();
+            if (SelectMode == SelectMode.Tags && CustomTagSelectOptionItem is not null)
+            {
+                SelectOptionItems.Remove(CustomTagSelectOptionItem);
+                CustomTagSelectOptionItem = null;
+            }
         }
 
         private async Task TokenizeSearchedPhrase(string searchValue)
--- a/components/select/Select.razor.cs
+++ b/components/select/Select.razor.cs
@@ -1311,7 +1311,7 @@ namespace AntDesign
             }
         }
 
-        private void FilterTagsOptionItems(string searchValue)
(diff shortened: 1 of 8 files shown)


Velociraptor45:
@dotnet-policy-service agree

ElderJames:
Thanks for contribution @Velociraptor45 , what is the difference between SearchInput and SearchValue? I thought you wanted custom input controls.

Velociraptor45:
Hey there, @ElderJames, thanks for taking a look at the PR.

The tricky thing here is that there are two components that manage their own input state.
- The `SelectContent` hosts the actual HTML input field and manages the value that's inside that input field (in `_inputString`)
- The `Select` gets a callback from `SelectContent` upon user input into the input field (after the debouncer is expired) and executes the search. However, it again keeps the state of the search value (`_searchValue`) inside its own component.

Previously, the only place where the user-invoked change of the input value could take place was the HTML input field (managed by `SelectContent`). Clearing the `_inputString` in `SelectContent` was triggered via an internally accessed `ClearSearch` method in `SelectBase` upon unfocus or item selection.
But now it's also possible that the input value comes from the `Select` component as that's the entrypoint for setting the `SearchInput`. Thus, we have to use the already existing propagation of `_searchValue` (from `Select`) into `SearchValue` (in `SelectContent`) and make sure this updates `_inputString`.

I tried to draw the relationship:
Select

Did that answer your question?

What I'm aiming at is a parameter that sets the value in the input field, no matter what happens inside the Select component. Just like the `Value` property, I want the `Select` to display exactly what I tell it to (via the new `SearchInput` parameter).

use_select

If the user types characters into the input field, the `OnSearch` callback in the `Select` is eventually triggered, and I can grab that and update the search input that I manage in my own code. This will achive the same behavior as today, using the `SearchInput`. However, it also enables me to alter the `SearchInput` from other places in my code even if there was no user input.

select_process

ElderJames:
Ok, that's fine. The parameter name need to be changed to `SearchValue`, and the behavior should like the ant design react.


image

Velociraptor45:
@ElderJames, I will do the name change, but could you elaborate on the "ant design react behavior" part? I'm not a react dev and can't check for myself real quick. What's the exact behavior of that parameter in your react solution? The documentation is not really verbose on that.

ElderJames:
It would show the searchValue allways , only can't be changed by the inputbox that only can be changed by code.

image

Velociraptor45:
@ElderJames Thanks for the clarification. With that requirement, I'll close the PR & issue as this would defeat the purpose of a search-select from my pov.

PR #4294, opened 2024-10-22
Move `Dialog` MaskClick Function into MouseUp Event Callback

### LINKED ISSUE #4272: Mousedown on the modal content and mouseup on the mask will sometimes close the modal
### Describe the bug
When using `Modal` and set `MaskClosable`, mousedown on the modal content and then mouseup on the mask will sometimes close the modal.   

**The problem will be easier reproduced on low-performance machine.**

All the `Modal` With `Form` can reproduce. So just use the code in  #3494.

### Exceptions (if any)
None

### Further technical details
 `0.20.4.2`
None.
`Rider 2024.2.6 `
- .Net
`8.0`

chazikaifa:
Move `Dialog` MaskClick Function into MouseUp Event Callback

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

[Mousedown on the modal content and mouseup on the mask will sometimes close the modal #4272](https://github.com/ant-design-blazor/ant-design-blazor/issues/4272)
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Move `Dialog` MaskClick Function into MouseUp Event Callback. Fixed #4272 |
| 🇨🇳 Chinese | 将 `Dialog` 遮罩层的点击事件功能移至鼠标松开回调。修复 #4272  |

### DIFF
--- a/components/modal/core/Dialog.razor.cs
+++ b/components/modal/core/Dialog.razor.cs
@@ -159,6 +159,11 @@ namespace AntDesign
             _dialogMouseDown = true;
         }
 
+        private void OnMaskMouseDown()
+        {
+            _dialogMouseDown = false;
+        }
+
         private async Task OnMaskMouseUp()
         {
             if (Config.MaskClosable && !_dialogMouseDown)
--- a/components/modal/core/Dialog.razor.cs
+++ b/components/modal/core/Dialog.razor.cs
@@ -161,19 +161,11 @@ namespace AntDesign
 
         private async Task OnMaskMouseUp()
         {
-            if (Config.MaskClosable && _dialogMouseDown)
-            {
-                await Task.Delay(4);
-                _dialogMouseDown = false;
-            }
-        }
-
-        private async Task OnMaskClick()
-        {
             if (Config.MaskClosable && !_dialogMouseDown)
             {
                 await CloseAsync();
             }
+            _dialogMouseDown = false;
         }
 
         #endregion mask and dialog click event
--- a/components/modal/core/Dialog.razor
+++ b/components/modal/core/Dialog.razor
@@ -8,7 +8,6 @@
             <div class=@($"{Config.PrefixCls}-mask {GetMaskClsName()}") style="@Config.MaskStyle"></div>
         }
         <div tabindex="-1" class=@($"{Config.PrefixCls}-wrap {Config.GetWrapClassNameExtended(this.Status)}") role="dialog"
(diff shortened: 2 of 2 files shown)


ElderJames:
Thanks for contacting us @chazikaifa .

components/modal/core/Dialog.razor.cs
```diff
@@ -160,20 +160,12 @@ private void OnDialogMouseDown()
...
-            }
-        }
-
-        private async Task OnMaskClick()
         {
             if (Config.MaskClosable && !_dialogMouseDown)
             {
                 await CloseAsync();
             }
+            _dialogMouseDown = false;
```
ElderJames, line 168: I think it also need to add the  OnMaskMouseDown to set _dialogMouseDown=true, because some time the OnMaskMouseUp would not be trigger and then click on the mask, it doesn't close.

chazikaifa: The Dialog should only close when `OnDialogMouseDown = false`.
`_dialogMouseDown` will be set to true in `OnDialogMouseDown`, which is bind to `onmousedown` of the dialog window. When users mousedown in the dialog window (`_dialogMouseDown = true`) and then drag out and mouseup on the mask, it should not close the dialog. 
The element of dialog window is a child of mask, and there is no `stopPropagation` on the mouseup event. The `_dialogMouseDown` should be set to false correctly when user clicking inside the dialog window.

But you do remind me that if user add elements that have event `mousedown:stopPropagation` in the content, clicking on these elements will close the dialog incorrectly.

chazikaifa: I found that in antd (Node.Js), it has the same problem. Since setting `mousedown:stopPropagation` is not normal, maybe it's not a problem

ElderJames: Please look at this screenshot, when you select the text and drag it to outside of modal, then click on the mask, the modal doesn't close until it was clicked again.

modal-closing

chazikaifa: I got what you mean. I should add `OnMaskMouseDown` to set `_dialogMouseDown = false` to prevent this problem. I will update the code tonignt.

ElderJames (approved):

Issue #4177, opened 2024-09-09
using GenerateColumns and general columns at the same time will course header and cell not align properly


lujie19930803:
### Describe the bug

table has ScrollX and ScrollY at the same time will course header and cell not align properly

### Steps to reproduce (please include code)
i have a table with proerty ScrollX and ScrollY at the same time. this course  header and cell not align properly. but when i only use ScrollX. the table work correctly.

### Exceptions 
![image](https://github.com/user-attachments/assets/1c470d8f-d84f-42a0-83d2-06fc8107c3d7)

![image](https://github.com/user-attachments/assets/87cdc4e2-0878-4d7a-a327-85dee7518441)

lujie19930803:
@ElderJames

ElderJames:
It seems works well
![image](https://github.com/user-attachments/assets/4ae5cb4a-242e-4cdb-9b09-1b619b97cfbc)

lujie19930803:
sorry this happed when i use GenerateColumns. Below is my simple code.

```
<Table TItem="Data" DataSource="@data" ScrollBarWidth="7px" ScrollX="1000" AutoHeight>
      <GenerateColumns Range="1.." Definitions=definitions />
      <ActionColumn Title="Action">
          <Space Size=@("middle")>
              <SpaceItem>
                  <a>aaa</a>
              </SpaceItem>
              <SpaceItem>
                  <a>Delete</a>
              </SpaceItem>
          </Space>
      </ActionColumn>
</Table>
```

```
@code {
    Data[] data = new Data[]
        {
        new()
        {
            Key = "1",
            Name = "John Brown",
            Age = 32,
            Address = "New York No. 1 Lake Park",
            Tags = new[] {"nice", "developer"},
            IdCard = "",
            Key1 = "1",
            Name1 = "John Brown",
            Age1 = 32,
            Address1 = "New York No. 1 Lake Park",
            Tags1 = new[] {"nice", "developer"},
            IdCard1 = "",
            Key2 = "2",
            Name2 = "Jim Green",
            Age2 = 42,
            Address2 = "London No. 1 Lake Park",
            Tags2 = new[] { "loser"},
            IdCard2 = ""
        },
        new()
        {
            Key = "2",
            Name = "Jim Green",
            Age = 42,
            Address = "London No. 1 Lake Park",
            Tags = new[] { "loser"},
            IdCard = "",
            Key1 = "2",
            Name1 = "Jim Green",
[...]

Issue #3697, opened 2024-02-23
Table Column Definition with fixed right column causes table view distorted


AathiraKhusiBhaskar:
### Describe the bug

![image](https://github.com/ant-design-blazor/ant-design-blazor/assets/131652936/95b6a0e3-526c-4458-ac14-81375000cb61)

### Steps to reproduce (please include code)
![image](https://github.com/ant-design-blazor/ant-design-blazor/assets/131652936/25270e2a-cea9-43c2-abeb-c508c9f6e3e3)

ElderJames:
Note: this is because the action column is render faster then generate columns,(wapper issue)

lujie19930803:
is this problem fixed. i also have this problem

shuangbaojun:
I've had the same problem @ElderJames

ElderJames:
This issue haven't been fixed @lujie19930803 @shuangbaojun

ElderJames:
Hello @shuangbaojun @lujie19930803 @AathiraKhusiBhaskar , you can use custom colindex to fix this issue, see #4343 . We will continue to try to automate the calculation.

PR #4343, opened 2024-12-01
Add disable auto column index feature

GoldSucc:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [x] Other (about what?)

### 💡 Background and solution

The problem is that component initialisation order is not guaranteed so assigning column index by column context is very brittle and in case if some columns in table have nested component structure it can mess up order and incorrect indexes will be assigned which leads to broken table. With this change, there is an option to disable automatic column index assignment through context and assign indexes to columns by hand. It does not change behaviour with default parameter values so if AutoColIndex property isn't assigned to table, nothing changes. Alternative way of stabilising table is to assign columns as an array of render fragments, however it will require significant changes in API.

### 📝 Changelog

Risks are new exposed API which then can be difficult to rollback once it will be used in projects.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/table/GenerateColumns.razor.cs
+++ b/components/table/GenerateColumns.razor.cs
@@ -51,6 +51,8 @@ namespace AntDesign
             {
                 showPropertys = _propertyInfos[Range.Value];
             }
+
+            var colIndex = 0;
             foreach (var property in showPropertys)
             {
                 if (HideColumnsByName.Contains(property.Name)) continue;
@@ -59,8 +61,10 @@ namespace AntDesign
                 
                 if (instance != null) 
                 {
-                    instance.ColIndex = i / 3;
+                    instance.ColIndex = colIndex;
                 }
+
+                colIndex++;
 
                 Definitions?.Invoke(property.Name, instance);
 
--- a/components/table/GenerateColumns.razor.cs
+++ b/components/table/GenerateColumns.razor.cs
@@ -56,6 +56,11 @@ namespace AntDesign
                 if (HideColumnsByName.Contains(property.Name)) continue;
                 var columnType = typeof(Column<>).MakeGenericType(property.PropertyType.GetUnderlyingType());
                 var instance = Activator.CreateInstance(columnType) as IFieldColumn;
+                
+                if (instance != null) 
+                {
+                    instance.ColIndex = i / 3;
+                }
 
                 Definitions?.Invoke(property.Name, instance);
 
--- a/components/table/GenerateColumns.razor.cs
+++ b/components/table/GenerateColumns.razor.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using AntDesign.Core.Documentation;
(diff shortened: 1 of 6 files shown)


GoldSucc:
@dotnet-policy-service agree

ElderJames:
Thanks for contributionn @GoldSucc , there are still some improvements to be made here:

1. Please revert changes that are not relevant to this submission, such as line breaks and spaces. Because his will hinder the evaluation of the true scope of the changes.
2. Please add a attribute `[PublicApi(1.1.0)]` for the new public parameters.
3. I think it's good to add this new feature to the [generate columns(https://antblazor.com/en-US/components/table#components-table-demo-generate-columns) demo, because this would solve the issue around it, such as #4177 #3697.

GoldSucc:
@ElderJames I will try to make changes on weekends, having a very busy week so far

GoldSucc:
@ElderJames I made changes to documentation of GenerateColumns and also reverted changes to formatting, added PublicApi attribute to new public attributes

components/table/GenerateColumns.razor.cs
```diff
@@ -56,6 +56,11 @@ protected override void BuildRenderTree(RenderTreeBuilder builder)
                 if (HideColumnsByName.Contains(property.Name)) continue;
                 var columnType = typeof(Column<>).MakeGenericType(property.PropertyType.GetUnderlyingType());
                 var instance = Activator.CreateInstance(columnType) as IFieldColumn;
+                
+                if (instance != null) 
+                {
+                    instance.ColIndex = i / 3;
```
ElderJames, line 62: Why the ColIndex is i / 3?

components/table/Table.razor.cs
```diff
@@ -69,6 +70,10 @@ public partial class Table<TItem> : AntDomComponentBase, ITable, IEqualityCompar
         private bool _shouldRender = true;
         private int _parametersHashCode;
 
+        [Parameter]
+        [PublicApi("1.1.0")]
+        public bool AutoColIndexes { get; set; } = true;
```
ElderJames, line 75: Please add the comment for the parameter.

components/table/GenerateColumns.razor.cs
GoldSucc: for every column i is incremented thrice as I see in this loop, and col index should be i / 3 because of this.

ElderJames: It's better to define another varibale for col index

GoldSucc:
@ElderJames I added variable for column index as separate variable, code readability is improved now

ElderJames:
I think the GenerateColumns need to get the current colindex because it may be placed between other columns.

GoldSucc:
@ElderJames Start column index for generated columns added

ElderJames (approved):

Issue #4378, opened 2024-12-26
Error when fetching data for the table (dynamic data sample)


zandiarash:
There is an error when fetching data for the dynamic data sample of the Table component from Github when debuging in local environment.  
![Untitled](https://github.com/user-attachments/assets/63de955c-aaa9-49a4-8ee5-e7cbfd15e167)

Please visit [pull request #4379](https://github.com/ant-design-blazor/ant-design-blazor/pull/4379)
### Further technical details
- AntDesign Nuget Package version 1.0.1
- `dotnet --info` 9.0.100
- The IDE is Visual Studio 17.13.0 Preview 1.0
- Error Happens in **ServerMode**

PR #4379, opened 2024-12-26
Fixed error when fetching data for the table (dynamic data sample) in local.

zandiarash:
### 🤔 This is a ...
- [ X] Bug fix

### 🔗 Related issue link
Please visit [issue #4378](https://github.com/ant-design-blazor/ant-design-blazor/issues/4378)

### 💡 Background and solution
**Issue reason :**
The GitHub API requires a User-Agent header to identify the client making the request.
For Dynamic data sample of the Table component **The error 403 (Forbidden)**  is happening in local debugging means that the GitHub API deny the request due to User-Agent Header Missing.

**Solution :**
I've added **User-Agent** to the header of request for local debugging.
Also if Github denied the request with unexpected reasons to avoid errors, I used Try/Catch.

### DIFF
--- a/site/AntDesign.Docs/Demos/Components/Table/demo/Dynamic.razor
+++ b/site/AntDesign.Docs/Demos/Components/Table/demo/Dynamic.razor
@@ -1,22 +1,30 @@
 ﻿@using System.ComponentModel
 @using AntDesign.TableModels
 
-<Table TItem="Dictionary<string, object>" OnChange="HanleChange" DataSource="@data" Loading="data==null" ScrollX="1500" PageSize="5" Size="TableSize.Small">
-  @foreach (var key in data?.FirstOrDefault()?.Keys.Take(10) ??new string[0])
-  {
-    <PropertyColumn Property=@(c=>c[key]) Title="@key"></PropertyColumn>
-  }
+<Table TItem="Dictionary<string, object>" OnChange="HanleChange" DataSource="@data" Loading="data==null" ScrollX="1500"
+       PageSize="5" Size="TableSize.Small">
+    @foreach (var key in data?.FirstOrDefault()?.Keys.Take(10) ?? new string[0])
+    {
+                    <PropertyColumn Property=@(c=>c[key]) Title="@key"></PropertyColumn>
+    }
 </Table>
 
 @inject HttpClient httpClient;
 @code {
 
-  List<Dictionary<string, object>> data;
+    List<Dictionary<string, object>> data;
 
-  string githubUrl = "https://api.github.com/repos/ant-design-blazor/ant-design-blazor/contributors?per_page=200";
+    string githubUrl = "https://api.github.com/repos/ant-design-blazor/ant-design-blazor/contributors?per_page=200";
 
-  protected async Task HanleChange(QueryModel<Dictionary<string, object>> query)
-  {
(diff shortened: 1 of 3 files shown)


site/AntDesign.Docs/Demos/Components/Table/demo/Dynamic.razor
```diff
@@ -1,22 +1,35 @@
...
-  {
-    data = await httpClient.GetFromJsonAsync<List<Dictionary<string, object>>>(githubUrl);
-  }
-}
+    protected async Task HanleChange(QueryModel<Dictionary<string, object>> query)
+    {
+        var environment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
+        if (environment == "Development")
+        {
+            httpClient.DefaultRequestHeaders.Add("User-Agent", "BlazorApp");
```
ElderJames, line 24: I think we can just add this header, which will reduce the complexity of this example. We believe that the simpler the example, the better!

zandiarash: Sure, you are right I will remove unnecessary codes.
But what about Try/Catch ?
Should I remove It too ?

zandiarash: > I think we can just add this header, which will reduce the complexity of this example. We believe that the simpler the example, the better!

Sure, you are right I will remove unnecessary codes.
But what about Try/Catch ?
Should I remove It too ?

ElderJames: Yes, it's better that add a DelegatingHandler for the HttpClient to handle the headers and exceptions, and this is out of the scope of the exsample.

zandiarash: > I think we can just add this header, which will reduce the complexity of this example. We believe that the simpler the example, the better!

Done.

ElderJames: How about using DelegatingHandler?

zandiarash: > How about using DelegatingHandler?

Do you mean adding a new class for http client then register as DI in Program.cs and using it in component ?

ElderJames: Yes, but please add it in the `AntDesign.Docs.Server`, and the finally using is in this example. So that the example is no need to be changed. 

Sorry I didn't think of this scenario until later.

zandiarash: > Yes, but please add it in the `AntDesign.Docs.Server`, and the finally using is in this example. So that the example is no need to be changed.
> 
> Sorry I didn't think of this scenario until later.

I'll do it ASP.

zandiarash: > Yes, but please add it in the `AntDesign.Docs.Server`, and the finally using is in this example. So that the example is no need to be changed.
> 
> Sorry I didn't think of this scenario until later.
@ElderJames 
That's done now, Please check commits again.

ElderJames (approved):
That's all well, thank you!

PR #4402, opened 2025-01-04
Adding WrapperClass for Input

zandiarash:
### 🤔 This is a ...
✅New feature

### 💡 Background and solution

Currently for **input** tags which have prefix and suffix does not have a feature to add  custom class to the **Wrapper Span**.
Suppose that there are a bunch of inputs in a form in the page which want to are expected to have same margins now we must use **WrapperStyle** for all of them with redundant styles.
Does it make sense to have a **WrapperClass** ?

✅Doc is updated
✅Demo is updated
✅Changelog is provided


components/input/Input.cs
```diff
@@ -252,6 +253,17 @@ public int DebounceMilliseconds
...
 
+        /// <summary>
+        /// Set Class of wrapper. Is used when component has visible: Prefix/Suffix
+        /// or has paramter set <seealso cref="AllowClear"/> or for components: <see cref="InputPassword"/>
+        /// and <see cref="Search"/>. In these cases, html span elements is used
+        /// to wrap the html input element.
+        /// <seealso cref="WrapperClass"/> is used on the span element.
+        /// </summary>
+        [Parameter]
+        [PublicApi("1.1.2")]
```
ElderJames, line 264: New API should release in minor version, so next minor version is 1.2.0.

ElderJames:
Please add a test for this new API.

components/input/Input.cs
zandiarash: Done.

zandiarash:
> Please add a test for this new API.

I've added test and all tests has been passed.

ElderJames (approved):
Thanks.

ElderJames:
Hello @zandiarash , could you please rebase the commit on feature branch? Just need the latest commit.

zandiarash:
> Hello @zandiarash , could you please rebase the commit on feature branch? Just need the latest commit.

Sure I do it ASAP.

Issue #4569, opened 2025-05-03
Error with Tree component when using StateHasChanged method


DarkElfes:
### Describe the bug

Bug with render TreeNode components inside Tree component when using StateHasChanged method. That is, when i have some list for DataSource parameter, in first time all items from this list will be rendered, but when i change this collection and call StateHasChanged method, it's should re-render all items like TreeNode again, but in your foreach cycle you use @key property for TreeNode, which check if item not was changed - if true, then this TreeNode should not be re-rendered. In result, not all items will be added to _allNodes collection in Tree comoponent.

### Steps to reproduce (please include code)
> https://drive.google.com/file/d/1H7ilh632wBSvuvzECuGHMnG8s7-xnVh_/view?usp=sharing

### Further technical details
- AntDesign Nuget Package version: 1.3.2
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version: VS

DarkElfes:
Hi, I don't understand why you can't view the link, because by default it's a video from Google Drive, but if it doesn't open for some reason, I can add it to YouTube and give you the link. In this video, it was easier for me to explain where the problem was in the code.

But if not, here's a quick code sample that describes the error. So the first time you run it, when you select any items, the selected key remains only 1, but if you add a new item and refresh the tree component, you will see that you can select multiple items at the same time

```
<Tree TItem="Data"
      Selectable
      TitleExpression="x => x.DataItem.Title"
      DataSource="dataSource" />

<Button OnClick="AddItem">Add new item</Button>

@code {
    private List<Data> dataSource = [];

    protected override void OnInitialized()
    {
        dataSource = GetNewData(5);
    }


    private List<Data> GetNewData(int count)
    {
        return [
            ..dataSource,
            ..Enumerable.Range(1, count)
            .Select(i => new Data(Guid.NewGuid().ToString(), $"Data {dataSource.Count + i}"))
            .ToList()
        ];
    }

    private void AddItem()
    {
        dataSource = GetNewData(1);
        StateHasChanged();
    }


    private record Data(string Id, string Title);
}
```

DarkElfes:
![Image](https://github.com/user-attachments/assets/663a9e30-e575-4331-bb34-cbfd887dd288)

I create gif, for show you this problem here

PR #4575, opened 2025-05-07
fix(module: tree): change the logic for updating the data source

DarkElfes:
- Removed code that clears the _allNodes and ChildNodes collections when updating the DataSource parameter
- Added the Dispose method for the TreeNode component to remove self from the collections of parent child nodes and tree nodes
- Convert the IsLastNode property to a method and check the predicate through the parent children or the tree of all node collections
- Remove the NodeIndex property, which was used only in the previous property, because when a some node is deleted, this index property should be updated for all nodes after it.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

> See: https://github.com/ant-design-blazor/ant-design-blazor/issues/4569

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tree/Tree.razor.cs
+++ b/components/tree/Tree.razor.cs
@@ -45,6 +45,11 @@ namespace AntDesign
         internal List<TreeNode<TItem>> _allNodes = new List<TreeNode<TItem>>();
 
         /// <summary>
+        /// Readonly collection of all nodes
+        /// </summary>
+        public IReadOnlyCollection<TreeNode<TItem>> AllNodes => _allNodes.AsReadOnly();
+
+        /// <summary>
         /// All the checked nodes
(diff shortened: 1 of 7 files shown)


DarkElfes:
@dotnet-policy-service agree

ElderJames (approved):

ElderJames:
Thank you for contribution @DarkElfes !

PR #4626, opened 2025-06-23
feat(module: upload):manual upload

stfei:
### 🤔 This is a ...

- [X] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog
Added the manual upload functionality attribute UseManuallyMode; when the attribute is set to true, the file will not be uploaded immediately after it is changed, but will wait for the StartUpload method to execute before calling the upload script.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/upload/Upload.razor.cs
+++ b/components/upload/Upload.razor.cs
@@ -4,8 +4,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
+using AntDesign.Internal;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
 using OneOf;
@@ -242,6 +244,14 @@ namespace AntDesign
         [Parameter]
         public bool WithCredentials { get; set; }
 
+        /// <summary>
+        /// Use the manual upload mode
+        /// </summary>
+        [Parameter]
+        public bool UseManuallyMode { get; set; } = false;
+
+        private UploadButton _uploadButton;
+
         private bool IsText => ListType == UploadListType.Text;
 
         private bool IsPictureCard => ListType == UploadListType.PictureCard;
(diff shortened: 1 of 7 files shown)


stfei:
@dotnet-policy-service agree

ElderJames:
Hi @stfei , 

I appreciate your effort in adding the new feature to the Upload component! After considering the naming of the property, I think **"Defer"** would be a better fit than **"UseManuallyMode"** for the following reasons:

1. **Conciseness**: "Defer" is shorter and avoids redundancy with the component name "Upload".
2. **Clarity**: It clearly conveys the intended behavior of delaying the upload until the user explicitly triggers it.
3. **Technical Convention**: "Defer" is a common term in programming (e.g., JavaScript's `defer` attribute, Go's `defer` keyword), making it more familiar to developers.

Could you please update the PR to use **`Defer`** as the property name? Let me know if you have any concerns!

Thanks again for your contribution!

Best regards,
James

stfei:
@ElderJames  Should we modify it and then re-submit a PR?

components/upload/Upload.razor.cs
```diff
@@ -242,6 +244,14 @@ public partial class Upload : AntDomComponentBase
         [Parameter]
         public bool WithCredentials { get; set; }
 
+        /// <summary>
+        /// Use the manual upload mode
+        /// </summary>
+        [Parameter]
+        public bool UseManuallyMode { get; set; } = false;
```
LeaFrock, line 251: 可否考虑新增`UploadModeEnum`枚举而不是bool值，以保留将来更多方式的扩展性？

ElderJames:
> @ElderJames Should we modify it and then re-submit a PR?

You just need to continue to commit to this branch.

components/upload/Upload.razor.cs
ElderJames: I think there are only two case

components/upload/Upload.razor.cs
```diff
@@ -248,7 +249,7 @@ public partial class Upload : AntDomComponentBase
         /// Use the manual upload mode
         /// </summary>
         [Parameter]
-        public bool UseManuallyMode { get; set; } = false;
+        public bool Defer { get; set; } = false;
```
ElderJames, line 252: It‘s also need [PublicApi("1.5.0")] here.

PR #4652, opened 2025-07-08
faat(module: mentions): supports customizable prefixes

### LINKED ISSUE #4649: Support multiple prefix for Mention component
### Describe the bug

As per documentation Mention component accepts list of prefix , however it seems component only support single prefix. Also, if I change the prefix to "#" after select it comes as '@'. 

### Expectation
1. After select it should consider the prefix for which it got triggered instead of default '@'
2. I could see React supports multiple prefix , is there any plan for the same?
3. Another feature could be to support fragments for selected item. eg, instead of simple text it could be **Badge**

ElderJames:
- Added a new `Prefix` parameter for customizable prefix symbols (e.g., "@", "#") and introduced an `OnSearch` event callback for handling search actions based on the prefix.
- Modified internal logic to populate `ShowOptions` based on the current prefix and updated mention name extraction to return both name and prefix.
- Introduced a `Hidden` property in the `MentionsOption` component to control visibility based on the `ShowOptions` list.
- Enhanced the `Prefix` component to demonstrate the new functionality, binding the `Value` property and handling search events for user and tag options.

These changes improve the usability and functionality of the `Mentions` component, providing a more dynamic user experience.

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

fixed #4649

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/mentions/Mentions.razor.cs
+++ b/components/mentions/Mentions.razor.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using AntDesign.Core.Documentation;
 using AntDesign.Internal;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
@@ -26,7 +27,7 @@ namespace AntDesign
     <seealso cref="MentionsOption" />
     */
     [Documentation(DocumentationCategory.Components, DocumentationType.DataEntry, "https://gw.alipayobjects.com/zos/alicdn/jPE-itMFM/Mentions.svg", Title = "Mentions", SubTitle = "提及")]
-    public partial class Mentions
+    public partial class Mentions : AntDomComponentBase
     {
         [Parameter]
         public RenderFragment ChildContent { get; set; }
@@ -62,7 +63,7 @@ namespace AntDesign
         public EventCallback<string> ValueChanged { get; set; }
 
         /// <summary>
-        /// Dynamically load mention options for display when the user types a value after the @ symbol
+        /// Dynamically load mention options for display when the user types a value after the prefix symbol
         /// </summary>
         [Parameter]
         public Func<string, CancellationToken, Task<IEnumerable<MentionsDynamicOption>>> LoadOptions { get; set; }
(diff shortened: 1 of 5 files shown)


components/mentions/Mentions.razor.cs
```diff
@@ -92,14 +101,34 @@ private void SetClassMap()
...
+        protected override void OnParametersSet()
+        {
+            base.OnParametersSet();
+            InitializePrefixes();
+        }
+
+        private void InitializePrefixes()
+        {
+            _prefixes = (Prefix ?? "@")
+                .Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
```
LeaFrock, line 123: Use `Split(',' , StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)` so that we do not need to handle the trimming manually next step.

components/mentions/Mentions.razor.cs
```diff
@@ -216,19 +237,28 @@ private async Task OnInput(ChangeEventArgs args)
                 return;
             }
 
-            // 检查光标前的文本中是否存在 @ 符号
+            // Get text before cursor
             var textBeforeCursor = Value.Substring(0, focusPosition);
```
LeaFrock, line 241: Use `Value.AsSpan(0, focusPosition)`?

components/mentions/Mentions.razor.cs
```diff
@@ -238,17 +268,17 @@ private async Task OnInput(ChangeEventArgs args)
...
             }
 
-            // 获取 @ 后面的搜索文本
-            var searchText = textBeforeCursor.Substring(lastAtIndex + 1);
+            // Get search text after the prefix
+            var searchText = textBeforeCursor.Substring(lastPrefixIndex + _currentPrefix.Length);
 
-            // 如果搜索文本中包含空格，说明已经选择完成，隐藏选项
+            // If search text contains space, selection is complete
             if (searchText.Contains(" "))
```
LeaFrock, line 275: Use the overload `searchText.Contains(' ')`?

components/mentions/Mentions.razor.cs
```diff
@@ -285,22 +318,15 @@ private void SetOptionsToShow(IEnumerable<MentionsDynamicOption> optionsToShow)
...
-        }
-
         internal async Task ItemClick(string optionValue)
         {
             var focusPosition = await JS.InvokeAsync<int>(JSInteropConstants.GetProp, _overlayTrigger.Ref, "selectionStart");
             var preText = Value.Substring(0, focusPosition);
-            preText = preText.LastIndexOf(Prefix) >= 0 ? Value.Substring(0, preText.LastIndexOf(Prefix)) : preText;
+            preText = preText.LastIndexOf(_currentPrefix) >= 0 ? Value.Substring(0, preText.LastIndexOf(_currentPrefix)) : preText;
             if (preText.EndsWith(' ')) preText = preText.Substring(0, preText.Length - 1);
             var nextText = Value.Substring(focusPosition);
```
LeaFrock, line 327: All `Substring` should be replaced with `AsSpan` so that we can avoid generating a lot of temp strings that lead to fragmented memory allocation.

anktsrkr:
Any plan to merge this?

PR #4685, opened 2025-09-23
fix(module: datepicker): fix for net10 bug on DatePicker

alchiweb:
On a .NET 10 (rc1) project, the DatePicker has bugs: for example, the year and month top buttons are useless (nothing on click).

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

This is a problem with StateHasChanged -> on DatePickerBase, the call to StateHasChanged don't resfresh the overlay component.
I don't know if this is due to CascadingValue behavior in .NET 10...

This workaround/fix will work for the DatePicker bug, but I suspect that other components that use Overlay (and OverlayTrigger) are also bugged if they use StateHasChanged...

If someone has an idea to better fix this, it will be very useful for the .Net 10 compatibility!

Notes :
 - this fix is useless for other versions of .Net (tested with .Net 9), but it will also work.
 - not needed to compile AntDesign package on .Net 10 (AntDesign components on .Net 9 with project on .Net 10 will work with this fix).
 
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/Component/Overlay/Overlay.razor.cs
+++ b/components/core/Component/Overlay/Overlay.razor.cs
@@ -403,6 +403,14 @@ namespace AntDesign.Internal
         }
 
         /// <summary>
+        /// Refresh the state of the component
+        /// </summary>
+        /// <remarks>
+        /// Call to StateHasChanged()
+        /// </remarks>
+        internal void RefreshComponentState() => StateHasChanged();
+
+        /// <summary>
         /// Will probe a check predicate every given milliseconds until predicate is true or until
         /// runs out of number of probings.
         /// </summary>
--- a/components/date-picker/internal/DatePickerBase.cs
+++ b/components/date-picker/internal/DatePickerBase.cs
@@ -552,8 +552,8 @@ namespace AntDesign
                 {
                     _activeBarStyle = "display: none";
                 }
-
-                StateHasChanged();
+                
+                RefreshComponentState();
             }
 
             _needRefresh = false;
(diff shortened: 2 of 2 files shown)


ElderJames:
Thanks for contribution @alchiweb , do you know what the issue in .NET 10 is? It's better that you create an issue to the official repo first and ask the real reason.

alchiweb:
> do you know what the issue in .NET 10 is? It's better that you create an issue to the official repo first and ask the real reason.

Thanks for the answer.
No, I don't know... I will try to better understand before creating a .NET 10 issue (I'm not sure if this is a .NET 10 issue, and I'm new in Ant Design Blazor).
But, I found a strange behavior in Ant Design Blazor on .NET 9, with the DatePickerPanelChooser: when the date picker is displayed, it seems that the DatePickerPanelChooser is continuously refreshed. I saw that with a breakpoint on the first line of the DatePickerPanelChooser.razor.
And in .NET10, it is refreshed just 2 times on the opening, and not continuously after. It seems more normal for me (especially since the Trigger in the OverlayTrigger of the DatePicker.razor is only defined with Trigger.Click, and not Hover or Focus).

ElderJames (approved):

ElderJames:
This PR fixed https://github.com/ant-design-blazor/ant-design-blazor/issues/4732, thank you!

alchiweb:
> This PR fixed #4732, thank you!

Cool! Unfortunately, I didn't have time to test it further... But it seems to work.

Issue #4059, opened 2024-08-02
Input in modal can not be focused automatically


pankey888:
### Describe the bug

```
<Input Width="200px" TValue="string" AutoFocus />
<Button Type="primary" OnClick="()=>{_visible = true;}">
    Open Modal
</Button>
<Modal Title="BasicModal" Width="300" Footer="null"
       @bind-Visible="@_visible">
    <Input Width="200px" TValue="string" AutoFocus />
</Modal>

@code{
    bool _visible = false;
}

```
![20240802205720](https://github.com/user-attachments/assets/df286839-7426-47e1-b31a-5e1d275bfb69)

### Further technical details
- AntDesign Nuget Package version
- Include the output of `dotnet --info`
- The IDE (VS / VS Code/ VS4Mac) you're running on, and it's version

PR #4692, opened 2025-10-07
Fix the problem that the autofocus of input in the pop-up window is invalid. It is a temporary solution.

bxjg1987:
### 🤔 this is a
Bug fix

### 🔗 Related issue link
[Input in modal can not be focused automatically](https://github.com/ant-design-blazor/ant-design-blazor/issues/4059)

### 💡 Background and solution

When an Input component is placed inside a Dialog/Modal component, the AutoFocus feature of the Input does not work as expected. The suspected cause is the popup animation.

A complete solution is difficult, because the Input can be placed inside any type of container, which involves interaction between the container component and the slotted inner component.

The current workaround is to add an AutoFocusDelay parameter. When the Input is rendered for the first time, it waits for AutoFocusDelay milliseconds before setting focus on the input box. In testing, this works reasonably well for Inputs inside a modal:
'''<Input @bind-Value="@context.Username" AutoFocus AutoFocusDelay="700" />'''

Note: By default, when the modal is closed, the DOM is not destroyed. Therefore, when reopening the modal, the Input still cannot be focused automatically, because it only auto-focuses on the first render.

As mentioned, a thorough fix is difficult. For now, you can try referencing the Input component via @ref, and then call the Input’s Focus method inside the modal’s VisibleChanged event. This approach has not been tested.
or 
set modalOptions.DestroyOnClose=true

### 📝 Changelog
When Dialog/Modal is displayed for the first time, AutoFocus of Input will work as expected.

- [ √] Demo is updated/provided or not needed

### DIFF
--- a/components/input/Input.cs
+++ b/components/input/Input.cs
@@ -106,6 +106,11 @@ namespace AntDesign
                     IsFocused = _autoFocus;
             }
         }
+        /// <summary>
+        /// Delay of auto focus
+        /// </summary>
+        [Parameter]
+        public int AutoFocusDelay { get; set; }
 
         /// <summary>
         /// Whether has border style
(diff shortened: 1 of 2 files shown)


ElderJames:
Thanks for contribution @bxjg1987 , did you try add the `autofocus` attribute at the input element?

PR #4719, opened 2025-11-11
Refactor of InvokeStateHasChanged

zandiarash:
### 🤔 This is a

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

`InvokeStateHasChanged` is written misspelled `InvokeStateHasChagned` 

If refactor the name of `InvokeStateHasChagned` to `InvokeStateHasChanged` it **overrides** the `InvokeStateHasChagned`  of `AntDomComponentBase` and we face with this warning
```
'Card.InvokeStateHasChanged()' hides inherited member 'AntComponentBase.InvokeStateHasChanged()'. Use the new keyword if hiding was
```

Since `AntComponentBase` has a `InvokeStateHasChagned` if we remove the implementation in `Card.razor.cs` to call the method from the base class we face with this error in `Tabs.razor.cs `

```
Cannot access protected member 'AntComponentBase.InvokeStateHasChanged()' via a qualifier of type 'Card'; the qualifier must be of type 'Tabs' (or derived from it)[CS1540](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS1540))
```

So we can Make these public in `AntComponentBase` to avoid error above.
```
InvokeStateHasChanged()
InvokeStateHasChangedAsync()
```

But I think the best way is keeping the structure as is and use `new` keyword and refactor the name
`internal new void InvokeStateHasChanged()`

this way we keep the structure and not override the parent but have same behavior in the body of `InvokeStateHasChanged` which triggers parent base `InvokeStateHasChanged`

### 📝 Changelog

There is no potential break changes or other risks.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/card/Card.razor.cs
+++ b/components/card/Card.razor.cs
@@ -145,7 +145,7 @@ namespace AntDesign
             StateHasChanged();
         }
 
-        internal void InvokeStateHasChagned()
+        internal void Refresh()
         {
             StateHasChanged();
         }
--- a/components/tabs/Tabs.razor.cs
+++ b/components/tabs/Tabs.razor.cs
@@ -576,7 +576,7 @@ namespace AntDesign
                 }
                 else
                 {
-                    Card.InvokeStateHasChagned();
+                    Card.Refresh();
                 }
             }
 
@@ -869,7 +869,7 @@ namespace AntDesign
 
             if (IsTabbedCard)
             {
-                Card.InvokeStateHasChagned();
+                Card.Refresh();
             }
             else
             {


ElderJames:
Thanks @zandiarash , how about make the method to be virtual?

zandiarash:
> Thanks @zandiarash , how about make the method to be virtual?
Well actually it is the best solution.
Then I added `virtual` to the `InvokeStateHasChanged()` and `InvokeStateHasChangedAsync() ` of `AntComponentBase`
And made `InvokeStateHasChanged` in Card as `override`
Now we get rid of warning and misspell.

ElderJames (approved):
Thank you!

ElderJames:
But the build fail:  error CS0122: 'Card.InvokeStateHasChanged()' is inaccessible due to its protection level

zandiarash:
> But the build fail: error CS0122: 'Card.InvokeStateHasChanged()' is inaccessible due to its protection level

> But the build fail: error CS0122: 'Card.InvokeStateHasChanged()' is inaccessible due to its protection level

Oh sorry, that was my mistake, VS ran the last successful build,

This is right that In the `Tabs` component we cannot use a **Protected** function of `Card` even when it is **virtual** in the base class (In the **AntComponentBase** it is **_protected_**)

Thus in my opinion we have 3 solutions :

1 - Making a new **public** wrapper method in the `Card` component with the name `Refresh` and trigger `StateHasChanged` in it, then use it in `Tabs` component.
2 - Use `internal new void InvokeStateHasChanged()` like I've mentioned before.
3 - Make `InvokeStateHasChanged()` and `InvokeStateHasChangedAsync()` public in **AntComponentBase**.

I've Amended my commit with the first solution which works well, please check and give me your feedback or any other idea you have.
Thanks.

PR #4752, opened 2026-01-07
Select Loading State Change Fixed

zandiarash:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Blazor cascading parameters work when you cascade `this`.
But the cascaded value is the **Select** instance. When **Loading** changes, the instance reference **doesn’t change** (it’s still the same this), so Blazor does not consider the cascading value **“changed”**, and it won’t re-render descendants like **IsLoading** because of that change.
Therefore **Loading** will not change until you focus the select, even with `StateHasChanged()` no hope :

msedge_wwlu2FELmi

I've passed `Loading` attribute as cascading parameter like what we have for **ShowArrowIcon** and **ShowSearchIcon**
Now problem has been **solved** :

msedge_DxPi02dM4N

I've provided a minimal sample code :
```CS
@using AntDesign
@code {
    public class TypesDto
    {
        public byte Id { get; set; }
        public string Name { get; set; }
    }
    List<TypesDto> _dataSource { get; set; } = new();
    byte? valueSelected = 0;
    bool IsLoading = true;
}
<h1>Is Loading : @IsLoading</h1>
<div style="max-width:200px">
    <Select  @bind-Value="valueSelected" DataSource=_dataSource Loading=IsLoading
            ValueName="@nameof(TypesDto.Id)"
            LabelName="@nameof(TypesDto.Name)" Placeholder="This is test">
    </Select>
    <Button OnClick="@(async () => { IsLoading = !IsLoading; StateHasChanged(); })">Change State</Button>
</div>
```

### 📝 Changelog

There is no potential break changes or other risks.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


ElderJames:
Thanks for contribution @zandiarash . But I don't think adding a cascading parameter just for one property is a good design.

stfei:
我尝试的改法是在Loading的set加个调用
https://github.com/stfei/ant-design-blazor/commit/519a78b581e28a253a0029a42ecc6cdda56731c8

ElderJames:
I found that it was the IsFixed of `<CascadingValue Value="this" Name="ParentSelect" IsFixed>` that caused the issue.

PR #4805, opened 2026-05-25
fix(module: Space): SpaceItem RTL Fixed

zandiarash:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

In **Space** component **Space Items** does not shows properly in **RTL** mode, It acts like **LTR** and add a right margin.

This is **LTR** with **right** margin
LTR

This is (current) **RTL**  with **right** margin which does not behaves as expected :
RTLErr

This is **RTL** that is expected with **left** margin :
image

In addition of current issue I've also added SetClass to build **PrefixCls-rtl** class — it’s not strictly necessary, just there for consistency and it can help RTL developers to play with CSS better. If you’d prefer, I can remove it; just let me know.
```
        private void SetClass()
        {
            ClassMapper
                .Add(PrefixCls)
                .If($"{PrefixCls}-rtl", () => RTL);
        }
```

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

- [x] Doc is not needed
- [x] Demo is updated
- [x] Changelog is not needed

### DIFF
--- a/components/space/Space.razor
+++ b/components/space/Space.razor
@@ -5,7 +5,7 @@
     SpaceItemCount = 0;
 }
 <CascadingValue Value="this" IsFixed="@true">
-    <div class="@ClassMapper.Class" style="@Style @InnerStyle" id="@Id" @ref="@Ref">
+    <div class="@ClassMapper.Class" style="@Style @InnerStyle @_gapStyle" id="@Id" @ref="@Ref">
         @ChildContent
     </div>
 </CascadingValue>
--- a/components/space/Space.razor.cs
+++ b/components/space/Space.razor.cs
@@ -45,6 +45,7 @@ namespace AntDesign
             set
             {
                 _size = value;
+                ChangeSize();
             }
         }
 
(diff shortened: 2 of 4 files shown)


ElderJames:
Thanks for contribution @zandiarash , but I found that in the v4.x of antd, Space is use `gap` instead of the margin, so could you please follow the antd's way?
image

> In addition of current issue I've also added SetClass to build PrefixCls-rtl class — it’s not strictly necessary, just there for consistency and it can help RTL developers to play with CSS better. If you’d prefer, I can remove it; just let me know.

Yes, this class name is no need.

zandiarash:
I appreciate the review and guidance, @ElderJames. I have refactored the implementation to align with your feedback.
As your expectation I removed **SetClass** and checked **AntD** `Space` and `SpaceItem` behaviour,
I implemented Gap as what is happening in AntD and moved all sizing codes from `SpaceItem` to the parent `Space` (like AntD), Please check and let me know if it need additional.
By the way This shift to **gap** not only resolves the original **RTL** margin issue but also brings us in line with modern CSS standards, Thank you

ElderJames (approved):

Issue #4809, opened 2026-06-09
Dropdown overlay disappearing in MS Edge


GodEmperorOfTime:
### Describe the bug

There is an issue with `Overlay` of `Dropdown` component. Its seems overlay too quickly disappear after hovering over `Dropdown`. Its very difficult to click on any button in overlay. It sometimes work, but mostly not.

<img width="648" height="303" alt="Image" src="https://github.com/user-attachments/assets/ab640fd6-9ef4-484f-8ad8-9be47a0af64c" />

It seems the issue started to manifesting in *Microsoft Edge* version `149.0.4022.52 (61-bit)`. Previous version of *Edge*  - `148.*` worked reasonably well.

In *Firefox*, version `140.11.0esr` it's also working. 

### Steps to reproduce (please include code)

Described behavior is possible to observe on official page:

https://antblazor.com/en-US/components/dropdown

Current version  `149.0.4022.52 (61-bit)` of *Microsoft Edge* is needed.

### Exceptions (if any)

No exception 😒 Not event in debug console in browser.

### Further technical details

*AntDesign* version `1.6.1`

OS: *Windows 11 Pro* version `25H2` build `26200.8457`

| Browser | Version | Is ok |
| --- | --- | --- |
| MS Edge | 149.0.4022.52 (61-bit) | No |
| MS Edge | 148.* (61-bit) | Yes |
| Firefox | 140.11.0esr | Yes |

GodEmperorOfTime:
So I did little bit of digging and found some things. Not enough to fix, but maybe enough to give someone head start.

I figured what is triggering the issue. Its somehow dependend on cursor speed:
- When I move cursor quickly from button to overlay, overlay stay displayed.
- When I move curosr slowly from button ro overlay, overlay disapper.

When investigating `Overlay.Hide` method, I found, that the issue when "moving cursor slowly", inner state variable `_mouseInOverlay` is set to `false`, so no early return, which would interrupt overlay hiding.

<img width="1109" height="588" alt="Image" src="https://github.com/user-attachments/assets/190bfe20-367e-4b5e-910a-9209d06e273d" />

So I started investigating how `Overlay.Hide` is called and put some `Debug.WriteLine` in various places.

## Correct behavior - quick cursor move

<img width="699" height="747" alt="Image" src="https://github.com/user-attachments/assets/b82fcadf-8570-48fc-9a07-9fedd2a4bc39" />

### Output
```
[...]

PR #4812, opened 2026-06-15
fix(module: Overlay): fix dropdown hover menu auto close 

superjerry88:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Dropdown overlay disappearing in MS Edge https://github.com/ant-design-blazor/ant-design-blazor/issues/4809

### 💡 Background and solution

- Chromium: 149.x and above breaks the overlay for Dropdown. 

- Initial investigation by https://github.com/ant-design-blazor/ant-design-blazor/issues/4809 share similar test result that `Overlay.Hide` is invoked when it shouldnt be

- The rootcause was the new Chromium update causes `mouseleave` to fire when moving from the overlay container into its child menu. I cant pinpoint which changes in the chromium update causes it but this bug only exist from 149 and above.

https://github.com/user-attachments/assets/e9bae0c3-78e3-4ce7-8206-b954a2cd77c4

- The current fix is to handle `mouseout` in JS and close the overlay only when the mouse fully leaves the overlay
---
## Before

https://github.com/user-attachments/assets/c0f8acd2-e545-4598-a28d-2ba0592b89ea

## After

https://github.com/user-attachments/assets/cf7ed94f-0cf8-440a-9eb1-19bcc4021386

### 📝 Changelog
---

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/core/Component/Overlay/Overlay.razor
+++ b/components/core/Component/Overlay/Overlay.razor
@@ -10,8 +10,7 @@
          style="position: absolute; @display @Trigger.OverlayStyle @_overlayStyle"
          @ref="Ref"
          @onmouseenter="OnOverlayMouseEnter"
-         @onmouseleave="OnOverlayMouseLeave"
-		 @onmouseup="OnOverlayMouseUp"
+         @onmouseup="OnOverlayMouseUp"
          @onclick:stopPropagation>
 
         <CascadingValue Value="this" Name="Overlay" IsFixed>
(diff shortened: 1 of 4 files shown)


superjerry88:
Better solution exist: https://github.com/ant-design-blazor/ant-design-blazor/tree/fix/overlay-mouse-leave

PR #4815, opened 2026-06-16
Changelog 1.6.2

ElderJames:


superjerry88:
The Github profiles are all wrongly linked 👀 

its showing as https://github.com/James Yeung rather than @ElderJames

PR #4820, opened 2026-07-23
fix(module: Sider): Sider Trigger RTL Fixed

zandiarash:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

As you know, the `Layout` component can contain a `Sider`.
In **LTR** mode, everything works as expected. When the `Sider` is **expanded** or **collapsed**, the bottom **trigger arrow** is displayed in the **correct direction**:
image

However, in **RTL** languages, the arrow is displayed in the **opposite direction**, as if the layout were still using **LTR** mode:
image

I have updated the code so that the trigger arrow is displayed **correctly** in both **LTR** and **RTL** modes.
RTL Right

Please note that because the following line exists in the **components.en-US.json** file:
```"Iframe": 360```
The `Layout` component on the demo page does not inherit the `<ConfigProvider />` settings from `AntDomComponentBase`. Therefore, it has not `ant-layout-rtl` class and the fix will not be visible on the **demo** page.
However, I have tested the changes in a production environment, and the component works as expected.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/layout/Sider.razor.cs
+++ b/components/layout/Sider.razor.cs
@@ -40,7 +40,7 @@ namespace AntDesign
         [Parameter]
         public bool NoTrigger { get; set; }
 
-        [CascadingParameter] 
+        [CascadingParameter]
         public Layout Parent { get; set; }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


ElderJames (approved):
LGTM, thanks!

PR #4842, opened 2026-08-22
fix(module: slider): support touch dragging with pointer events

### LINKED ISSUE #3388: Sliders do not work with touch
The slider components on the demo site do not seem to work at all with touch. 
I am using Microsoft Edge on Windows 10 laptop. I also tried iPhone with Brave Browser and they still don't work although I somehow managed to move one a bit after trying multiple times

Cdorey:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [x] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Closes #3388

### 💡 Background and solution

Slider currently listens to mouse-specific events for its drag lifecycle, so touch and pen input cannot reliably drag the control.

This change uses Pointer Events for the existing interaction path, tracks the active pointer ID so unrelated pointers are ignored, and resets the interaction on `pointercancel`. The shared window callbacks remain `JsonElement`-based because all supported target frameworks provide the required page coordinates through that path, while older `PointerEventArgs` versions do not expose `PageX` and `PageY`.

There are no public API or visual style changes. Touch dragging was manually verified in a Blazor application consuming a locally built package. Because the appearance is unchanged, no screenshot is included.

### ✅ Validation

- `npm run lint`
- Slider tests: 27 passed on net6
- Full .NET tests with coverage: 3,020 passed on each of net6, net8, net9, and net10.0
- JavaScript tests with coverage: 70 passed
- Full solution build with .NET SDK 10.0.400: 0 errors after the test-only baseline workaround described below

Current `master` (`11397d04`) has four pre-existing `CS1503` errors in `TreeNodeTitle.razor` under the current .NET 10 SDK. The same errors reproduce in a detached, unmodified `master` worktree and in recent upstream PR checks. For local validation only, I parenthesized those four existing Razor directive expressions, ran the build and tests above, and then reverted that file. This PR contains no Tree changes.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Slider now supports dragging with touch and pen input in addition to mouse input. |
| 🇨🇳 Chinese | Slider 现在除鼠标外，也支持通过触摸和触控笔进行拖动。 |

### DIFF
--- a/components/slider/Slider.razor
+++ b/components/slider/Slider.razor
@@ -5,7 +5,7 @@
 @using System.Globalization
 
 <div class="@ClassMapper.Class" style="user-select:none;@Style" id="@Id" @ref="Ref"
-     @onmousedown="(e) => OnMouseDown(e)">
+     @onpointerdown="(e) => OnPointerDown(e)">
     <div class="ant-slider-rail">
     </div>
     @if (Included)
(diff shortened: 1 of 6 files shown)


Cdorey:
@dotnet-policy-service agree

PR #4843, opened 2026-08-22
test(module: pagination): add unit tests

thaildhe172591:
### This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### Related issue link

#2644

### Background and solution

Pagination had no unit tests. This adds a bUnit test class covering the paging math (pager count
for a given total and page size, the zero total case, the active page, `Current` taking precedence
over `DefaultCurrent`), prev/next and the jump-prev/jump-next ellipsis by both click and Enter,
`OnChange`, the disabled state, `HideOnSinglePage`, `Simple` mode, the mini class for small size,
RTL, `ShowTotal` ranges, `ItemRender`, `ShowTitle`, the size changer boundary and the quick jumper.

No production code is changed. Locally on net10.0 the full suite passes (3062 tests) and line
coverage for `components/pagination` goes from 1.0 percent to 82.9 percent, measured with the same
command in both directions. The tests also pass on net8 and net9, and the project still builds for
net5 and net6.

A few branches are left uncovered on purpose, where asserting on them would lock in behaviour that
looks unintended rather than deliberate, such as the `aria-disabled` attributes on the prev and next
buttons rendering as literal strings. Happy to open a separate issue for those.

### Changelog

| Language | Changelog |
| -------- | --------- |
| English  |           |
| Chinese  |           |

### Self Check before Merge

### DIFF
--- a/tests/AntDesign.Tests/Pagination/PaginationTests.razor
+++ b/tests/AntDesign.Tests/Pagination/PaginationTests.razor
@@ -180,7 +180,7 @@
     }
 
     [Fact]
-    public void Any_other_key_leaves_the_current_page_alone()
+    public void Pressing_a_key_other_than_enter_does_not_change_the_page()
     {
         var cut = Render(@<Pagination Total="50" Current="3" />);
 
@@ -262,7 +262,7 @@
     {
         var cut = Render(@<Pagination Total="@total" />);
 
-        cut.FindAll(".ant-pagination-options-size-changer").Any().Should().Be(expected);
+        cut.FindAll(".ant-pagination-options-size-changer").Count.Should().Be(expected ? 1 : 0);
     }
 
     [Fact]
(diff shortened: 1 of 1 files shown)


thaildhe172591:
@dotnet-policy-service agree

thaildhe172591:
A note for whoever picks this up: the three red checks here are not caused by this branch.

`build`, `build preview` and `start preview info` fail the same way on other current pull requests from forks, #4842 and #4841 among them. They need repository secrets that a fork does not get, so they cannot pass from outside the org regardless of the change.

The test suite itself is green. I ran it locally across the target framework matrix: net8.0, net9.0 and net10.0 all pass 43 of 43, and net5 and net6 build clean, which is as far as I can go without those runtimes installed. No new warnings come from the added file.

ElderJames:
@thaildhe172591 I wiil fix this, thank you!

PR #4849, opened 2026-09-07
fix(module: tree): parenthesise the drag preventDefault expressions

### LINKED ISSUE #4848: Build fails on master with CS1503 in TreeNodeTitle.razor under .NET SDK 10.0.400
### Describe the bug

`dotnet build` fails on master with four `CS1503` errors in `components/tree/TreeNodeTitle.razor`, so every open pull request is red and nothing can merge. The last commit on master is from 5 August, and PRs opened since then all fail the same way, including maintainer ones such as #4840 and #4847.

The component itself did not change. `TreeNodeTitle.razor` was last touched in January (#4749). What moved is the SDK: `pull-request.yml` asks `actions/setup-dotnet` for `10.0.x` and there is no `global.json`, so the build floats to whatever 10.0 SDK is current on the runner. That is now 10.0.400.

The four attributes pass their condition as an implicit expression:

```razor
@ondragover:preventDefault="@TreeComponent.DragItem!=null"
```

The newer Razor parse reads `!` as the null-forgiving operator, so `DragItem!=null` becomes an assignment rather than a comparison and the attribute value ends up typed `TreeNode<TItem>` where a `bool` is expected.

No app project is needed, the component library alone reproduces it. On a clean checkout of master:

```
dotnet build components/AntDesign.csproj -f net10.0
```

Result depends only on the SDK:

| SDK | Result |
| --- | --- |
| 10.0.302 | Build succeeded, 0 errors |
| 10.0.400 | 4 errors, the ones below |

Other target frameworks fail the same way on 10.0.400, since the error is in the Razor parse rather than the target.

### Exceptions (if any)

```
components/tree/TreeNodeTitle.razor(10,31): error CS1503: Argument 4: cannot convert from 'AntDesign.TreeNode<TItem>' to 'bool'
components/tree/TreeNodeTitle.razor(11,32): error CS1503: Argument 4: cannot convert from 'AntDesign.TreeNode<TItem>' to 'bool'
components/tree/TreeNodeTitle.razor(12,32): error CS1503: Argument 4: cannot convert from 'AntDesign.TreeNode<TItem>' to 'bool'
components/tree/TreeNodeTitle.razor(13,27): error CS1503: Argument 4: cannot convert from 'AntDesign.TreeNode<TItem>' to 'bool'
```

### Further technical details

- master at 11397d0
- Reproduced with SDK 10.0.400 on Ubuntu, same versions CI installs; SDK 10.0.302 on Windows builds it clean
- Only these four sites are affected. The other `!=null` occurrences under `components/` sit inside `@if (...)` or `@(...)`, where the parse is unambiguous.

Writing each condition as `@(TreeComponent.DragItem != null)` restores the comparison and builds on both SDKs. Happy to send that as a PR.

Worth deciding separately whether the build should pin an SDK through `global.json`, since floating `10.0.x` means a runner update can break master without any commit.

thaildhe172591:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Fixes #4848

### 💡 Background and solution

`dotnet build` fails on master with four `CS1503` errors in `TreeNodeTitle.razor`, so every open pull request is red and nothing merges. The component did not change, the file was last touched in January. The build asks `setup-dotnet` for `10.0.x` with no `global.json`, so it floats to the current SDK, and 10.0.400 parses the drag conditions differently.

The four attributes pass their condition as an implicit expression:

```razor
@ondragover:preventDefault="@TreeComponent.DragItem!=null"
```

The `!` is read as the null-forgiving operator, which turns `DragItem!=null` into an assignment instead of a comparison, so the attribute value is typed `TreeNode<TItem>` where a `bool` is expected.

Writing the condition explicitly makes it a comparison again:

```razor
@ondragover:preventDefault="@(TreeComponent.DragItem != null)"
```

Same change on all four drag attributes, nothing else touched. `components/AntDesign.csproj -f net10.0` before and after:

| SDK | Before | After |
| --- | --- | --- |
| 10.0.302 | succeeded | succeeded |
| 10.0.400 | 4 errors | succeeded |

The other `!=null` occurrences under `components/` are inside `@if (...)` or `@(...)`, where the parse is unambiguous, so they are left alone.

One thing worth deciding separately, not part of this PR: a floating `10.0.x` means a runner update can turn master red without any commit. Pinning through `global.json` would make that a deliberate change instead.

### DIFF
--- a/components/tree/TreeNodeTitle.razor
+++ b/components/tree/TreeNodeTitle.razor
@@ -7,10 +7,10 @@
 <span class="@TitleClassMapper.Class" title="@SelfNode.Title" @onclick="OnClick" @ondblclick="OnDblClick"
 	@oncontextmenu="OnContextMenu" @oncontextmenu:preventDefault="TreeComponent.OnContextMenu.HasDelegate"
 	draggable="@(CanDraggable ? "true" : "false")" aria-grabbed="@(CanDraggable ? "true" : "false")"
-	@ondragover:preventDefault="@TreeComponent.DragItem!=null" @ondragover="OnDragOver"
-	@ondragleave:preventDefault="@TreeComponent.DragItem!=null" @ondragleave="OnDragLeave"
-	@ondragenter:preventDefault="@TreeComponent.DragItem!=null" @ondragenter="OnDragEnter"
-	@ondrop:preventDefault="@TreeComponent.DragItem!=null" @ondrop="OnDrop"
+	@ondragover:preventDefault="@(TreeComponent.DragItem != null)" @ondragover="OnDragOver"
+	@ondragleave:preventDefault="@(TreeComponent.DragItem != null)" @ondragleave="OnDragLeave"
+	@ondragenter:preventDefault="@(TreeComponent.DragItem != null)" @ondragenter="OnDragEnter"
+	@ondrop:preventDefault="@(TreeComponent.DragItem != null)" @ondrop="OnDrop"
 	@ondragstart="e => { if (CanDraggable) OnDragStart(e); }" @ondragend="e => { if (CanDraggable) OnDragEnd(e); }">
     @if (TreeComponent.TitleIconTemplate != null && TreeComponent.ShowIcon)
     {


ElderJames:
Thank you @thaildhe172591 !

PR #4850, opened 2026-09-10
fix(module:date-picker): accept date-only input when ShowTime is enabled

DeanOrnys:
### 🤔 This is a ...

- [x] Bug fix

### 🔗 Related issue link

_No upstream issue filed by the reporter; surfaced from a user report._

### 💡 Background and solution

**Reproduction**

```razor
<DatePicker TValue="DateTime?"
            ShowTime="@true"
            Format="dd.MM.yyyy HH:mm"
            Mask="dd.MM.yyyy HH:mm"
            @bind-Value="date" />
```

1. Type a date into the input without time, e.g. `01.01.2024`.
2. Blur / press Enter.
3. The field clears instead of keeping the date with `00:00`.
4. Open the dropdown — the calendar is on the default date, not on `01.01.2024`.

**Root cause**

`DatePickerBase.OnInput` calls `FormatAnalyzer.TryPickerStringConvert` which
rejects any input whose length is shorter than the full format (`dd.MM.yyyy HH:mm`
= 16 chars). For a mask-converted value of `01.01.2024 ` (11 chars) the parse
fails, `_pickerStatus.SelectedValue` is set to `null`, and neither
`ChangePickerValue` nor `ChangeValue` runs. On blur the field is rebuilt from
the (still-null) value, and the calendar stays on the default.

**Fix**

In `DatePickerBase.OnInput`, when the full mask parse fails, try to parse just
the date portion of the mask and combine it with a new `DefaultTime` parameter
(defaults to `TimeSpan.Zero` = `00:00:00`). On success, run the same
`SelectedValue` → `ChangePickerValue` → `ChangeValue` path as a complete input,
so the field keeps the date and the calendar syncs to it.

The `DefaultTime` parameter is exposed publicly so callers can pick a different
fallback time:

```razor
<DatePicker ShowTime="true"
            Format="dd.MM.yyyy HH:mm"
            Mask="dd.MM.yyyy HH:mm"
            DefaultTime="@(new TimeSpan(9, 0, 0))"
            @bind-Value="date" />
```

`FormatAnalyzer.GetDateOnlyFormat(string)` is a new public helper that returns
the date portion of a format (everything before the first `h/H/m/s/t`, with
trailing separators trimmed). It is reused internally by the fallback but is
also available for any future caller that needs the same extraction.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | `DatePicker`: typing a date without time no longer clears the field when `ShowTime` is enabled together with a `Format`/`Mask` that includes time. The missing time portion is now filled with the new `DefaultTime` parameter (default `00:00:00`). The calendar also syncs with the typed date. |
| 🇨🇳 Chinese | `DatePicker`: 在 `ShowTime` 与含时间的 `Format`/`Mask` 组合下，只输入日期不再清空字段，未填的时间部分将使用新增的 `DefaultTime` 参数（默认 `00:00:00`）填充。日历也会同步到所输入的日期。 |

- [x] Doc is updated/provided or not needed _(the parameter is self-documented via XML doc; a demo update can be added if the maintainers want it)_

### DIFF
--- a/components/date-picker/internal/DatePickerBase.cs
+++ b/components/date-picker/internal/DatePickerBase.cs
@@ -204,6 +204,22 @@ namespace AntDesign
             }
         }
 
+        private TimeSpan? _defaultTime;
+
+        /// <summary>
+        /// Time of day used to fill the time portion when the user types only a date
+        /// (e.g. <c>01.01.2024</c>) into a picker whose format includes time
+        /// (e.g. <c>dd.MM.yyyy HH:mm</c>). Defaults to <see cref="TimeSpan.Zero"/>
+        /// (00:00:00) when <c>null</c>.
+        /// </summary>
+        /// <default value="null"/>
+        [Parameter]
+        public TimeSpan? DefaultTime
+        {
+            get => _defaultTime;
+            set => _defaultTime = value;
+        }
+
         /// <summary>
         /// Allow clearing the selected value or not
         /// </summary>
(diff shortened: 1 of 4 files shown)


ElderJames:
Thank you for contribution @DeanOrnys ，please sign the CLA first.

DeanOrnys:
@dotnet-policy-service agree company="Quirco"

DeanOrnys:
@ElderJames done

ElderJames (approved):

